PR #37, opened 2018-11-04
feat: add semantic-dom-diff

daKmoR:


LarsDenBakker:
@daKmoR pulled over original source and tests, fixed a whitespace bug and improved error path with class names. can be merged I think

daKmoR:
ah finally ci is back 👍  asking one browser to test different packages in parallel was not too smart 🙈 🤣 

I will cleanup the history and merge it tomorrow

PR #47, opened 2018-11-16
chore: don't depend on lit-html directly for usage from bower

LarsDenBakker:
We only use lit-html in the lit fixture. If people use lit fixture, they must already have lit-html installed so we can get away with this trick.

### DIFF
--- a/packages/chai-dom-equals/package.json
+++ b/packages/chai-dom-equals/package.json
@@ -22,7 +22,6 @@
     "@open-wc/testing-wallaby": "^0.1.3",
     "@webcomponents/webcomponentsjs": "^2.0.0",
     "chai": "^4.0.0",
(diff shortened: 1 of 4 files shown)


LarsDenBakker:
As far as I could tell, only testing-helpers uses lit-html. The other's didn't import it anywhere.

What kind of range do you think we should set?

daKmoR:
I think every version since the "syntax" change to `.?@` should be fine... that would be `"lit-html": "^0.11.0"` although 0.12 will just be fine as well I guess - not many users for it I would say... 

so LGTM - feel free to merge

PS:  I thought ranges need to be defined special for peer but that works just fine - my bad

daKmoR:
uh I love this "auto" release mode ❤️

PR #50, opened 2018-11-17
🎉 Initial commit `tslint-config` and `typescript-lit-element`

motss:
Please help review the new generator. Feel free to make changes if needed.

P/S: Please expect to see an error regarding invalid @open-wc/tslint-config as this is newly created config for TSLint which yet to exist in the NPM registry.


motss:
I see what's going on now. Thank you for pointing that out.

PR #51, opened 2018-11-17
fix: report snapshot for the right HTML tree

LarsDenBakker:


LarsDenBakker:
@daKmoR I shortened the commit message, shall I merge it?

PR #54, opened 2018-11-18
feat(testing): auto-register side-effects

LarsDenBakker:
Turns out we aren't using the testing package in any of the other packages, so it's just one change.

EDIT: Did have to update docs and generator :)


packages/testing/register-chai-plugins.js
```diff
@@ -0,0 +1,4 @@
+import { chai } from '@bundled-es-modules/chai';
```
daKmoR: that look nice :)

packages/testing/index.js
```diff
@@ -1,14 +1,4 @@
...
-  aTimeout,
-  litFixture,
-  litFixtureSync,
-  fixture,
-  fixtureSync,
-} from '@open-wc/testing-helpers';
+import './register-chai-plugins.js';
+import './register-fixture-cleanup.js';
+
+export * from './index-no-side-effects.js';
```
daKmoR: we probably want to list all the exports here as well... as otherwise, IDEs may not provide nice autocomplete... also iirc there is still a rollup bug that will break if you do it

packages/testing/package.json
```diff
@@ -14,6 +14,7 @@
     "test:bs": "karma start karma-bs.config.js"
   },
   "dependencies": {
+    "@bundled-es-modules/chai": "^4.2.0-rc.1",
     "@open-wc/chai-dom-equals": "^0.3.2",
     "@open-wc/testing-helpers": "^0.2.1",
     "chai": "^4.0.0",
```
daKmoR: we probably can get rid of sinon then completely? if users want it they should add it on their own to their devDependencies
or do we want to reexport it in testing? so all you need is one dependency?

I think I would be for users to import it themselves (as long as we do not offer anything on top of it - which I hope we never need to do)

LarsDenBakker: Yes I agree, let users import it themselves. Sinon is a great library on itself, I don't expect we will need to add anything

daKmoR:
yes there should be no circle dependencies...
but they all of their "own" setup-bdd files... so we should change these tests then to also use the es module version of chai

packages/testing/package.json
LarsDenBakker: so we should removed it from all packages? a lot of them have a dependency

LarsDenBakker: Let's do that in a separate MR?

LarsDenBakker:
Updated the tests of all the packages

packages/testing/package.json
daKmoR: we could do it in separate MR but I would rather not have 2 separate releases for that - so then we would need a dev branch* and merge both there and then merge to master to release.
removal of sinon in package.json and in some index.html files and some wallaby and karma configs... is I think not worth it... (so let's do it in a separate commit but in the same branch) I could do that 👍 

* we will definitely need that at some point, but I would like to avoid it as long as possible :p

PR #63, opened 2018-12-01
fix: double // in generated path + tests

daKmoR:
- fixes double `.//` in for generated path e.g. `.//path/to/file`
- Adds mocha/node tests for import-meta-url-loader
- runs these tests in ci


LarsDenBakker:
LGTM!

PR #66, opened 2018-12-02
fix(testing-karma): ensure karma-webpack creates a single bundle

LarsDenBakker:
This fixes the issue of karma-webpack creating a separate bundle per test file.

I think that configs should have smart defaults, but still be overrideable and understandable. I iterated over different solutions, but couldn't find a good solution that covers these. At it's core, this is a bug in karma-webpack, so we should look into whether we can fix it on their side.

With this approach, by default all files in the `/test` folder ending with `.test` are run. It can be overwritten through by parameters to the config function (can't do it through the files property). Consumers can also opt out entirely by setting their own files property.


daKmoR:
closed in favor of https://github.com/open-wc/open-wc/pull/89

PR #72, opened 2018-12-05
issue-71

giantpanpan:
Hi I updated the register-fixture-cleanup. Also, I found that in packages/chai-dom-equals/test/bdd-setup.js, we still need to do the register-cleanup. So, I add try and catch to it as well to make it safe.


daKmoR:
oups I forgot to actually "submit" the review 🙈

daKmoR:
solved by https://github.com/open-wc/open-wc/pull/136

PR #84, opened 2018-12-12
feat(prettier-config): add prettier and eslint-config-prettier

### LINKED ISSUE #59: [linting] Add prettier
Let's consider adding `prettier` to the list of recommendations.

Pros:
- Supports `html` tagged literals and HTML syntax in general starting from v1.15, see [blog post](https://prettier.io/blog/2018/11/07/1.15.0.html#html-vue-angular)
- Supports `css` tagged literals being used by `lit-css` and `styled-lit-element`

Also worth mentioning:
- `eslint-config-prettier`
- `stylelint-config-prettier`
- `stylelint-processor-styled-components`

There is working example in https://github.com/web-padawan/lit-components

web-padawan:
Fixes #59

### DIFF
--- a/packages/prettier-config/README.md
+++ b/packages/prettier-config/README.md
@@ -1,0 +1,51 @@
+# Linting Prettier
+
+> Part of Open Web Component Recommendation [open-wc](https://github.com/open-wc/open-wc/) Recommendations [open-wc](https://open-wc.org/)
+
+We want to provide a good set of defaults on how to facilitate your web component.
+
+[![CircleCI](https://circleci.com/gh/open-wc/open-wc.svg?style=shield)](https://circleci.com/gh/open-wc/open-wc)
+[![BrowserStack Status](https://www.browserstack.com/automate/badge.svg?badge_key=M2UrSFVRang2OWNuZXlWSlhVc3FUVlJtTDkxMnp6eGFDb2pNakl4bGxnbz0tLUE5RjhCU0NUT1ZWa0NuQ3MySFFWWnc9PQ==--86f7fac07cdbd01dd2b26ae84dc6c8ca49e45b50)](https://www.browserstack.com/automate/public-build/M2UrSFVRang2OWNuZXlWSlhVc3FUVlJtTDkxMnp6eGFDb2pNakl4bGxnbz0tLUE5RjhCU0NUT1ZWa0NuQ3MySFFWWnc9PQ==--86f7fac07cdbd01dd2b26ae84dc6c8ca49e45b50)
+[![Renovate enabled](https://img.shields.io/badge/renovate-enabled-brightgreen.svg)](https://renovatebot.com/)
+
+Uses [Prettier](https://prettier.io) to format your JS, CSS and HTML code.
+
+## Manual
+
+- Install `@open-wc/prettier-config`.
+
+```bash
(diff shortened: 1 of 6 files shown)


packages/prettier-config/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "homepage": "https://github.com/open-wc/open-wc/",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": "https://github.com/open-wc/open-wc/tree/master/packages/prettier-config",
+  "main": ".prettierrc.js",
+  "dependencies": {
+    "eslint-config-prettier": "^3.3.0",
+    "husky": "^1.2.0",
```
daKmoR: I would say husky & lint-staged should not be dependencies of the config itself.
We should, however, explain how to set it up in the readme.

=> not (needed) for this MR (unless you are really eager to work on it) but in the long run we should have a "meta" package `linting` (like we have for testing) which will have a dependency on husky/lint-staged and that will be the entry point for most users.

packages/prettier-config/README.md
```diff
@@ -0,0 +1,60 @@
...
+- Create `.prettierrc.js` to root directory of the project as following.
+
+```js
+module.exports = require('@open-wc/prettier-config');
+```
+
+- Add these lines to your package.json
+
+```js
+"husky": {
```
daKmoR: I think I would prefer for both husky and lint-staged the `*.config.js` version
e.g. `lint-staged.config.js` and `husky.config.js`

what do you think?

according to http://wiki.commonjs.org/wiki/Packages/1.1
> Package managers and loaders should ignore unknown fields in the package descriptor file.

so it's sort of not in line (they have more Reserved Properties - rest "could be" open) but it's also very common...

packages/prettier-config/.prettierrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  printWidth: 80,
```
daKmoR: we are extending `eslint-config-airbnb-base` for eslint which has a max-len of 100

packages/prettier-config/package.json
web-padawan: Let's consider adding these dependencies in scope of separate PR.

packages/prettier-config/README.md
web-padawan: Good point, thanks.

packages/prettier-config/.prettierrc.js
web-padawan: Done.

packages/prettier-config/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@open-wc/prettier-config",
+  "version": "0.1.0",
```
daKmoR: this will release 0.2.0 within the ci as standard version will bump minor for 'feat' commits...
probably best to set it to 0.0.0 and let the ci release 0.1.0
=> again not blocking we can merge as is if you want

packages/prettier-config/package.json
daKmoR: sounds good :)

daKmoR:
added to homepage
will merge and release now :)

PR #88, opened 2018-12-12
fix: apply prettier

daKmoR:


daKmoR:
all was good except for chai-dom-equals that needed a `// prettier-ignore` for one function (but only on Firefox and Safari)... thank god we have tests covering that :)

PR #92, opened 2018-12-18
fix: add testing generator + update docu

daKmoR:


packages/testing/README.md
```diff
@@ -24,12 +24,24 @@ This is part of the default [open-wc](https://open-wc.org/) recommendation
 
 ## Setup
 ```bash
-npx -p yo -p generator-open-wc -c 'yo open-wc:testing-bare'
+npx -p yo -p generator-open-wc -c 'yo open-wc:testing'
 ```
 
+This runs the setup for multiple packages (testing-bare, testing-karma, testing-karma-bs).
```
bennypowers: Instead of describing what these packages do in inline-comments in the code block for the manual section, describe the whole recommendation above in `## Setup`, then add specific install instructions under `### Automatic Setup` and `### Manual Setup`

packages/testing/README.md
```diff
@@ -41,6 +53,29 @@ This will have the following side effect:
   - use the plugin [chai-dom-equals](https://www.npmjs.com/package/@open-wc/chai-dom-equals)
   - enables cleanup after each test for `fixture` and `litFixture`
 
+### Manual Karma Setup
```
bennypowers: Should these next two `###` be a part of `### Manual Setup` above?

packages/testing/README.md
daKmoR: updated it can you pls recheck

packages/testing/README.md
daKmoR: made it to ####

packages/testing/README.md
bennypowers: I would add a new paragraph immediately under `Setup`, and change the heading `Setup` to `Automating Tests`

```
## Automating Tests
Normally, you'll want some way of automatically running all of your tests, for that we recommend karma via @open-wc/testing-karma and browserstack via @open-wc/testing-karma-bs. If you use a different runner or a different browser grid then these steps can be skipped.

### Automatic Install
...
### Manual Install
...
```

bennypowers:
I would also maybe change the heading `Example` to `Example Test`

bennypowers:
And maybe show the output from the example test.

packages/testing/README.md
daKmoR: change it to suggested way

daKmoR:
added a runnable/testable demo and now also show the output in the readme.

you can see it here: https://deploy-preview-92--nervous-stonebraker-1f429b.netlify.com/recommendations/testing.html#example-tests

PR #94, opened 2018-12-18
feat: use extendable karma configs by default

daKmoR:
This is a proposal for a new way of configuring karma.

Essentially you now do all your configurations within karma.conf and all other configs are extensions and they only merge in their specifics.
This makes it easier to use but still allow full control about the config.

karma.conf.js
```js
/* eslint-disable import/no-extraneous-dependencies */
const defaultSettings = require('@open-wc/testing-karma/default-settings.js');
const merge = require('webpack-merge');

module.exports = config => {
  config.set(
    merge(defaultSettings(config), {
      files: [
        // allows running single tests with the --grep flag
        config.grep ? config.grep : 'test/**/*.test.js',
      ],

      // you custom config
    }),
  );
  return config;
};
```

karma.es5.config.js
```js
/* eslint-disable import/no-extraneous-dependencies */
const merge = require('webpack-merge');
const es5Settings = require('@open-wc/testing-karma/es5-settings.js');
const karmaConf = require('./karma.conf.js');

module.exports = config => {
  config.set(merge(es5Settings(config), karmaConf(config)));
  return config;
};
```

karma.es5.bs.config.js
```js
/* eslint-disable import/no-extraneous-dependencies */
const merge = require('webpack-merge');
const bsSettings = require('@open-wc/testing-karma-bs/bs-settings.js');
const karmaEs5Config = require('./karma.es5.config.js');

module.exports = config => {
  config.set(merge(bsSettings(config), karmaEs5Config(config), {
    browserStack: {
      project: 'your-name',
    }
  }));
  return config;
};
```


daKmoR:
added a second commit showing the version of using a `setTestFiles` function

PR #101, opened 2018-12-21
fix: remove deprecated flush

thepassle:
Removes flush from the docs at https://open-wc.org/recommendations/testing-helpers.html#timings


daKmoR:
can we maybe remove the function in this PR as well... as merges to master trigger a release

also pls make sure to use a proper commit message. e.g. `fix: remove deprecated flush`

Issue #53, opened 2018-11-18
[semantic-dom-diff] splice diffs not correctly reported


LarsDenBakker:
When there is a 'splice' diff, for example:

expected: `<div a="1" b="2" c="3"></div>`
actual: `<div a="1" c="3"></div>`

you expect the error message to say something like `attribute b was removed at path div` but in reality it says `attribute c was removed at path div`. 

This is because `deep-diff` does not see this as one change, but as multiple changes. I don't remember exactly, but I think it reports 'c was removed, b was removed, c was added'. We only take the first change and report that back so we need be a bit smarter about how we parse the response from `deep-diff`.

I've been working on fixing this in a branch, but didn't get too far yet.

PR #111, opened 2018-12-29
feat(chai-dom-equals): improve chai-dom-equals error reporting

LarsDenBakker:
fixes https://github.com/open-wc/open-wc/issues/53

The error reporting of chai-dom-equals and semantic-dom-diff were always a bit strange in complex cases. This improves it by taking advantage of built-in chai ability to diff strings with newlines. 

This improvement will only be visible when utilizing mocha's advanced/deep error reporting (our karma setup uses it, but WCT does not).

I also removed the `.semantically` part of the chai plugin, with these changes a non-semantic quality is no longer necessary. I kept it in for backwards compatibility so that we don't break people's tests.


packages/chai-dom-equals/test/chai-dom-equals.test.js
```diff
@@ -152,19 +152,19 @@ describe('getCleanedShadowDom', () => {
 });
 
 describe('dom', () => {
-  it('can strictly compare dom nodes', async () => {
+  it('can compare dom nodes', async () => {
     const el = await fixture(`<div><!-- comment --><h1>${'Hey'}  </h1>  </div>`);
-    expect(el).dom.to.equal('<div><!-- comment --><h1>Hey  </h1>  </div>');
```
daKmoR: so if I want to do a "strict" comparison of dom I would need to write it myself right?
e.g.
```
expect(el.innerHTML).to.equal('<div><!-- comment --><h1>Hey  </h1>  </div>');
```

problem is that this will also include all the shadycss markers if you use shadodom... so is there a solution to compare strict but to ignore the shadycss stuff?

packages/semantic-dom-diff/src/parse5-utils.js
```diff
@@ -1,6 +1,6 @@
-export const isNode = arg => arg && 'nodeName' in arg;
+// export const isNode = arg => arg && 'nodeName' in arg;
```
daKmoR: if no longer needed => let's just remove it

packages/chai-dom-equals/test/chai-dom-equals.test.js
LarsDenBakker: Updated readme

PR #114, opened 2019-01-01
feat(testing-helpers): fixture can handle strings and TemplateResults

daKmoR:
- `fixture` => `stringFixture`
- new `fixture` can handle strings and lit-html TemplateResults
- types and documentation update

### DIFF
--- a/packages/testing-helpers/index.d.ts
+++ b/packages/testing-helpers/index.d.ts
@@ -1,12 +1,13 @@
-export { html, unsafeStatic } from './lit-html';
+export { html, unsafeStatic } from './lit-html.js';
 export {
-  aTimeout,
-  defineCE,
-  isIE,
-  nextFrame,
-  oneEvent,
   triggerBlurFor,
   triggerFocusFor,
-} from './helpers';
-export { litFixture, litFixtureSync } from './litFixture';
-export { fixture, fixtureSync } from './fixture';
+  oneEvent,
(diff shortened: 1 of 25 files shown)


daKmoR:
@bennypowers @tpluscode can you pls check the typings - I am no master at this 🙈

LarsDenBakker:
Looks good to me, the types as well.

packages/testing-helpers/index.d.ts
```diff
@@ -1,12 +1,13 @@
-export { html, unsafeStatic } from './lit-html';
+export { html, unsafeStatic } from './lit-html.js';
```
tpluscode: Are extensions correct here?

daKmoR: yes that is the local file... where we support dynamic tags (which is not yet available in official version) (note: this implementation is only suitable for testing as it's not performant)

bennypowers: Should probably rename the tag and file then.

daKmoR: it hopefully will land in lit-html right after 1.0 and then this would mean no change is needed at all for your test files... so "forward" compatible :p

I thought it was a worthwhile goal to not have to adopt every test file... with your replies I am not so sure anymore... anyways I think that change should be another PR... so if really needed pls open an issue :)

PR #116, opened 2019-01-03
docs(testing-helpers): fix fixture call on samples

### LINKED ISSUE #115: Clarify use of await before fixture helper
The docs are not consistent about usage of await before fixture helper:

In test custom element [example](https://open-wc.org/recommendations/testing-helpers.html#test-a-custom-element) await is used.
Just below, in test custom class [example](https://open-wc.org/recommendations/testing-helpers.html#test-a-custom-class) it does not use await.

This confused me: is necessary or not use await?

After looking into the source i found that fixture is a wrapper around fixtureSync followed to a nextFrame call.

It would be good to newcomers to clarify when await is needed before a fixture call and when not.

As i can see when using a string or a lit-html tagged template as fixture is not needed, but when using a lit-element await is necessary.

blikblum:
Fixes #115

### DIFF
--- a/packages/testing-helpers/README.md
+++ b/packages/testing-helpers/README.md
@@ -47,7 +47,7 @@ const tag = defineCE(class extends MyMixin(HTMLElement) {
     this.foo = true;
   }
 });
-const el = fixture(`<${tag}></${tag}>`);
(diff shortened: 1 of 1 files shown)


daKmoR:
thxxx :heart:

PR #119, opened 2019-01-09
fixes code blocks

arikimolloy:


packages/testing-karma/README.md
```diff
@@ -84,12 +84,12 @@ By default karma runs all your test files. To test only a single file or folder,
 Grep supports file globs.
 
 ### Watch mode
-Use `npm run test -- --auto-watch=true --single-run=false` to keep karma running. Any code changes will trigger a re-run of your tests.
```
daKmoR: that is actually correct as the first "--" is the separation between the npm command then the "real" command
e.g.
```bash
npm run test -- --auto-watch=true --single-run=false
# executes
karma run --auto-watch=true --single-run=false
```

packages/testing-karma/README.md
```diff
@@ -84,12 +84,12 @@ By default karma runs all your test files. To test only a single file or folder,
...
+Use `npm run test --auto-watch=true --single-run=false` to keep karma running. Any code changes will trigger a re-run of your tests.
 
 ### Debugging in the browser
 While testing, it can be useful to debug your tests in a real browser so that you can use the browser's dev tools.
 
-Use `npm run test -- --single-run=false` to keep karma running. Then open the URL printed by karma when it boots up. By default this is `http://localhost:9876/`. Click the debug button in the top right corner, or go directly to `http://localhost:9876/debug.html`. You can bookmark this page for easy access.
```
daKmoR: same as above

packages/testing-karma/README.md
arikimolloy: ahh okay I see. I was not able to run it with npm, probably because my npm client is set to yarn. So I ran it with yarn and it didn't need the "--" separation. Will add that back.

daKmoR:
thx ❤️

PR #122, opened 2019-01-13
feat(webpack): added webpack default settings

LarsDenBakker:


packages/webpack/src/default-es5-settings.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = userOptions => {
+  return {
+    entry: [
+      require.resolve('@webcomponents/webcomponentsjs/webcomponents-bundle.js'),
+      require.resolve('@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js'),
+    ],
+    module: {
+      rules: [
+        {
+          test: /\.m?js$/,
```
daKmoR: do we want to include the import.meta loader by default as well?

packages/webpack/src/default-es5-settings.js
```diff
@@ -0,0 +1,41 @@
...
+          use: {
+            loader: 'babel-loader',
+            options: {
+              presets: ['@babel/preset-env'],
+              exclude: /(node_modules|bower_components)/,
```
daKmoR: `node_modules/lit-html` and `node_modules/lit-element`
need to be transpired right?
for storybook we have something that goes over all excludes and replaces it with an exclude function except this and this...
see: https://github.com/open-wc/open-wc/blob/master/packages/generator-open-wc/generators/publish-storybook/templates/static/.storybook/webpack.config.js#L14

packages/webpack/src/default-es5-settings.js
LarsDenBakker: I'd say so. It's a web standard.

LarsDenBakker: I'd like to do some profiling before concluding that it's slow :)

LarsDenBakker:
This config was inspired by your work @web-padawan could you do a review?

LarsDenBakker:
Ready for review. Still have issues on IE11 due to: https://github.com/webcomponents/webcomponentsjs/issues/972

packages/webpack/README.md
```diff
@@ -8,28 +8,127 @@ We want to provide a good set of defaults on how to facilitate your web componen
...
+  "scripts": {
+    "start": "webpack-dev-server --mode development --open",
+    "start:es5": "webpack-dev-server --mode development --open --es5",
+    "start:build": "http-server dist/ -o",
+    "build": "rm -rf dist && webpack --mode production"
+  }
+}
+```
+- `start` will run your app for development, it works only browsers which supports `es2015` for faster builds
+- `start:es5` will run your for development, it works also on browsers which support `es5`
```
daKmoR:
```suggestion
- `start:es5` will run your app for development, it works also on browsers which only support `es5`
```

thepassle: "it works also" -> "it also works"

packages/webpack/README.md
```diff
@@ -8,28 +8,127 @@ We want to provide a good set of defaults on how to facilitate your web componen
...
+### Automatic module type selection
+Based on `package.json` it selects the most optimal type of module format. es modules and es2015 code is preferred, but falls back to support other module types such as CommonJS and UMD. These can be a useful way to get done writing code today, but these are not future proof formats so be aware when importing non-es module dependencies.
+
+### HTML, JS and CSS minifications
+JS is minified for performance. HTML, CSS and is minified when they are used in combination with the `html` and `css` template tags from `lit-html` and `lit-element`.
+
+### es2015 and es5 output
+Using [webpack-babel-multi-target-plugin](https://www.npmjs.com/package/webpack-babel-multi-target-plugin), our build outputs an es5 and es2015 version of your app. Using the `nomodule` trick, we can serve es2015 code on modern browsers and es5 on older browsers (IE11 mostly). This significantly recudes the size of your app on modern browsers.
+
+Read more about this trick here: https://jakearchibald.com/2017/es-modules-in-browsers/
```
daKmoR:
```suggestion
Read more about this trick at [Jakes blog post "ECMAScript modules in browsers"](https://jakearchibald.com/2017/es-modules-in-browsers/).
```

packages/webpack/README.md
```diff
@@ -8,28 +8,127 @@ We want to provide a good set of defaults on how to facilitate your web componen
...
+This lets you use bare imports: `import { bar } from 'foo'` instead of `import { bar } from '../node_modules/foo/foo.js'`
+
+### Automatic module type selection
+Based on `package.json` it selects the most optimal type of module format. es modules and es2015 code is preferred, but falls back to support other module types such as CommonJS and UMD. These can be a useful way to get done writing code today, but these are not future proof formats so be aware when importing non-es module dependencies.
+
+### HTML, JS and CSS minifications
+JS is minified for performance. HTML, CSS and is minified when they are used in combination with the `html` and `css` template tags from `lit-html` and `lit-element`.
+
+### es2015 and es5 output
+Using [webpack-babel-multi-target-plugin](https://www.npmjs.com/package/webpack-babel-multi-target-plugin), our build outputs an es5 and es2015 version of your app. Using the `nomodule` trick, we can serve es2015 code on modern browsers and es5 on older browsers (IE11 mostly). This significantly recudes the size of your app on modern browsers.
```
daKmoR:
```suggestion
Using [webpack-babel-multi-target-plugin](https://www.npmjs.com/package/webpack-babel-multi-target-plugin), our build outputs an es5 and es2015 version of your app. Using the `nomodule` trick, we can serve es2015 code on modern browsers and es5 on older browsers (IE11 mostly). This significantly reduces the size of your app on modern browsers.
```

packages/webpack/README.md
```diff
@@ -8,28 +8,127 @@ We want to provide a good set of defaults on how to facilitate your web componen
...
+
+Read more about this trick here: https://jakearchibald.com/2017/es-modules-in-browsers/
+
+For modern browsers code is compiled to es2015, for older browsers (IE11 mostly) there is an es5 output. Using the `nomodule`
```
daKmoR:
```suggestion
```

daKmoR: it's the same info as 2 sentences before?

LarsDenBakker: Ah, yes :)

packages/webpack/README.md
```diff
@@ -8,28 +8,127 @@ We want to provide a good set of defaults on how to facilitate your web componen
...
+
+### es2015 and es5 output
+Using [webpack-babel-multi-target-plugin](https://www.npmjs.com/package/webpack-babel-multi-target-plugin), our build outputs an es5 and es2015 version of your app. Using the `nomodule` trick, we can serve es2015 code on modern browsers and es5 on older browsers (IE11 mostly). This significantly reduces the size of your app on modern browsers.
+
+Read more about this trick at [Jakes blog post "ECMAScript modules in browsers"](https://jakearchibald.com/2017/es-modules-in-browsers/).
+
+For modern browsers code is compiled to es2015, for older browsers (IE11 mostly) there is an es5 output. Using the `nomodule`
+
+### No regenerator runtime / transform
+By default babel compiles async functions and `for of` loops using an expensive transform and runtime. A lot of runtime code is added, and a lot of boilerplate code is added in the transform.
```
daKmoR:
```suggestion
By default, babels compilation of async functions and `for of` loops is pretty heavy. A big chunk of runtime code is added and a lot of boilerplate code is generated in the transform.
```

packages/webpack/README.md
```diff
@@ -8,28 +8,125 @@ We want to provide a good set of defaults on how to facilitate your web componen
...
+
+### es2015 and es5 output
+Using [webpack-babel-multi-target-plugin](https://www.npmjs.com/package/webpack-babel-multi-target-plugin), our build outputs an es5 and es2015 version of your app. Using the `nomodule` trick, we can serve es2015 code on modern browsers and es5 on older browsers (IE11 mostly). This significantly reduces the size of your app on modern browsers.
+
+Read more about this trick at [Jakes blog post "ECMAScript modules in browsers"](https://jakearchibald.com/2017/es-modules-in-browsers/).
+
+### No regenerator runtime / transform
+By default babel compiles async functions and `for of` loops using an expensive transform and runtime. A lot of runtime code is added, and a lot of boilerplate code is added in the transform.
+
+For async functions we opt for the [fast-async](https://github.com/MatAtBread/fast-async) plugin instead, `for of` loops are disallowed.
```
daKmoR: so we do not allow `for of` loops? what should you do instead? maybe link to a blog post or so?

`for of` loops are advertised to be easier and to be actually the future itself 😅🤔 

so our reasoning should be pretty good and/or we should have a good resource on this

packages/webpack/default-config.js
```diff
@@ -0,0 +1,144 @@
...
+        stats: 'errors-only',
+      },
+    },
+  };
+
+  if (ENV === 'production') {
+    config.plugins.push(
+      new CopyWebpackPlugin([
+        {
+          from: path.resolve(__dirname, 'node_modules/@webcomponents/webcomponentsjs/**/*.js'),
```
daKmoR: I'm curious why is this needed?

LarsDenBakker: The webcomponents-loader loads these files at runtime, so they need to be copied over to the build output.

docs/guide/building.md
```diff
@@ -0,0 +1,22 @@
...
+1. Code optimization (minification, dead code removal)
+2. Bundling
+3. Module resolution
+4. Transforming/polyfilling for older browsers
+5. Supporting non-standard code patterns
+
+We think it's best to avoid nr. 5 as non-standard code will lock your source code to a specific build setup which can be hard to escape in the future. Javascript is evolving fast, your code can quickly become out of sync with new development.
+
+## Webpack
+Webpack is a popular and flexible build tool, with a growing ecosystem of plugins. We recommend as a general purpose tool to get started. See [building-webpack](https://open-wc.org/recommendations/building-webpack.html) to get started.
```
daKmoR:
```suggestion
Webpack is a popular and flexible build tool, with a growing ecosystem of plugins. We recommend it as a general purpose tool to get started. See [building-webpack](https://open-wc.org/building/building-webpack.html) to get started.
```

docs/guide/building.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Rollup
+Rollup is a great build tool which recently reached 1.0. It's especially suitable for optimizing and minifying es modules, and can output es modules for browsers which have native support.
```
daKmoR:
```suggestion
[Rollup](https://rollupjs.org/) is a great build tool which recently reached 1.0. It's especially suitable for optimizing and minifying es modules, and can output es modules for browsers which have native support.
```

packages/webpack/README.md
LarsDenBakker: This is definitely up for discussion. Most linting plugins disallow `for of`, because of the runtime overhead and the minimal gain. Because we have an es2015 and es5 output, we could argue for allowing it.

packages/webpack/default-config.js
daKmoR: ah year right

docs/guide/building.md
daKmoR: updated suggestion to relative `/building/building-webpack` so it also links correctly on the preview pages

packages/webpack/README.md
web-padawan: Note that `lit-element` is now using `for of` internally, starting from 0.7.0

packages/webpack/src/default-es5-settings.js
LarsDenBakker: @daKmoR do you agree to pick this up later with profiling?

docs/building/README.md
```diff
@@ -0,0 +1,24 @@
+# Building
+
+> Part of Open Web Component Recommendation [open-wc](https://github.com/open-wc/open-wc/)
```
daKmoR: this header is only needed for readmes of individual packages to show them in npmjs.org
here we can remove them

packages/webpack/README.md
LarsDenBakker: They removed `for of` now :) I'd like to not have regenerator by default, it's easy add this option in yourself.

I did also see that `for of` over arrays gets compiled to a simple `for` loop, I don't know how well it can detect arrays though.

packages/building-webpack/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+5. Add the following commands to your `package.json`:
+```json
+{
+  "scripts": {
+    "start": "webpack-dev-server --mode development --open",
+    "start:es5": "webpack-dev-server --mode development --es5",
+    "start:build": "http-server dist/ -o",
+    "build": "rm -rf dist && webpack --mode production",
+    "bundle-stats": "webpack --mode production --profile --json > bundle-stats.json"
```
daKmoR: maybe "build:stats"?

packages/building-webpack/loaders/import-meta-url-loader.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable */
+const regex = /import\.meta/g;
+
```
daKmoR: maybe have a separate package? 
how about `@open-wc/webpack-import-meta-loader`?

packages/building-webpack/README.md
LarsDenBakker: Good suggestion, done.

packages/building-webpack/default-config.js
```diff
@@ -0,0 +1,130 @@
...
+    },
+
+    resolve: {
+      mainFields: ['es2015', 'module', 'main'],
+    },
+
+    module: {
+      rules: [
+        {
+          test: /\.js$/,
```
daKmoR: `npm run build` only works if I add here
```
 exclude: [modulePath => {
   return (
     /node_modules/.test(modulePath) &&
     !/node_modules\/lit-html/.test(modulePath) &&
     !/node_modules\/@open-wc/.test(modulePath)
   );
 }],
```

probably something to do with too many files to process or so :/
=> I guess having a static include/exclude list is still needed  
=> and to honest even if the auto magic from BabelMultiTargetPlugin would work I don't understand why this plugin would tackle that totally unrelated problem

daKmoR:
This is my output for `dist/index.html` of `npm run build`
```html
<body>
  <demo-app></demo-app>
  <script type="module" src="main.8d3151b7399aa23e1edd.js"></script>
  <script type="text/javascript" src="main.es5.c173fd37ad152e982b97.js"></script>
</body>
```

so basically the second `script` is missing the `nomodule` attribute... so it doesn't work if you open it in chrome

LarsDenBakker:
Somehow the `nomodule` option got turned off. I added it back, it works now.

packages/building-webpack/default-config.js
LarsDenBakker: It's not an unrelated problem. In order for the differential serving to work, the es2015 output should contain es2015 code while the es5 output should contain es5 code. Therefore there needs to be come logic to determine what should be transpiled and what shouldn't.

This project is not really large at all, the dependencies are the web component polyfills, lit-html and lit-element. In your exclude you're not compiling lit-element and the polyfills so probably it won't work on IE11. Maybe it's one of the polyfills that's causing issues for you>

daKmoR: it's an include list and we could have it in the default config?

e.g. something like this
```
defaultConfig({ babelInclude: ['lit-element', 'lit-html', '@open-wc'] })
```

I guess now it's a problem either with the BabelMultiTargetPlugin => think's it needs to transpiles all of npm and then it "breaks"
or with npm itself e.g. it can't process that many files on windows10

in either case I would not like to have a config that only works on mac/linux :/ as i could not use it myself then :(

if this needs a workaround until it's fixed somewhere - I would say so be it... 

what do you think?

LarsDenBakker: If the issue is the amount of files, then this setup won't work at all. Because as soon as you add a few more dependencies or develop some code then you will reach this point. There isn't a lot of code in there right now at all.

Let's check if someone with windows has the same issue.

daKmoR: oh boy it really is lit-element 😭 
if I do
````
 exclude: [modulePath => {
   return (
     /node_modules/.test(modulePath) &&
     !/node_modules\/lit-html/.test(modulePath) &&
     !/node_modules\/lit-element/.test(modulePath) &&
     !/node_modules\/@open-wc/.test(modulePath)
   );
 }],
```

it again stops working... :/

packages/webpack/src/default-es5-settings.js
daKmoR: sure

packages/building-webpack/loaders/import-meta-url-loader.js
daKmoR: yes better do the refactor of webpack => webpack-import-meta-loader later

packages/building-webpack/default-config.js
daKmoR: ok that is a window10 WSL problem 😭 
works fine in my plain bash... 😒

packages/building-webpack/default-config.js
```diff
@@ -0,0 +1,168 @@
...
+  browsers: [
+    'last 2 Chrome major versions',
+    'last 2 ChromeAndroid major versions',
+    'last 2 Edge major versions',
+    'last 2 Firefox major versions',
+    'last 3 Safari major versions',
+    'last 3 iOS major versions',
+  ],
+  tagAssetsWithKey: false,
+  esModule: true,
```
daKmoR: while playing around with just webpack I found out that combining browsers + esmodule is not allowed at least babel shows a warning for it (one of the reasons I'm in doubt of the plugin now as it apparently suppresses these messages...)
> Please note: when specifying the esmodules target, browsers targets will be ignored.

https://babeljs.io/docs/en/babel-preset-env#targetsesmodules

=> makes sense either you say work for these browsers or work for all browsers that support es modules

LarsDenBakker: This is not the preset-env config, but the multi target config. The `esModule` option here means whether it should be loaded as `type="module"`

packages/building-webpack/README.md
```diff
@@ -0,0 +1,139 @@
...
+    "build": "webpack --mode production",
+    "bundle-stats": "webpack --mode production --profile --json > bundle-stats.json"
+  }
+}
+```
+- `start` runs your app with auto reload for development, it only works on browsers which support modules for faster builds
+- `start:es5` runs your app for development, it only works on browsers that don't support modules (IE11)
+- `build` builds your app for production and outputs it in the `/dist` folder
+- `start:build` runs your built app using a plain web server, to prove it works without magic :)
+- `bundle-stats` creates an analysis report of your app bundle to be consumed by Webpack [Visualizer](https://chrisbateman.github.io/webpack-visualizer/) and [Analyser](https://webpack.github.io/analyse/)
```
daKmoR:
```suggestion
`build:stats` creates an analysis report of your app bundle to be consumed by Webpack [Visualizer](https://chrisbateman.github.io/webpack-visualizer/) and [Analyser](https://webpack.github.io/analyse/)
```

packages/polyfills-loader/polyfills-loader.js
```diff
@@ -20,8 +20,8 @@ function needsTemplatePolyfill() {
   template.content.appendChild(template2);
   const clone = template.cloneNode(true);
   return (
-    clone.content.childNodes.length !== 0 &&
-    clone.content.firstChild.content.childNodes.length !== 0
+    clone.content.childNodes.length === 0 ||
```
daKmoR: let's rebase => that's already on master

LarsDenBakker: I was going to squad this MR, that would achieve the same, right?

daKmoR: I rebased and force pushed

PR #130, opened 2019-01-16
Docs/styling

thepassle:


docs/.vuepress/styles/index.styl
```diff
@@ -0,0 +1,8 @@
+.home .hero .action-button
+	background-color: #217FF9
```
daKmoR:
```suggestion
	background-color: #354eff
```

docs/.vuepress/styles/palette.styl
```diff
@@ -0,0 +1 @@
+$accentColor = #217FF9
```
daKmoR:
```suggestion
$accentColor = #354eff
```

docs/.vuepress/styles/index.styl
thepassle: I'd say make it lighter then, this color is still pretty bright, and it doesnt really go well with the blue/purple that we have in the logo. A lighter blue makes it a bit easier on the eyes instead:
screen shot 2019-01-16 at 21 21 12

PR #132, opened 2019-01-17
docs: restructure menu

thepassle:
Restructures the menu in the docs as discussed on slack, it separates the menu into different sections that have the original '/guide/' docs, as well as any other tools/configs/setups we have for each category

Preview: https://5c4268583f37820009235548--nervous-stonebraker-1f429b.netlify.com/

screen shot 2019-01-17 at 22 35 03


daKmoR:
this "disables" the left menu for "FAQ" and "Contact"... e.g. it no longer changes/updates if you click on FAQ... probably need to add some routers or so

thepassle:
Fixed, really starting to get the hang of vuepress :p

daKmoR:
uh very nice :)

one final thing do you know if we could redirect those urls?
https://deploy-preview-132--nervous-stonebraker-1f429b.netlify.com/testing/

maybe to this?
https://deploy-preview-132--nervous-stonebraker-1f429b.netlify.com/guide/testing.html

it's not really important as you should never end up there... but I like to keep each layer of the url "functional"...

if we can't redirect we probably want to change the content you get presented? or you think it's ok?

PR #136, opened 2019-01-19
fix: move fixtureCleanup to testing helpers

daKmoR:
This moves the fixtureCleanup "magic/side-effect" from `@open-wc/testing` to `@open-wc/testing-helpers`.

So even if only the testing helpers are used the fixtures do get cleaned up 💪


LarsDenBakker:
Nice :)

PR #148, opened 2019-01-25
fix(polyfills-loader): needsTemplatePolyfill returned inverted values

daKmoR:


LarsDenBakker:
Force merge as its the flaky IE11 test

PR #150, opened 2019-01-26
feat: add webpack-import-meta-loader

daKmoR:


packages/webpack-import-meta-loader/README.md
```diff
@@ -0,0 +1,35 @@
+# Webpack Helpers
+
+> Part of Open Web Component Recommendation [open-wc](https://github.com/open-wc/open-wc/)
+
+We want to provide a good set of defaults on how to facilitate your web component.
```
LarsDenBakker: Not the same header as the other packages

daKmoR: fixed + force pushed

PR #151, opened 2019-01-26
feat(polyfills-loader): use only es modules in polyfills loader

LarsDenBakker:
Need to test it after webpack is merged

### DIFF
--- a/packages/polyfills-loader/package.json
+++ b/packages/polyfills-loader/package.json
@@ -11,14 +11,16 @@
   },
   "repository": "https://github.com/open-wc/open-wc/tree/master/packages/polyfills-loader",
   "keywords": [
-    "open-wc"
+    "open-wc",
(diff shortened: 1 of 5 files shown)


LarsDenBakker:
I've confirmed these changes work with the merged webpack branch. So we can merge this.

packages/polyfills-loader/package.json
```diff
@@ -9,16 +9,22 @@
   "publishConfig": {
     "access": "public"
   },
+  "scripts": {
+    "start": "http-server -o"
```
daKmoR: what does this do? I don't see an index.html? should there be a demo folder or so?

LarsDenBakker: I had set up a demo for testing, but realized it wouldn't work without module resolution. So I removed it again. Removed the scripts now too.

PR #152, opened 2019-01-26
docs: fix headers for consistency

thepassle:


daKmoR:
commit message should be a `fix` as it will trigger a new release (to be visible on npmjs).
And that release should have an entry in the changelog. (which docs does not generate)

how about
```
fix: align all open-wc readme headers
```

PR #153, opened 2019-01-26
Mixin Discussion [not ready for merge]

ZempTime:
## Context
Following up on:

screen shot 2019-01-26 at 3 15 47 pm

For the page I've added, here's a corresponding StackBlitz roughly modeled after the `open-wc` one: https://stackblitz.com/edit/open-wc-mixins

In there I have a variety of TODO's listed in there I think would be valuable to resolve. 

## Questions & Editing

- [x] is "steppable" a good mixin example?
- [ ] where should this go in the navigation?
- [x] is this too `LitElement`-specific?
- [ ] when should I use a mixin vs when should I create an element?

If this PR isn't a good place to conduct this process, then feel free to decline and we can do elsewhere.

What would be most valuable to me, specifically, is:
* reference on the syntax
* guidance around "when" - when to lean on a mixin? When do I actually have a component that isn't a component yet?

I looked for it on open-wc, which is why this PR. I'm down to do all the grunt work on getting this section up. I'm hoping to get a lot of perspective on how to properly use mixins as a tool, so drive-by comments are quite welcome!

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -11,7 +11,11 @@ const sidebar = [
   {
     title: 'Developing',
     collapsable: true,
-    children: [['/developing/', 'Getting started'], ['/developing/generator', 'Generators']],
+    children: [
+      ['/developing/', 'Getting started'],
+      ['/developing/generator', 'Generators'],
+      ['/developing/mixins', 'Mixins'],
+    ],
   },
   {
     title: 'Linting',
--- a/docs/developing/mixins.md
+++ b/docs/developing/mixins.md
@@ -1,0 +1,265 @@
+# Mixins
+
+Mixins are a great way to share common capabilities across your web components. They're a great fit when:
+
+* You need to centralize behavior you use to enhance the classes which power your components, or
+* You need to layer the way properties and state flow into your components
+
+## Sharing Capabilities
+
+A common interaction in many systems are forms with multiple steps:
+
+```js
+class SimpleWizard1 extends LitElement {
+  static get properties() {
+    return {
+      currentStepIndex: { type: Number },
+    }
+  }
+
+  constructor() {
+    super();
+    this.currentStepIndex = 0;
+  }
+
+  get currentStep() {
+    return this.steps[this.currentStepIndex];
+  }
+
+  get steps() {
+    return [
+      {
+        name: "Step 1",
+        template: () => html`
+          <label for="name">Name</label>
+          <input type="text" name="name" id="name" />
+        `
+      },
+      {
+        name: "Step 2",
+        template: () => html`
+          <label for="description">Description</label>
+          <input type="text" name="description" id="description" />
+        `
+      },
+      {
+        name: "Step 3",
+        template: () => html`
+          <label for="address">Address</label>
+          <input type="text" name="address" id="address" />
+        `
+      }
+    ]
+  }
+
+  previous() {
+    this.currentStepIndex = this.isFirstStep
+      ? this.currentStepIndex
+      : this.currentStepIndex - 1;
+  }
+
+  next() {
+    this.currentStepIndex = this.isLastStep
+      ? this.currentStepIndex
+      : this.currentStepIndex + 1;
+  }
+
+  get isFirstStep() {
+    return this.currentStepIndex === 0;
+  }
+
+  get isLastStep() {
+    return this.currentStepIndex === this.steps.length - 1;
+  }
+
+  render() {
+    return html`
+      ${this.currentStep.template()}
+      <div>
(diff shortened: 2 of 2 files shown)


web-padawan:
I have an ones for `TabIndexMixin` which could be extracted from my `lit-components` project and added as a small example.

It is small, though implementation still specific to LitElement. The problem it solves is lack of support for `attachShadow({ delegateFocus: true })`

LarsDenBakker:
I think it's a good idea to have a section on mixins. One thing I would add is a comparison to similar concepts in other language. For instance interfaces with default implementations in Java, and traits in Scala.

thepassle:
I agree with Lars, it would be really nice to have comparisons with other language. 

I get that its just an initial draft, but i'd like to align the design of the stackblitz a bit more with the stackblitz demo's we have [here](https://stackblitz.com/edit/open-wc-lit-demos), I can help with that as well. Also im thinking of how to store the stackblitz, I think we might have to make a separate repo for it like we did with the lit-demos [like this](https://github.com/open-wc/lit-demos), to make sure we keep everything central.

As for navigation, I think we can put it under the `Developing` dropdown as submenu `Mixins`. I'm hopeful we'll have a `Accessibility` page sometime soon as well so it would be nice to have it all together.

daKmoR:
> is "steppable" a good mixin example?

I think a stepper is already too complex to learn what mixins do... however from the top of my head I also can't think of anything simple 🙈  maybe just a simple alert or so... hmm no clue

=> to have a more complete example as a stackblitz is perfect :)

When I think about mixins I still remember this codelab https://codelabs.developers.google.com/codelabs/polymer-2-carousel/index.html?index=..%2F..index#4

> is this too LitElement-specific?

nah Mixins can be applied to every js class... => oh maybe that's a better example just show how you can mixin multiple function with "plain" classes.
=> and then map this knowledge to webcomponents?

> when should I use a mixin vs when should I create an element?

that is probably really tough to answer 🙈 I think there is no simple rule for this... that really depends on your use-case => so it's probably best to give a few good examples for mixins.

> Go into more depth on how `LitElement` can consume two differen `static get properties()`

that is "black magic"... nah not really it's just that they run the init of properties on every prototype... so technically the property exists only in the mixin but because of inheritance you can access it down the "tree"
=> I personally think that a simple `super.properties` and a single run at the resulting object would be nicer... and we have proposed this to lit-element but polymer seemed to like their "black magic" approach more [I don't know their reasoning - maybe this might better integrate with decorators? maybe not - no clue]
```
static get properties { 
  return { 
     ...super.properties, 
     addedProperty: { type: String }
  }
}
```

> Go into how to test these

uh yeah that's a good one :) => for a detail section it might be even interesting to do a "full story" e.g. start with concept => tests => implementation... like how it should be in a perfect word 🙈 


and a giant thank you for starting this 🤗

ZempTime:
I've been doing a little more thinking here. Since we haven't arrived at a clear-cut "here's a great example of a mixin," then I think there isn't one. Rather, mixins are a tool with multiple uses. This makes sense as there really aren't that many ways to share code between 'classes' in js. So truly good and useful documentation would account for the breadth of potential uses, but have special emphasis and curation on what's been practically useful in our libraries.

*Other Language Examples*:
* Scala (Lars)
* Java (Lars)
* Ruby Modules (Me)
* Dart (Justin)

So what would make sense for this kind of problem is a survey of the dimensions to consider, and then links out to canonical or illustrative examples of each.
*Capabilities List*:
* *Enforcing an Interface* -> (see below)
* *Layering* -> Sometimes you can minimize the implementation details of your components and centralize specific layers of responsibility. Primary example: [pwa-helpers connect mixin](https://github.com/Polymer/pwa-helpers/blob/master/src/connect-mixin.ts)
* *Concerns* -> sharing common capabilities and behaviors across components Primary example: the one already inside this pr.
* *Internal Composition* -> the means by which shared building blocks are composed into a resulting component. Primary example -> something somewhere `Vaadin`

Additionally, I found a [great article](http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/) by none other than Justin Fagnani. I'm thinking this should definitely be both linked to, and cherry-picked from. 

> A mixin is an abstract subclass; i.e. a subclass definition that may be applied to different superclasses to create a related family of modified classes.
> Gilad Bracha and William Cook, Mixin-based Inheritance

If we can roughly agree on an initial capabilities list, I'll go ahead and get an update out that follows roughly this new structure:

* What mixins are
* Mixins in other languages
* Mixins with web components (the capabilities list)
* Further Reading

The other task here would be:
- [ ] initial capabilities list
- [ ] update & sync stackblitz

PR #155, opened 2019-01-26
feat(dev-server): added dev server

LarsDenBakker:
No global docs changes, I want to do that as a bigger MR of updating the building/development docs.

PR #158, opened 2019-01-26
WIP: Update docs on dev and building

LarsDenBakker:
Extends https://github.com/open-wc/open-wc/pull/155. Wait until that is merged.


daKmoR:
just a small tip if you do a complete separate change it's best to create a branch from master.
This branch now is based off dev-server...
so if we do any force pushes or squashes or bigger changes to dev-server before  merging then it can become quite tough to do a rebase master for this branch...
nothing you can't overcome but it's easier to just avoid these kind of things :)

LarsDenBakker:
I branched off because I need the changes in the other branch. I will do a reset after it is merged.

daKmoR:
ah ok sure in that case it makes sense 👍

PR #165, opened 2019-01-27
Chore: Remove comments on build

anoblet:


anoblet:
I'm having a hard time with CI. I'm using Windows/Atom and the prettier erros might be a line-break issue?

web-padawan:
Please note that Webpack by default uses `terser-webpack-plugin` as a minifier since 4.27
Terser is a successor of UglifyJS with better support for ES2015+.

Here is how the relevant part of the config might look like:
```js
    optimization: {
      minimizer: [
        new TerserWebpackPlugin({
          terserOptions: {
            output: {
              comments: false
            }
          }
        })
      ]
    },
```

LarsDenBakker:
Thanks for the contribution. Indeed lets go with Terser. My only worry is whether removing license comments poses any legal risks. Does anyone know?

thepassle:
Looks like removing comments saves about 6kb (33.6kb to 27.5kb) on a demo app, but I'm not sure on the legalities of it.

There is this [stackoverflow comment](https://stackoverflow.com/questions/53476595/how-to-minify-lit-html-including-comments) by the author of hyperHTML: 

> remember that your site, app, project, that is using third parts software, must include such software license in a way or another.

As for the failing build; I think you forgot to require `const UglifyJsPlugin = require('uglifyjs-webpack-plugin');`. Though if we decide to go for this, we should probably use `terser-webpack-plugin` like web-padawan mentioned.

anoblet:
Whoops! I'll write a pull request with terser.

LarsDenBakker:
Wow that's a lot of saved kbs!

thepassle:
@anoblet Could you keep your changes to 1 PR? We have 4 PRs for the same issue right now 😅

anoblet:
Sorry about that.

anoblet:
I may need a little more practice with pull requests, as well as getting the linting setup correctly for the project.

daKmoR:
usually, the simplest is to just keep working in one branch (where you mad the PR from).
You can add new commits (or even force push to rewrite the history of your branch) and the PR will get automatically updated 🤗

daKmoR:
all changes have been moved to https://github.com/open-wc/open-wc/pull/168

PR #168, opened 2019-01-28
fix(building-webpack): remove comments on build

anoblet:


daKmoR:
So I merged all changes here and pushed into your branch.

Results:
before: 31.8 KB
after: 27.0 KB

I did some small adjustments:
- moved TerserPlugin dependency to `packages/building-webpack/package.json`
- provided extract option to extract comments (e.g. to sadisfy licencing)
- added licencing info to readme

this is under your name - if you could pls give your ok (simply reply ok 🤗) then I think we can merge this :)

daKmoR:

PR #171, opened 2019-01-29
fix: change year

newbornfrontender:


daKmoR:
hey,

thx for taking a look at it :)

it currently says
```
Copyright © 2018-present open-wc
```
I would assume that is 2018 till now...

or would you write it differently maybe?
```
Copyright © 2018-2019 open-wc
Copyright © 2018,2019 open-wc
Copyright © 2019 open-wc
```

as copyright takes a very very long time (70+ years) to expire I think we should be save with just sticking with 2018 or the 2018 - present

what do you think?

newbornfrontender:
Oh... Sorry, confused.

`Copyright © 2018-2019 open-wc` looks very good.

daKmoR:
thxxxxx ❤️

PR #173, opened 2019-01-30
fix(generator-open-wc): add building-webpack generator

thepassle:


packages/generator-open-wc/generators/building-webpack/templates/static/index.html
```diff
@@ -0,0 +1,7 @@
+<!doctype html>
+<html>
+  <head></head>
+  <body>
+    <my-app></my-app>
```
thepassle: Is there a way to make this less hardcoded? @daKmoR

packages/generator-open-wc/generators/building-webpack/templates/static/index.js
```diff
@@ -0,0 +1,5 @@
+import loadPolyfills from '@open-wc/polyfills-loader';
+
+loadPolyfills().then(() => {
+  import('./my-app.js');
```
thepassle: Same question for this

packages/generator-open-wc/generators/building-webpack/templates/static/index.html
daKmoR: sure that would be a scaffold generator (take a look at the testing one)... so this also means that this should probably be split in two generator :p

thepassle:
Not really sure how to split this up as two separate `scaffold-building` and `building-webpack`, for `building-webpack` you need an index.html and index.js, so then what would be the difference between `scaffold-building` and `building-webpack`? should we only make `scaffold-building` instead then?

daKmoR:
`open-wc:building` runs
- `open-wc:building-webpack`
=> adopt package.json (adds scripts + dependencies)  
=> create webpack config file
e.g. modifies and creates new files (which you usally do not have in an existing project)

`open-wc:scaffold-building` runs
- `open-wc:building`
=> adds index.js
e.g. creates new files (which you usually have in an existing project)

packages/generator-open-wc/generators/building-webpack/templates/_package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "scripts": {
+    "start": "webpack-dev-server --mode development --open",
```
daKmoR: I think for your generators we will need to have a concept for all the scripts we will offer in the end
```
start => should be owc-dev-server? (in the end - not part of this generator)
start:build, start:building? => webpack-dev-server?
start:dist, start:built, start:constructed? => serving the dist (build vs built?)
```

I think I would go with `start:build`, `start:dist`  in this generator (+ rest stays as is)

thepassle: what about `"start:dev": "webpack-dev-server --mode development --open"` and `"start:prod": "http-server dist/ -o"`?

daKmoR: uh `start:prod` is nice 👍 

`start:dev` would suggest that just `start` is not for development...
`start:build` is sort of start the building server

but I have no super strong opinion about it...

thepassle: Lets go with this for now?
```
    "start:dev": "webpack-dev-server --mode development --open",
    "start:dev:es5": "webpack-dev-server --mode development --es5",
    "start:prod": "http-server dist/ -o",
```

Issue #140, opened 2019-01-20
[building-webpack] Consider removing async function transformation


LarsDenBakker:
In our webpack setup we use the `fast-async` plugin to compile async functions. However that makes it always kick in. Based on: https://caniuse.com/#search=async%20function any browser which supports `type="module"` supports async functions. 

Probably we can remove this transform, but we should properly test this.

thepassle:
I tested this by removing:
`@open-wc/building-webpack/default-config.js:131`
```
'module:fast-async',
```

and adding:
```
async asyncFunction() {
  await new Promise(resolve => setTimeout(resolve, 500));
}
asyncFunction();
```

to my demo app.

I tested on the following browsers (on browserstack):
- firefox 62 and up
- chrome 69 and up
- opera 55 and up
- edge 15 and up
- safari 10.1

where everything works.

Bundlesize savings:

with fast-async
`27.9 kb`

without fast-async
`27.4 kb`

LarsDenBakker:
Thats great, thanks for testing! Performance is probably better too.

thepassle:
Very slightly, yeah

PR #180, opened 2019-01-31
fix: remove fast-async

thepassle:
closes https://github.com/open-wc/open-wc/issues/140


LarsDenBakker:
This removes it from the es5 build too, right? Need to figure out how to split it

thepassle:
Oh woops, my bad

PR #181, opened 2019-02-01
Adds karma-source-map-support plugin

arikimolloy:
https://github.com/tschaub/karma-source-map-support


daKmoR:
thaaankk you - that really improved the workflow via terminal ❤️

PR #182, opened 2019-02-01
docs(building-webpack): Add some notes to avoid user error

arthurevans:
Added notes to avoid problems I ran into as a Polymer CLI user using webpack for the first time.

Also moved the npm scripts to the last step. When I hit this part of the instructions, I immediately tried running the scripts, which failed because the necessary HTML and JS files weren't set up yet. Moving this step to the end prevents this user error.

Changed one header to make it more descriptive.

I'd suggest also adding a note (either in Step 4 above or in Change the default entrypoints) describing what to do if you have more than one HTML/JS entrypoint. I'd have done that but for my complete ignorance (Can indexJS take an array? Or do they need to go to the method described in "Customizing the configuration"? Can you pass a config object with indexHTML/indexJS to createDefaultConfig?)

### DIFF
--- a/packages/building-webpack/README.md
+++ b/packages/building-webpack/README.md
@@ -16,7 +16,7 @@ To set up the default configuration manually:
 
 1. Install the required dependencies:
 ```bash
-npm i -D @open-wc/building-webpack webpack webpack-dev-server http-server
+npm i -D @open-wc/building-webpack webpack webpack-cli webpack-dev-server http-server
 ```
 
 2. Create a file `webpack.config.js`:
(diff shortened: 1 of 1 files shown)


arthurevans:
Hi. CLA requirement was not listed on the contributing page. Going to have to pull this for now.

daKmoR:
oh too bad 😭 
is there anything we can do to improve this experience? (besides adding it to the contribution info)

it's the simples CLA we could find - is it too complicated?

There are multiple resources showing the pros and cons of CLAs... 
- not having one might leave you in this mambo jamo state... so we thought picking a very simple one would not do any harm... 
- heaving one might mean less contributions :/ 
- still there are many many project that don't have one... and you really seldomly hear of problems... but if they occure it's usually bad...

Beeing this the first contribution that's being pulled because of the CLA... we are interesting in hearing your opinion.

Sorry for the "complications" 🙈

arthurevans:
Not your problem, just some bureaucratic hoops I need to jump through before submitting a patch.

PR #186, opened 2019-02-02
fix(building-webpack): improve setup docu to minify user errors

arthurevans:
Added notes to avoid problems I ran into as a Polymer CLI user using webpack for the first time.

Also moved the npm scripts to the last step. When I hit this part of the instructions, I immediately tried running the scripts, which failed because the necessary HTML and JS files weren't set up yet. Moving this step to the end prevents this user error.

Changed one header to make it more descriptive.

I'd suggest also adding a note (either in Step 4 above or in Change the default entrypoints) describing what to do if you have more than one HTML/JS entrypoint. I'd have done that but for my complete ignorance (Can indexJS take an array? Or do they need to go to the method described in "Customizing the configuration"? Can you pass a config object with indexHTML/indexJS to createDefaultConfig?)

### DIFF
--- a/packages/building-webpack/README.md
+++ b/packages/building-webpack/README.md
@@ -10,7 +10,7 @@ To set up the default configuration manually:
 
 1. Install the required dependencies:
 ```bash
-npm i -D @open-wc/building-webpack webpack webpack-dev-server http-server
+npm i -D @open-wc/building-webpack webpack webpack-cli webpack-dev-server http-server
 ```
 
 2. Create a file `webpack.config.js`:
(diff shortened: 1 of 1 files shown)


packages/building-webpack/README.md
```diff
@@ -50,7 +37,15 @@ module.exports = defaultConfig();
 </html>
 ```
 
-5. Create a `index.js` which kickstarts your JS code.
+Note that the `index.html` file doesn't reference the JavaScript file—webpack will add a `<script>` tag when it builds your project. Also note that webpack does **not** process inline script tags, like:
+
+```html
+<script>
+  import {MyApp} from './src/my-app.js';
```
daKmoR: nit pick
```suggestion
  import { MyApp } from './src/my-app.js';
```

packages/building-webpack/README.md
```diff
@@ -50,7 +37,15 @@ module.exports = defaultConfig();
...
+</script>
+```
+
+4. Create a `index.js` which kickstarts your JS code.
 For older browser we need to load the web component polyfills. We use the [polyfills-loader](https://open-wc.org/building/polyfills-loader.html) for that:
```
thepassle:
```suggestion
For older browsers we need to load the web component polyfills. We use the [polyfills-loader](https://open-wc.org/building/polyfills-loader.html) for that:
```

Issue #172, opened 2019-01-30
[Testing] Testing Stencil Web Component


vogloblinsky:
Hi,

I am benchmarking many solutions for testing a simple Web Component created with Stencil.

Did you already test this kind of component with open-wc ?

I have tested :
- webdriver.io : facing shadow Dom querying issues with Selenium + IE11. Seems to be a complicated way to achieve e2e testing for WC.
- Web Component tester : having difficulties with test-fixture package. Seems to not be very stable, the root README is not up to date. Also have difficulties with IE11.

Thanks

daKmoR:
As Stencil build outputs pure native webcomponents you can definitely test these build outputs.
You will, however, need to handle the triggering of the stencil build yourself.

Stencil has a recommended way of testing if you prefer a more intergrated solution https://stenciljs.com/docs/unit-testing

vogloblinsky:
Hi,

Ok just started with your test tooling, seems very promising :+1: 

Regards

vogloblinsky:
First result are good, but i have an issue testing it. 

Here is my setup : 
![image](https://user-images.githubusercontent.com/2841805/52055256-2b4b8080-255f-11e9-9820-41f73e1cb78d.png)

The dist folder contains the generated Web Component with Stencil, this one : https://github.com/vogloblinsky/nutrition-web-components/tree/develop/packages/nova

Here is my test file :

```
import { expect, assert, fixture } from '@open-wc/testing';

import { nextFrame } from '@open-wc/testing-helpers';

import { defineCustomElements } from '../dist/esm/es2017/nova.define.js';
defineCustomElements(window);

describe('<nova-badge>', () => {
  it('Rendered by Stencil', async () => {
    const el = await fixture('<nova-badge></nova-badge>');
    await nextFrame();
    expect(el.getAttribute('class')).to.equal('hydrated');
    const scoreElement = el.shadowRoot.querySelector('.score');
    assert.notEqual(scoreElement.getAttribute('class').indexOf('one'), -1);
  });
  it('should have the nova badge score to 1', async () => {
    const el = await fixture('<nova-badge></nova-badge>');
    await nextFrame();
    const scoreElement = el.shadowRoot.querySelector('.score');
    assert.equal(scoreElement.textContent, "1");
  });
});
```

Each "it" works fine independently, but grouped not. I got this error, where am i wrong ?
Thanks

```
[...]

PR #190, opened 2019-02-03
fix(testing-helpers): add `await elementUpdated(el)` supports stencil

daKmoR:
closes https://github.com/open-wc/open-wc/issues/172

### DIFF
--- a/packages/testing-helpers/README.md
+++ b/packages/testing-helpers/README.md
@@ -64,18 +64,37 @@ expect(el.bar).to.equal('baz');
 ```
 
 ## Timings
-If you need to wait for multiple elements to update you can use `nextFrame`.  
+By default fixture awaits the elements "update complete" Promise.
+- for [lit-element](https://github.com/polymer/lit-element) that is `el.updateComplete`;
+- for [stencil](https://github.com/ionic-team/stencil/) that is `el.componentOnReady()`;
 
+If none of those specfic Promise hooks are found, it will wait for one frame via `await nextFrame()`.<br>
+**Note**: this does not guarantee that the element is done rendering - it just waits for the next JavaScript tick.
+
+Essentially, `fixture` creates a synchronous fixture, then waits for the element to finish updating, checking `updateComplete` first, then falling back to `componentReady()`, and `nextFrame()` as a last resort.
+
+This way, you can write your tests more succinctly, without having to explicitly `await` those hooks yourself.
 ```js
-import { nextFrame, aTimeout, html, fixture } from '@open-wc/testing-helpers';
-
 const el = await fixture(html`<my-el .foo=${'bar'}></my-el>`);
 expect(el.foo).to.equal('bar');
-el.foo = 'baz';
+
+// vs
+
+const el = fixtureSync(html`<my-el .foo=${'bar'}></my-el>`);
+await elementUpdated(el);
+expect(el.foo).to.equal('bar');
+```
+
+### nextFrame
+Uses `requestAnimationFrame` to wait for the next frame.
+```js
 await nextFrame();
-// or as an alternative us timeout
-// await aTimeout(10); // would wait 10ms
-expect(el.shadowRoot.querySelector('#foo').innerText).to.equal('baz');
+```
+
+### aTimeout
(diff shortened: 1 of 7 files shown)


packages/testing-helpers/README.md
```diff
@@ -64,20 +64,39 @@ expect(el.bar).to.equal('baz');
...
 
 ## Timings
-If you need to wait for multiple elements to update you can use `nextFrame`.  
+By default fixture awaits the elements "update complete" Promise.
+- for [lit-element](https://github.com/polymer/lit-element) that is `el.updateComplete`;
+- for [stencil](https://github.com/ionic-team/stencil/) that is `el.componentOnReady()`;
 
-```js
-import { nextFrame, aTimeout, html, fixture } from '@open-wc/testing-helpers';
+If non of those specfic Promise hooks is found it will wait for one frame via `await nextTrame()`.
```
bennypowers: s/non/none/
s/nextTrame/nextFrame/

packages/testing-helpers/README.md
```diff
@@ -64,20 +64,39 @@ expect(el.bar).to.equal('baz');
...
+Note that this is **NO** guarantee that the element is done rendering - it's "just" the next "frame".
```
bennypowers:
```
**Note**: this does not guarantee that the element is done rendering - it just waits for the next JavaScript tick.
```

packages/testing-helpers/README.md
```diff
@@ -64,20 +64,39 @@ expect(el.bar).to.equal('baz');
...
 
-const el = await fixture(html`<my-el .foo=${'bar'}></my-el>`);
+Basically fixture does that for you
```
bennypowers:
```
Essentially, `fixture` creates a synchronous fixture, then waits for the element to finish updating, checking `updateComplete` first, then falling back to `componentReady()`, and `nextFrame()` as a last resort.

This way, you can write your tests more succinctly, without having to explicitly `await` those hooks yourself.
```

packages/testing-helpers/README.md
```diff
@@ -64,20 +64,39 @@ expect(el.bar).to.equal('baz');
...
+```js
+const el = fixtureSync(html`<my-el .foo=${'bar'}></my-el>`);
+await elementUpdated(el);
+// which does the following:
```
bennypowers: Remove. put text in the text body, code in the snippet

packages/testing-helpers/README.md
```diff
@@ -64,20 +64,39 @@ expect(el.bar).to.equal('baz');
...
+// await el.updateComplete;
+// OR await el.componentReady();
+// OR await nextFrame();
 expect(el.foo).to.equal('bar');
-el.foo = 'baz';
+// so you can instantly start testing without doing the awaits yourself
```
bennypowers: remove

packages/testing-helpers/README.md
```diff
@@ -64,20 +64,39 @@ expect(el.bar).to.equal('baz');
...
+```
+
+### nextFrame
+Wait for the next frame via `requestAnimation`;
```
bennypowers:
```
Uses `requestAnimationFrame` to wait for the next frame.

File: packages/testing-helpers/elementUpdated.js [hunk 7]
```diff
@@ -0,0 +1,33 @@
+import { nextFrame } from './helpers.js';
+
+const isDefinedPromise = action => typeof action === 'object' && Promise.resolve(action) === action;
+
+/**
+ * Awaits for "update complete promises" of elements
+ * - updateComplete [lit-element]
+ * - componentOnReady() [stencil]
+ *
```
bennypowers: maybe add `GluonElement#render`

daKmoR: is that an update complete Promise? if so we could add it but pls in separate PR

packages/testing-helpers/README.md
thepassle:
```suggestion
If none of these specific Promise hooks are found, it will wait for one frame via `await nextFrame()`.
```

PR #195, opened 2019-02-05
fix(testing-helpers): integration test for stencil fixture

daKmoR:


daKmoR:
because of double transpilation this is probably not worth it - we know it work from users in the wild :)

PR #196, opened 2019-02-05
docs(ide): replaced es6-string-html by lit-html

odahcam:
Just a proposal, maybe it isn't better at all, but that you will say.

I use this extension daily to develop Polymer's LitElements and I like it very much.


daKmoR:
yes it's the most used one and does a fine job... we may need to reevaluate this at some point but for now it's probably the best choice :) thank you :)

odahcam:
You're welcome, thanks for all the good job here.

PR #208, opened 2019-02-11
fix(owc-dev-server): fix CLI options table formatting in README

vlukashov:
before and after


daKmoR:

PR #219, opened 2019-02-13
fix: dash to camelcase

thepassle:
A few minor changes:

- The generator would break because it didn't convert to camelcase for the classname
- Fixed readme

I think the solution for the double `index.html` is to just... remove it. I feel like it might cause confusion having 2 ways to start a dev server, but open for discussion

(sorry, i did it all in 1 commit 🙈 )


daKmoR:
imho `owc-dev-server` should be the default start... it's the most pure situation and brings the most "joy" (at least for me)

for the double index.html we need a different solution so we can have certain parts in there that are "ignored" by the webpack... I'm pretty sure we can make it work until then 2 files are needed :/

thepassle:
@daKmoR PTAL

packages/generator-open-wc/generators/starter-app/templates/static/index.html
```diff
@@ -3,10 +3,8 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
-  <meta name="description" content="Put your description here.">
-
-  <script type="module" src="./src/<%= tagName %>.js"></script>
-
+  <meta name="Description" content="Put your description here.">
+  <script type="module" src="/src/index.js"></script>
```
daKmoR:
```suggestion
  <script type="module" src="./src/index.js"></script>
```

thepassle:
Considering the new generators have a barebones generator now, maybe we can remove it from the starter-app generator?

daKmoR:
with the new generators this will be way different... pls reopen a new one if needed 🤗

PR #227, opened 2019-02-16
fix(owc-dev-server): read and apply .owc-dev-server.config.js

daKmoR:


LarsDenBakker:
Should we make proxy a command line option?

daKmoR:
this is highly inspired by `create-react-app` and they had an option before and replaced it with a node config file... as usually, you need various options if you set up a proxy or anything specific... so adding options will probably never cut it

here are the options for https://github.com/chimurai/http-proxy-middleware
but we probably don't even want to force a specific proxy middleware... if you want to use a different middleware go ahead just add it to your `.owc-dev-server.confg.js` like so
```
const proxy = require('other-proxy-service');

module.exports = function(app) {
  app.use(proxy({ other-config-options: true });
});
```

do we want to add this "rational" to the readme as well?

syu93:
In my opinion, people really don't like when configurable thing are hidden, so adding those to the readme make sense

daKmoR:
@syu93 I added this to the readme... would you expect more?
https://github.com/open-wc/open-wc/pull/227/files#diff-936809f726fb7ed5235b7c4a3ef061a2R85

syu93:
I guess I haven't used enough the owc dev server to tell, but we should be flexible about what came back from the community and improve the document as we go

Issue #224, opened 2019-02-14
update all package.json to follow npm suggestion for monorepos


daKmoR:
basically just split url to git and path to package

Simple example see https://github.com/npm/cli/commit/3663cdef205fa9ba2c2830e5ef7ceeb31c30298c

daKmoR:
fixed by @anoblet thx 🤗

PR #230, opened 2019-02-16
fix: update package repository fields

anoblet:
https://github.com/open-wc/open-wc/issues/224


daKmoR:
uuuhhh awesome thxxxxx ❤️

PR #233, opened 2019-02-18
docs: add recommendation to separate class and element definitions

### LINKED ISSUE #232: Add rational for splitting class with registration
Could be something like this:

We recommend to split class definition and custom element registration in 2 files to enable your users to extend your class without needing to pollute the custom elements registration.

Example:
foo-foo/FooFoo.js // class
foo-foo/foo-foo.js // custom element registration

I can now extends
bar-bar/BarBar.js
import { FooFoo } from 'foo-foo';
class BarBar extend FooFoo {}

bar-bar/bar-bar.js

So you can now import bar-bar without polluting the registry with the foo-foo tag

bennypowers:
Fixes #232

### DIFF
--- a/docs/developing/README.md
+++ b/docs/developing/README.md
@@ -28,22 +28,26 @@ npm i -D owc-dev-server lit-element
 </html>
 ```
 
-3. Create a `my-component.js`
+3. Create a `MyComponent.js`
 ```javascript
 import { LitElement, html } from 'lit-element';
 
-class MyComponent extends LitElement {
+export class MyComponent extends LitElement {
   render() {
     return html`
       <p>Hello world!</p>
     `;
   }
 }
+```
 
+4. Create a 'my-component.js'
+```javascript
+import { MyComponent } from './MyComponent.js';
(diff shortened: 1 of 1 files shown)


docs/developing/README.md
```diff
@@ -52,11 +56,35 @@ customElements.define('my-component', MyComponent);
...
+
+    return html`
+      ${super.render()}
+      <style>p { color: red; }</style>
+      <p>${alert}</p>
+    `
+  }
+})
+```
+
```
daKmoR:
```suggestion
Note: this only makes sense if you expect your web component to be used by others. If you have very app specific components it's probably easier to skip this separation.
Note: a specific usecase would be for example an app which uses an auto registration of custom elements like [careHtml](https://www.npmjs.com/package/carehtml) then they will always be only interested in the class.

```

LarsDenBakker: I would reverse the suggestion personally. So for example say, 'if you are writing a public component which will be used by others, it is a good idea to split the definition and the class'.

PR #234, opened 2019-02-18
Bump lit versions to the 1.0 stable releases

stramel:
lit-html bumped to 1.0.0
lit-element bumped to 2.0.1 (That was the version which uses 1.0.0 of lit-html)


daKmoR:
thxxxx :heart:

Issue #245, opened 2019-02-21
Vulnerability Found


Westbrook:
Run `yarn audit` and you get the following:

![image](https://user-images.githubusercontent.com/1156657/53174990-12ac1480-35b9-11e9-8d3f-227b7c1972ff.png)

I'm sure I'm doing something wrong, but I'm pretty sure that this is preventing me from publishing work to Netlify. 😖 

Can look to make a PR later, but must head to the office... 😢

Westbrook:
Correcting PR has been merged at the Karma level: https://github.com/karma-runner/karma/issues/3269 actual release pending...

daKmoR:
with the release of karma 4.0.1 this should be solved 🤗 

pls reopen if this is still an issue

Issue #246, opened 2019-02-21
webpack.config.js does not pass commit linting


Westbrook:
The following lint errors occur when trying to commit webpack.config.js. Not sure if there's something requiring this, so they should use some sort of `ignore` commenting, or just a bunch of removals to be made here.

![image](https://user-images.githubusercontent.com/1156657/53175143-5f8feb00-35b9-11e9-9bbb-6c2da6b21821.png)

Will try to confirm later and put together a PR.

daKmoR:
yes we probably should add `**/*.config.js` here: https://github.com/open-wc/open-wc/blob/master/packages/eslint-config/index.js#L15

that way you can use devDependencies in these config files 👍

Westbrook:
Cool, I can follow up with that PR later tonight or tomorrow.

Westbrook:
Closed via #249

PR #249, opened 2019-02-23
Allow config files to depend on devDependencies.

Westbrook:
Fixes: #245


Westbrook:
Sorry, just realized I linked the wrong issue here...

Fixes #246

PR #250, opened 2019-02-24
Merge Strategy Documentation

### LINKED ISSUE #226: [Testing] Coverage Reporting
I'd like to suggest an update to the default reporting style from `text-summary` to `text`.

The following image outlines the output differences visually, but functionally the ability to see coverage by individual files along with the lines in those files that would benefit from further testing is very useful when you are looking to ensure a specific level of test coverage in your work.

image

At current, the steps to alter a generated project to support this output requires either relying on the current `merge` technique available at https://github.com/open-wc/open-wc/blob/master/packages/generator-open-wc/generators/testing-karma/templates/static/karma.conf.js#L13 which would cause you to receive both versions of the report, or do something like:

```
  let customConfig = defaultSettings(config);
  customConfig.coverageIstanbulReporter.reports = ['html', 'lcovonly', 'text'];
  config.set(
    merge(customConfig, {
```
To hack the config before merging it with other local settings.

I could see a mix of alternate approaches to solving this:
- Not setting reporting by default
  - which would directly preface the need for the following
- Better config extension documentation
  - would be particularly useful if there are other things in the config that would benefit from being "replaced", but I've not seen it.
- A variant set of configs to alter the point of control in manage these settings (i.e. `default-settings`, `default-with-text`, `default-with-text-summery`, etc)
  - might lead to "if you give a mouse a cookie" syndrome, but could be externalized through community defaults if it became of interest at a larger scale

However, all of the alternate approach involve knowing a little more about usage that I'm not sure has happened at the scale necessary to support going down one of those paths at this time.

I think this fairly simple change would provide a good amount of benefit, and I'd be happy to supply a PR if the team agrees.

Westbrook:
fixes #226

I chose the merge strategy approach namely because I'm not sure how to plumb the depths of possible issues around @daKmoR's note that:

> iirc using only config set however breaks webpack configs as they need to be "carefully" copied with keeping function in tact and such... that's why we use webpack-merge generally

If anyone else has deeper insight on this, I'd be happy to make updates here.

### DIFF
--- a/packages/testing-karma/README.md
+++ b/packages/testing-karma/README.md
@@ -64,6 +64,33 @@ For a minimal setup, extend the base config and specify where your tests are:
     return config;
   };
   ```
+   <details>
+    <summary>Replacing Default Settings Instead of Merging</summary>
+    <br/>
+    In some cases you'll want `your custom config` to include config values that replace, rather than extend, the defaults provided. To make this possible you can make advanced usage of webpack-merge to set a [merge strategy](https://github.com/survivejs/webpack-merge#mergestrategy-field-prependappendreplaceconfiguration-configuration) to follow when joining the defaults and your custom config. See below for an example that uses `replace` to change the `reports` used by `coverageIstanbulReporter`.
+  
+    ```js
+    module.exports = config => {
+      config.set(
+        merge.strategy(
+          {
+            'coverageIstanbulReporter.reports': 'replace',
+          }
+        )(defaultSettings(config), {
+          files: [
+            // allows running single tests with the --grep flag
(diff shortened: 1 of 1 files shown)


daKmoR:
hmm strange in the preview the markdown seem to not be correctly rendered?

https://deploy-preview-250--nervous-stonebraker-1f429b.netlify.com/testing/testing-karma.html#manual-setup

image

Westbrook:
Oh, indeed. Tweaked it and I think the parse likes me again...

Westbrook:
image

daKmoR:
thxxxxx :heart:

Issue #221, opened 2019-02-13
[building] consider rollup


LarsDenBakker:
I'd like to look into a rollup based default config. I think we need to have measurable benefits from using rollup VS webpack, and give different recommendations for different scenarios.

I'd like to open a discussion in this issue about this.

Some areas I am interested in:

**Module output**
This is one of the obvious benefits, but how practical is it? Not all browsers which support modules, support dynamic imports. How do you make sure you distribute the right code?

**Code optimization**
Focusing on es-modules has the potential for better tree shaking, dead code elimination and minification (inlining variables etc) because it's easier for the bundler to take certain assumptions. I'd like to see if this can be measured objectively.

**Code splitting**
One thing that webpack does very well is code splitting. It is more complicated than it seems, as shared dependencies between lazy loaded chunks creates quite some coordination demands on the bundler. I'm interested in seeing how rollup stacks up against webpack in this area.

**Build tasks**
For larger apps building is more than just compiling the JS. Static assets need to be moved around, environment variables need to be set etc. How well does rollup handle these things?

**Caching**
Webpack handles caching, both short and long-term. For .js but also other assets like css and pngs. How does rollup handle this?

Does anyone have any insights in this area?

bennypowers:
> Module Output
Rollup can output es, IIFE, cjs, amd, system, umd. you're covered

> Code Splitting
Landed stable in Rollup 1.0. I use it in my apps, and developed `rollup-plugin-modulepreload` to help devs automatically preload chunks based on a predefined or  custom predicate

> Build Tasks
There are lots of plugins for this like `rollup-plugin-cpy`. npm scripts are also good for this

bennypowers:
I like rollup's plugin-based architecture.
I like rollup's focus on browser standards rather than kitchen-sinking.
I like the quality of rollup's generated code.

PR #251, opened 2019-02-24
feat(building-rollup): added building-rollup

LarsDenBakker:
Fixes https://github.com/open-wc/open-wc/issues/221

This is a draft PR for investigating and adding a rollup setup.

## Comparison with webpack
Working with rollup has been a much better experience than webpack. There is a lot less magic going on, and it is easy to grok what happens. It has definitely changed my opinion on the tool. 

The main area I was worried about is code splitting, but so far it's looking great. It's still quite a new feature for rollup, whereas webpack is more battle tested. There have been some bugs in the past around this, but I am confident it will improve.

The primary selling point for rollup is that it can output esm format, which webpack can't do right now. It really fits our standards-based approach. Shipping esm to production is complicated though, as browsers which don't support dynamic imports will still load `<script type="module">`. 

## TODO

- [x] Split current config into bullets 1. and 2.
- [x] Make the config work with watch
- [x] Split plugin into several plugins
- [x] Write docs
- [x] Support custom babelrc
- [x] Analyze code optimization (inlining, dead core removal, tree shaking etc.)
- [x] Add more code optimization, such as template literal minification @bennypowers maybe you can help here?


LarsDenBakker:
To test out the branch, clone it and run:

```
npm run build:demo
npm run start:build
```

bennypowers:
Rollup is life.

I've got some rollup plugins to aid app development:

npm.im/rollup-plugin-lit-css

npm.im/rollup-plugin-workbox

packages/building-rollup/default-config.js
```diff
@@ -0,0 +1,73 @@
...
+    },
+    plugins: [
+      indexHTML({
+        legacy,
+        polyfillWebComponents: legacy,
+      }),
+      // resolve bare import specifiers
+      // TODO: Check options
+      resolve(),
+      babel({
```
bennypowers: There's really no reason to run babel on a modern build. babel should be disabled in favour of terser for modern browsers, imo.

packages/building-rollup/demo/demo-app.js
```diff
@@ -0,0 +1,84 @@
...
+    this.addEventListener('foo-event', e => {
+      console.log('foo event fired through multiple shadow roots', e.composedPath());
+    });
+  }
+
+  render() {
+    return html`
+      <p>Hello world</p>
+
+      <button @click="${() => this._lazyLoad()}">Lazy load</button>
```
bennypowers:
```suggestion
      <button @click="${this._lazyLoad}">Lazy load</button>
```

packages/building-rollup/default-config.js
LarsDenBakker: That's debatable. If you want to use new syntax which isn't yet supported on all evergreens you still need babel to handle that for you. So it's question of what is the best default here.

not sure what babel has to do with terser though?

bennypowers: https://open-wc.org/about/rationales.html#rationales
> Language or platform features/APIs must be released without a flag in the stable version of a least one browser
> Note: We currently have ONE exception to this rule and that is 'bare modules'.

We can show how to add their own babel config, but 'babel all the things' as default for modern browsers is a mistake, in my opinion.

LarsDenBakker: If we don't 'babel all the things', it means that people won't be able to use new javascript features until they are supported by chrome, safari, firefox and edge. If only one browser supports it, you still can't ship it to production.

bennypowers: Right, I'm advocating that we recommend against using features that aren't widely available.

I also advocate that we show examples for how to opt in to babel transforms for when you do want to use features like private fields or decorators.

bennypowers: This is, of course, a matter of opinion, and i'm happy to defer to the group if it's decided that babel is always necessary..

but in my experience it's not, and babel transforms aren't free - they often add lots of overhead in terms of bundle size and runtime requirements.

LarsDenBakker: I understand what you're saying. I can go either way here, personally I prefer to use new syntax when it becomes part of the spec.

Transform size is a problem for older specs, like regenerator. I don't see this with newer specs but that that might come again with future specs. 

The question is what's the best default here.

LarsDenBakker:
I've greatly simplified the setup by utilizing the new dynamic import rename function available in rollup 1.4.0 which was released 13 hours ago :) https://github.com/rollup/rollup/releases

This way we can use es modules everywhere except IE11, and use a dynamic import polyfill for browsers that haven't implemented it yet.

Tomorrow I will work on docs and further optimization plugins.

packages/building-rollup/default-config.js
LarsDenBakker: we actually do need babel for the import meta transform, because you need the URLs from before the bundling rather than after

LarsDenBakker:
I've added a readme, still some points TODO but it is ready to review as it is now.

@daKmoR 
@thepassle

packages/building-rollup/README.md
```diff
@@ -0,0 +1,93 @@
...
+`modern-config.js` we setup above works for browsers which support modules and dynamic imports (Chrome 63+, Safari 11.1+, Firefox 67+).
+If you need to support other or older browsers, use our `modern-and-legacy-config.js` in your `rollup.config.js`:
+
+```javascript
+import createDefaultConfig from '@open-wc/building-rollup/modern-and-legacy-config';
+
+export default createDefaultConfig({ input: './index.html' });
+```
+
+In additional to outputting your app as a module, it also outputs a legacy build of your app and loads the appropriate version based on browser support. Depending on your app's own code, this will work on Chrome, Safari, Firefox, Edge and IE11.
```
thepassle:
```suggestion
In addition to outputting your app as a module, it also outputs a legacy build of your app and loads the appropriate version based on browser support. Depending on your app's own code, this will work on Chrome, Safari, Firefox, Edge and IE11.
```

packages/building-rollup/README.md
```diff
@@ -0,0 +1,93 @@
+# Rollup
+
+[//]: # (AUTO INSERT HEADER PREPUBLISH)
+
+## Default configuration
+We provide a good default configuration to help you get started with using rollup with web components and modules.
```
thepassle:
```suggestion
We provide a good default configuration to help you get started using rollup with web components and modules.
```

LarsDenBakker:
I want to postpone parsing `.babelrc` and supporting file hashing to a later MR.

LarsDenBakker:
I don't understand the linting error in the CI. I also get it locally when I check out master. Prettier should auto format these files, right?

PR #255, opened 2019-02-27
docs: add coverage config to karma testing example

bennypowers:


daKmoR:
cherry-picked and merged via https://github.com/open-wc/open-wc/pull/262

PR #261, opened 2019-03-02
feat: add demoing storybook

daKmoR:
Lots of awesome stuff :)

auto generating knobs could even be a replacement for api docs?

preview gif here:
https://github.com/open-wc/open-wc/blob/b59bbb002a962c2678e6040ad51a21ed45cbb449/packages/demoing-storybook/dev_assets/storybook.gif


packages/demoing-storybook/README.md
```diff
@@ -0,0 +1,172 @@
...
+    // show textarea where you can input json
+    { key: 'complexItems', fn: () => object('complexItems', el.complexItems, 'Inherited') },
+    // move property to a different group
+    { key: 'locked', group: 'Security' },
+  ],
+}),
+```
+
+By default it will create one simple node from the given Class.
+However for a nicer demo it may be needed to set properties or add more lightdom.
```
thepassle:
```suggestion
However for a nicer demo it may be necessary to set properties or add more lightdom.
```

packages/demoing-storybook/README.md
```diff
@@ -0,0 +1,172 @@
...
+  `,
+  { notes: { markdown: notes } },
+)
+```
+
+### Helper to setup transpilation for dependencies
+
+Whenever you add a dependency that is written in ES modules you will need to enable transpilation for it.
+Open your configuration and add the new package to the array.
+Below you see the default settings.
```
thepassle:
```suggestion
You can find the default settings below:
```

packages/demoing-storybook/demo/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+<style>
+  iframe {
+    width: 100%;
+    border: 1px solid rgba(0,0,0,.1);
+    min-height: 300px;
+  }
+</style>
+
+This example element display some information about...
```
thepassle: ?

packages/demoing-storybook/demo/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Features:
+- bigger then you would expect
```
thepassle:
```suggestion
- bigger than you would expect
```

packages/demoing-storybook/demo/my-el.js
```diff
@@ -0,0 +1,124 @@
...
+  constructor() {
+    super();
+    this.header = 'Default Header';
+    this.headerColor = '#ff0000';
+    this.disabled = false;
+    this.items = ['A', 'B', 'C'];
+    this.type = 'medium';
+    this.age = 10;
+    this.time = new Date();
+    // time needs to stay the same so storybook knows can work with them
```
thepassle:
```suggestion
    // time needs to stay the same so storybook knows it can work with them
```

packages/demoing-storybook/demo/README.md
daKmoR: this is just for the demo of storybook and just an example how a README for a component could look like... I think making a nice readme could be a task on it's own... and hopefully after storybook 5.0... there will be some focus on docs mode which may influence how a readme looks...

so don't heed too much about it now 🙈

thepassle: Ah I see, ty for explaining

PR #267, opened 2019-03-06
fix(testing-karma): exclude spec files in node_modules

bennypowers:
previously we only excluded `test` files.


daKmoR:
good catch 💪

PR #269, opened 2019-03-06
docs(demoing-storybook): add knobs import

blikblum:
In the overrides example, color, select and object are referenced but not imported nor cited in text.
Also add note about the custom element position in template


daKmoR:
thxxxxx ❤️

blikblum:
Thanks for your job. It allowed me to setup tests and storybook with minimal effort

PR #275, opened 2019-03-08
Make fixtures generic

Lodin:
Current Typescript definition doesn't allow setting the type of a result `fixture` returns. It is `Element` every time, so to use all methods and properties of the real element you have to do the tricky cast like the following. 
```typescript
const test = (await fixture(`<${tag}></${tag}>`)) as Test;
```
Unfortunately, if you enable all strict rules, you cannot go the more straightforward way.
```typescript
// Throws an error that type 'Element' is missing properties from type 'Test'
const test: Test = await fixture(`<${tag}></${tag}>`);
```
This PR permits to define the type of the result `fixture` returns. It makes `fixture` generic with the `Element` set by default and forbids using types that don't extend Element.
```typescript
const test = await fixture<Test>(`<${tag}></${tag}>`);
```


daKmoR:
hey, thank you for this nice idea :)

we wanted to go more in the direction of doing real type linting and providing better intellisense for our users... so we may get rid of those isolated `*.d.ts` files.

Here is the first try:
https://github.com/open-wc/open-wc/pull/277

Could you pls verify if this still allows for your use-case... and as my typescript/typing foo is not too great maybe also help with some of those type errors? 😬

Lodin:
@daKmoR, since we have some problems with #277, could we proceed with this PR until all jsdoc issues are solved? I see this change as a really useful one for the typescript testing.

daKmoR:
@Lodin indeed make sense to give value now :)

thxxxx for PR and reminder :)

PR #276, opened 2019-03-09
feat: add cli options

thepassle:
Adds options to the generator cli

### DIFF
--- a/packages/create/src/Generator.js
+++ b/packages/create/src/Generator.js
@@ -6,7 +6,7 @@ import { parseCliOptions } from './helpers.js';
 
 const optionDefinitions = [
   { name: 'tag-name', type: String, defaultValue: '' },
-  { name: 'no-npm', type: Boolean, defaultValue: false },
+  { name: 'npm-install', type: String, defaultValue: 'true' }, // set to emptry string later
   { name: 'scaffold', type: String, defaultValue: '' },
 ];
 
@@ -16,8 +16,8 @@ class Generator {
   constructor() {
     this._destinationPath = process.cwd();
     this.templateData = {};
+    this.wantsNpmInstall = '';
     this.cliOptions = parseCliOptions(cliOptions);
-    this._noNpm = this.cliOptions['no-npm'];
   }
 
   // eslint-disable-next-line class-methods-use-this
@@ -40,7 +40,13 @@ class Generator {
   }
 
   async end() {
-    if (!this._noNpm) {
+    if (this.cliOptions['npm-install'] === '') {
+      // this.wantsNpmInstall = await askYesNo('Do you want to run npm install?'); // eslint-disable-line
+    } else {
+      this.wantsNpmInstall = this.cliOptions['npm-install'];
+    }
+
+    if (this.wantsNpmInstall) {
       await installNpm(this._destinationPath);
     }
   }
--- a/packages/create/src/Generator.js
+++ b/packages/create/src/Generator.js
@@ -1,11 +1,23 @@
 import path from 'path';
+import commandLineArgs from 'command-line-args';
+
 import { copyTemplates, copyTemplate, copyTemplateJsonInto, installNpm } from './core.js';
+import { parseCliOptions } from './helpers.js';
+
+const optionDefinitions = [
+  { name: 'tag-name', type: String, defaultValue: '' },
+  { name: 'no-npm', type: Boolean, defaultValue: false },
+  { name: 'scaffold', type: String, defaultValue: '' },
+];
+
+export const cliOptions = commandLineArgs(optionDefinitions, { partial: true });
 
 class Generator {
   constructor() {
     this._destinationPath = process.cwd();
     this.templateData = {};
-    this._installNpm = true;
+    this.cliOptions = parseCliOptions(cliOptions);
+    this._noNpm = this.cliOptions['no-npm'];
   }
 
   // eslint-disable-next-line class-methods-use-this
@@ -28,7 +40,7 @@ class Generator {
   }
 
   async end() {
-    if (this._installNpm) {
+    if (!this._noNpm) {
       await installNpm(this._destinationPath);
     }
   }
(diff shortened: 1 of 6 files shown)


packages/create/src/helpers.js
```diff
@@ -1,4 +1,13 @@
 import qoa from 'qoa';
+import commandLineArgs from 'command-line-args';
+
+const optionDefinitions = [
+  { name: 'tag-name', type: String, defaultValue: '' },
+  { name: 'no-npm', type: Boolean, defaultValue: false },
+  { name: 'no-scaffold', type: Boolean, defaultValue: false },
+];
```
daKmoR: come to think of it I guess having it like this might be better
```
npm init @open-wc --tag-name foo-bar --npm no --scaffold yes
```

e.g. we may want to force in "both" direction...

packages/create/src/core.js
```diff
@@ -10,6 +10,8 @@ import deepmerge from 'deepmerge';
 
 import Generator from './Generator.js';
 
+import { cliOptions } from './helpers.js';
+
 const execAsync = util.promisify(exec);
```
daKmoR: core should stay pretty clear of specific logic (e.g we may want to extract it at some point)

probably best to move it to `create.js` and then pass it on like so
```
    const module = await import(fullGeneratorPath);
    await executeMixinGenerator(module.default, { cliOptions }); // you will need to make that option as well
```

that way should be able to access it and pass it along inside of the generator mixins

packages/create/src/helpers.js
thepassle: Alright, so:

`--no-npm` is just a boolean now. Its false by default, and when specified it wont npm install
`--no-scaffold` is a string. If not specified, will prompt the user. If specified as `--no-scaffold=true`, it will not prompt and not scaffold, if specified as `--no-scaffold=false` it will not prompt but it will scaffold.

packages/create/src/core.js
thepassle: I tried this, but I couldnt get it to work nicely like this, I instead set the cliOptions as a property on the Generator class

packages/create/src/Generator.js
```diff
@@ -1,11 +1,23 @@
 import path from 'path';
+import commandLineArgs from 'command-line-args';
+
 import { copyTemplates, copyTemplate, copyTemplateJsonInto, installNpm } from './core.js';
+import { parseCliOptions } from './helpers.js';
+
+const optionDefinitions = [
+  { name: 'tag-name', type: String, defaultValue: '' },
+  { name: 'no-npm', type: Boolean, defaultValue: false },
```
daKmoR: I would like to (later also) add a yes/no question for npm installation asa well... so let's use already use
```
{ name: 'install-npm', type: String, defaultValue: '' },
```

packages/create/src/helpers.js
```diff
@@ -1,20 +1,20 @@
...
 function getClassName(tagName) {
   return tagName
     .split('-')
     .reduce((previous, part) => previous + part.charAt(0).toUpperCase() + part.slice(1), '');
 }
 
+export function parseCliOptions(optns) {
+  Object.keys(optns).forEach(key => {
+    if (optns[key] === 'false') optns[key] = false; // eslint-disable-line no-param-reassign
+    if (optns[key] === 'true') optns[key] = true; // eslint-disable-line no-param-reassign
```
daKmoR: so usage is
```
npm init @open-wc scaffold=true
npm init @open-wc scaffold=false
```
?

thepassle: `npm init @open-wc --scaffold=true`

packages/create/src/Generator.js
thepassle: oki

thepassle: fixed, we just have to uncomment line 44 in `Generator.js` and set the default to `{ name: 'npm-install', type: String, defaultValue: '' },` later.

packages/create/src/Generator.js
```diff
@@ -1,11 +1,23 @@
...
+];
+
+export const cliOptions = commandLineArgs(optionDefinitions, { partial: true });
 
 class Generator {
   constructor() {
     this._destinationPath = process.cwd();
     this.templateData = {};
-    this._installNpm = true;
+    this.wantsNpmInstall = '';
```
daKmoR: shouldn't that be `true` instead of an empty string? 🤔 

also this option is set to a different value in `generators/app` pls update there as well 🤗

packages/create/src/helpers.js
daKmoR: we changed approach

packages/create/src/Generator.js
daKmoR: sound good 👍

packages/create/src/helpers.js
daKmoR: true/false sound good 🤗 
wanted just to verify 🤗

packages/create/src/Generator.js
thepassle: the command-line-args module doesnt seem to support having a default undefined value for booleans, so this is the only way to have 3 states/possibilities (default = `''`, force yes = `'true'`, force no = `'false'`)

packages/create/src/helpers.js
```diff
@@ -1,27 +1,46 @@
...
+          handle: 'tagName',
+        },
+      ]);
+      // eslint-disable-next-line prefer-destructuring
+      tagName = result.tagName;
+    } while (/^([a-z])(?!.*[<>])(?=.*-).+$/.test(tagName) === false);
+    className = getClassName(tagName);
+  } else {
+    tagName = cliOptions['tag-name'];
+    className = cliOptions['tag-name']
```
daKmoR: hmm can't make a suggestion that replaces 3 lines...

but that should be just
```
className = getClassName(tagName);
```
right?

thepassle: Good one, fixed

Issue #113, opened 2019-01-01
Type linting via tsc


daKmoR:
Also async is not allowed in definition files
https://stackoverflow.com/questions/48448151/how-can-i-add-an-async-function-to-a-type-def-file

daKmoR:
This is probably a very good starting point for it 🤗 

https://medium.com/@trukrs/type-safe-javascript-with-jsdoc-7a2a63209b76

daKmoR:
made some progress here https://github.com/open-wc/open-wc/pull/277

but there are still some problems with typescript... will need to raise an issue there

daKmoR:
release via https://github.com/open-wc/open-wc/pull/277

PR #277, opened 2019-03-09
feat: add types + linting & improve intellisense

daKmoR:
This will allow 
- for type linting (via typescript)
- better integration with typescript 
- to provides way better/nicer intellisense

Additional tasks done:
- only put needed files in published npm packages (so there will be no linting errors for demo folder or so)
- temporary forked dom5 (in building-rollup) as it has outdated type definitions (see [issues/3385](https://github.com/Polymer/tools/issues/3385))
- moved most testing-helpers files to `src` folder (only `index.js` & `index-no-side-effects.js` stay in root folder)

Note: adds a little time to CI as we now do type linting as well

Example:
better-intellisense

closes: https://github.com/open-wc/open-wc/issues/302, https://github.com/open-wc/open-wc/issues/113


daKmoR:
hmmm I have no clue why I can't find `lit-html` or `chai`
```
packages/testing-helpers/lit-html.js:1:33 - error TS2307: Cannot find module 'lit-html/lit-html.js'.

1 import { html as litHtml } from 'lit-html/lit-html.js';

packages/testing-helpers/test/elementUpdated.test.js:2:24 - error TS2307: Cannot find module '@bundled-es-modules/chai'.

2 import { expect } from '@bundled-es-modules/chai';
```

is there a special setting for the tsconfig you need for monorepos?

daKmoR:
ok setting moduleResolution to node helped 🤗

Lodin:
I just installed package created from your branch instead of the stable version, and now it refuses to show any type at all, and my type check ends without any issue even if I make a mistake on purpose. It looks like `fixture` is marked as `any`. I'm not sure what the reason, I use the following [`tsconfig.json`](https://github.com/corpusculejs/corpuscule/blob/97d06d5877cb63f874be5625f24033be861ebaaa/tsconfig.json):
```json
{
  "compilerOptions": {
    "allowJs": true,
    "allowSyntheticDefaultImports": true,
    "declaration": false,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "lib": [
      "dom",
      "es2018"
    ],
    "mapRoot": "dist",
    "module": "esnext",
    "moduleResolution": "node",
    "noImplicitAny": false,
    "noImplicitReturns": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "outDir": "dist",
    "sourceMap": true,
    "strict": true,
    "strictFunctionTypes": false,
    "stripInternal": true,
    "target": "esnext"
  },
  "include": [
    "packages/**/*"
  ],
  "exclude": [
    "packages/**/dist",
    "packages/**/node_modules",
    "node_modules"
  ]
}
```

daKmoR:
@Lodin thx for checking it out... I also made some additional tests and had the same experience 
https://github.com/daKmoR/jsdoc-intellisense/issues/1

I reached out to some others and it seems to be a "bug/feature" with typescript
https://github.com/daKmoR/jsdoc-intellisense/issues/1

so we probably need to investigate this further before we can actually do it 🙈

LarsDenBakker:
Thats really unfortunate

daKmoR:
sooo apparently that does work if you set 
```
--maxNodeModuleJsDepth 2
```

but that would mean everyone using this will need to set it... 
it leads to the nice intellisense and full type linting...

still it's a setting we need to teach everyone to set would we feel comfortable doing so?

daKmoR:
@Lodin may I ask you again to test this but this time with something like this as the `tsconfig.json`

```json
{
  "compilerOptions": {
    "target": "ES2015",
    "moduleResolution": "node",
    "lib": ["es2017", "dom"],
    "allowJs": true,
    "checkJs": true,
    "noEmit": true,
    "strict": false,
    "noImplicitThis": true,
    "alwaysStrict": true,
  },
  "include": [
    "**/*.js",
    "node_modules/@open-wc/**/*.js"
  ]
}
```

the important part is the `include` of `@open-wc` and that there is NO global ignore of `node_modules` as this would apparently override it.

with that you should get correct types in typescript and awesome intellisense... try hovering over any open-wc helper and use F12 (in vscode) to jup directly to the source code.

daKmoR:
oh it probably needs to be
```
"node_modules/@open-wc/testing-helpers/*.js"
```

as not every package from `open-wc` you probably include has the updated types yet...  if that works we will add support for all our packages

Lodin:
@daKmoR, yay, it works! Still less supported by IDE though, but I don't think it is a blocker.

Lodin:
However, as you said, it doesn't work with `exclude: "node_modules"` which could possibly cause issues.

daKmoR:
hmmm if you do not specify an exclude it's still excluded right?
https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

> The "exclude" property defaults to excluding the node_modules, bower_components, jspm_packages and <outDir> directories when not specified.

so strange why it would not work if you specify it yourself?

Lodin:
@daKmoR, hmm, doesn't it mean that typescript ignores these directories by default if `exclude` option is not specified completely?

daKmoR:
if you still want the exclude you could do it like this
```
/node_modules/!(@open-wc)/
```

daKmoR:
Soooo this now allows tsc to work with all our files 🎉 

Except for `dom5` which is apparently outdated and include definition files (which expect no longer available files) you can see it here: https://circleci.com/gh/open-wc/open-wc/1464?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@LarsDenBakker how hard would it be to get rid of dom5?

Also, we definitely need to explain how to set it up - but once you have it it's really really nice 🤗  awesome intellisense and type safety for js and ts :)

packages/building-rollup/src/dynamic-import-polyfill.js
```diff
@@ -18,7 +29,7 @@ window.importModule = function importModule(url) {
       URL.revokeObjectURL(script.src);
       script.src = '';
     };
-    script.defer = 'defer';
+    script.defer = true;
```
LarsDenBakker: Just to double check, this still produces the correct output?

daKmoR: typing said this only accepts true/false and not a string => so it was a type linting finding :)

I did run the demo and it worked fine... but pls verify 🤗

LarsDenBakker:
Looks good to me, a few questions about this though:

I worked with types in JS before, but my experience wasn't that great. I had to resort to a lot of casting and ignores, which is awkward to write in JS. Going full TS was a much nicer experience. For example `this.shadowRoot` is optional, in TS you can write `this.shadowRoot!.querySelector()` but with annotations you can't.

With this change I think we are making type annotations mandatory for the entire repository? How is your experience with it so far? One of the arguments against going full TS is the increased complexity for new devs coming to a project. Does this not introduce a similar complexity then?

daKmoR:
For what we have in this library it was actually pretty butter smooth... we mostly offer helpers/tools and for those, the experience was really good. I especially liked the nicer intellisense (e.g. not only types but also a description and examples) and that you can hit F12 (in vs code) to go to the actual js source (without an in-between definition file) felt really useful and I hope it helps to reduce the distance to our code for our users (e.g. make it easier to fix bugs/make PRs as you can just do it directly in your node_modules and then you can just copy paste directly into github (if you are lazy)... but yeah we will probably need to have some teachings about this as people usually see node_modules as a magical black box)

For our users, they can choose to use TS or JS - whatever they prefer.

PS: Regarding what we should recommend to our users how they should create web component - let's move that to a dedicated discussion/issue/slack...

LarsDenBakker:
Not arguing against it at this point, but I do wonder about the benefits VS going full typescript if we anyway need to educate people and use tooling to check the types.

Bottom line is that our users get better IDE support out of the box with this, so that's a big win I think :)

Issue #278, opened 2019-03-11
Rollup/Support TypeScript documentation incomplete sentence


budajeff:
"Make sure your set your tsconfig.json to compile to"

There is no period at the end of that sentence and I don't understand what it means which makes me wonder if some of the sentence is missing.

I tried following the Rollup+Typescript documentation but I was not able to get it working. Is there an existing repo somewhere that demonstrates how to use Rollup and Typescript?

Thanks

LarsDenBakker:
Thanks for the notice, it should say: `Make sure you set your tsconfig.json target and module field to exnest. This way typescript doesn't do any compilation so that babel can take care of it`.

There is a working demo here: https://github.com/open-wc/open-wc/tree/master/packages/building-rollup/demo/ts

daKmoR:
@budajeff care to do a MR?

budajeff:
Sure I can do that in a few hours.

budajeff:
#281

daKmoR:
thx for the contribution - closed 💪

PR #281, opened 2019-03-12
fix(building-rollup): fixed tsconfig.json instructions

budajeff:


packages/building-rollup/README.md
```diff
@@ -252,7 +252,7 @@ To import a typescript file, use the `.ts` extension in your `index.html`:
 </html>
 ```
 
-Make sure your set your `tsconfig.json` to compile to
+Make sure you set your `tsconfig.json` `target` and `module` fields to `exnest`. This way typescript won't do any compilation so that babel can take care of it.
```
daKmoR: is that really `exnest`? seems like a typo of `esnext`?

budajeff: You're probably correct. I copied the text in my merge request from a message from @LarsDenBakker because I don't know how to get roll-up to work with typescript myself.

https://github.com/open-wc/open-wc/issues/278#issuecomment-471504554

LarsDenBakker: Oops my mistake. It should be esnext indeed.

budajeff: I can resubmit this later today.

daKmoR:
thaaankkk you ❤️

PR #282, opened 2019-03-12
fix(owc-dev-server): use current working directory for app index

dsnabel:
fixes the app-index path by using the current working directory instead of the __dirname path which locates to the module path.


LarsDenBakker:
Looks good! If you sign the CLA we can merge this. Thanks for the conteibution

dsnabel:
Alright, it’s signed. :)

daKmoR:
awesome thxxx ❤️

PR #285, opened 2019-03-13
fix(create): fix generated prettier scripts to work on Windows

jorenbroekema:
Fixes prettier scripts on Windows.

https://github.com/prettier/prettier/issues/4804 (and you can find many more similar issues on this topic)

Essentially, the globstar pattern isn't recognized on Windows if used with single quotes. Darn Windows.
Anyhow, this fixes it from the looks of it :).


daKmoR:
awesome thxxxxx ❤️

PR #286, opened 2019-03-14
[testing-helpers] ensure ShadyDOM finished its job in fixture

bashmish:
# Expected behavior

In IE11 and older versions of other browsers without native ShadowDOM this test works:

```js
it('my test', async () => {
  class MyElement extends LitElement {
    static get styles() {
      return css`
        :host .wrapper ::slotted(div) {
          position: absolute;
        }
      `;
    }

    render() {
      return html`
        <div class="wrapper">
          <slot></slot>
        </div>
      `;
    }
  }
  const tag = defineCE(MyElement);
  const el = await fixture(`<${tag}><div></div></${tag}>`);
  const div = el.querySelector('div');
  expect(window.getComputedStyle(div).position).to.equal('absolute');
});
```

# Actual behavior

AssertionError: expected 'static' to equal 'absolute'

# Explanation

ShadyCSS polyfill hasn't yet rewrite all class names and apply styles to elements. To ensure it did its job we can run `window.ShadyDOM.flush()`.

```js
  const el = await fixture(`<${tag}><div></div></${tag}>`);
  const div = el.querySelector('div');
  if (window.ShadyDOM) {
    window.ShadyDOM.flush();
  }
  expect(window.getComputedStyle(div).position).to.equal('absolute');
```

I implemented this in the fixture helper since I think the examples above shows very well that this is kinda useless boilerplate and is expected by a fixture user to be done automatically.

### DIFF
--- a/packages/testing-helpers/elementUpdated.js
+++ b/packages/testing-helpers/elementUpdated.js
@@ -9,6 +9,8 @@ const isDefinedPromise = action => typeof action === 'object' && Promise.resolve
  *
  * If none of these is available we await the next frame.
  *
+ * Ensures that ShadyDOM finished its job if available.
+ *
  * @param {HTMLElement} el
  * @returns {Promise<Element>}
  */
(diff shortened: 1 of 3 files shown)


daKmoR:
awesome 👍  thxxxxx 🤗

Issue #279, opened 2019-03-11
Missing building-rollup generator


mario-aleo:
`npm init @open-wc building-rollup` is throwing an error:
`Could not find generator at /<npm init path>/.npm/_npx/1497/lib/node_modules/@open-wc/create/dist/generators/building-rollup/index.js`

I took a look at the master branch and it's really missing the rollup-building folder at open-wc/packages/create/src/generators.

![image](https://user-images.githubusercontent.com/13382278/54124532-6acc7e80-43e1-11e9-9126-b63984de0441.png)

daKmoR:
oh it seem we have been a little eager with adding that info... it does not exist yet 🙈

maybe you could help out adding it? 🤗

mario-aleo:
Yeah, sure, I'll be glad to help :D

daKmoR:
thx for the contribution it's now released 💪

PR #287, opened 2019-03-14
feat(create): building-rollup generator

mario-aleo:
Related to: #279

### DIFF
--- a/packages/create/src/generators/building-rollup/index.js
+++ b/packages/create/src/generators/building-rollup/index.js
@@ -1,0 +1,20 @@
+import { askTagInfo } from '../../helpers.js';
+
+const BuildingRollupMixin = subclass =>
+  class extends subclass {
+    async execute() {
+      await super.execute();
+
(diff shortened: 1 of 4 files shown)


daKmoR:
thx :)
all the symlinks in the docs folder probably don't need to change... (but I can just ignores these changes)

I am currently not in front of a computer.. so I will take a look the day after tomorrow :)

daKmoR:
thaaankkk you for your contribution ❤️

mario-aleo:
I'm really happy that I could help with something, I'm looking forward to keeping helping with anything that I can :D

daKmoR:
awesome :) there is still toons of work to do :p

especially the generators... there is the idea to make it a real separate product; rollup should become the default building generator; also the starter app should adopt it...

the dom compare could use some slimming down as it uses htmlparser2 and parse5 (e.g. pick one)

besides that looking at issues and checking what's needed...

sooo if you wanna pick something would be awesome... and if you need any more info just let us know here or in slack :)

PR #290, opened 2019-03-17
fix(demoing-storybook): allow untyped properties

bennypowers:
properties do not need a type field. see https://lit-element.polymer-project.org/guide/properties#conversion


daKmoR:
LGTM 👍

PR #296, opened 2019-03-19
Update building webpack

LarsDenBakker:
Updates `building-webpack` based on learnings from working with rollup.

- Same config structure as rollup build: `modern` and `modern-and-legacy`
- Browser support based on `.babelrc`
- Ditches `webpack-multi-target-plugin`. It does too much hidden magic which is not needed, and will only cause confusion for users. Webpack already supports multiple outputs, we just need a small plugin to inject both pieces of code.
- Simplified polyfill loading setup. Turns out we don't need to feature detect individual web component APIs, it can just be all or nothing.

I'm working on updating the readme, still a WIP.


LarsDenBakker:
@daKmoR @thepassle @bennypowers this is ready for review / merge.

This needs to be a breaking change release, as the API changed. I think we just add `BREAKING:` to the commit message, right?

daKmoR:
just make it a feat... for release < 1.0.0 every 0.x release is already "breaking" and will not be picked up automatically even when using `^`

bennypowers:
pls rebase

LarsDenBakker:
@daKmoR good idea. We can change the commit message on squash merge

daKmoR:
I have this multiple times?
```
Error: Request to http://localhost:8080/demo/js/a/b/foo.txt failed with 404
    at XMLHttpRequest.s.onreadystatechange (request.js:9)
```

after
```
npm run build:demo
npm run start:demo
```

is this on purpose? the file is that location in the source but of course way different once webpack did it's thing...

LarsDenBakker:
It requires the copy step, I only added it to the modern build config. We could add it to the regular config as well, but it's just for the demo.

daKmoR:
> It requires the copy step, I only added it to the modern build config. We could add it to the regular config as well, but it's just for the demo.

yes pls as otherwise, I see myself coming to look at it again a few days from know and asking the same question again 🙈

packages/building-webpack/package.json
```diff
@@ -26,26 +31,34 @@
...
     "@webpack-cli/serve": "^0.1.3",
     "http-server": "^0.11.1",
     "lit-element": "^2.0.1",
+    "ts-loader": "^5.3.3",
+    "typescript": "^3.3.3333",
     "webpack-bundle-analyzer": "^3.0.3",
     "webpack-cli": "^3.2.1",
-    "webpack-dev-server": "^3.1.14"
+    "webpack-dev-server": "^3.1.14",
+    "webpack-merge": "^4.2.1"
```
daKmoR:
```suggestion
    "webpack-merge": "^4.1.5"
```

we should depend on the same version in the monorepo

packages/building-webpack/package.json
```diff
@@ -26,26 +31,34 @@
   ],
   "dependencies": {
     "@babel/core": "^7.3.3",
+    "@babel/polyfill": "^7.2.5",
```
daKmoR:
```suggestion
    "@babel/polyfill": "^7.0.0",
```

packages/building-webpack/package.json
```diff
@@ -26,26 +31,34 @@
...
     "@babel/preset-env": "^7.0.0",
-    "@babel/runtime-corejs2": "^7.2.0",
-    "@open-wc/polyfills-loader": "^0.2.2",
-    "@open-wc/webpack-import-meta-loader": "^0.1.3",
+    "@babel/plugin-syntax-dynamic-import": "^7.2.0",
+    "@babel/plugin-syntax-import-meta": "^7.2.0",
+    "@webcomponents/webcomponentsjs": "^2.2.7",
```
daKmoR:
```suggestion
    "@webcomponents/webcomponentsjs": "^2.2.0",
```

LarsDenBakker:
@daKmoR done

PR #306, opened 2019-03-21
feat: add version detection

thepassle:
- Adds version detection

Screen Shot 2019-03-21 at 21 19 34

### DIFF
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -31,10 +31,12 @@
     "dist"
   ],
   "dependencies": {
+    "chalk": "^2.4.2",
     "command-line-args": "^5.0.2",
     "deepmerge": "^3.2.0",
     "glob": "^7.1.3",
-    "qoa": "^0.2.0"
+    "qoa": "^0.2.0",
+    "semver": "^5.6.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.2.3",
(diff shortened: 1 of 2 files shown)


daKmoR:
are the added dependencies worth the cost?
did installation time change?

why not use the built in npm check? with package json engine?
https://docs.npmjs.com/files/package.json#engines

thepassle:
semver is 4kb gz and chalk is 7kb gz, if you think its too much we can revert

daKmoR:
I still think adding the info for the engine and npm version to the `package.json` is something we should do https://docs.npmjs.com/files/package.json#engines 

although this only takes effect if you have 
```
npm config set engine-strict true
```
and as it defaults to false... we should keep the custom check... it's a pitty that npm does not respect their own flags

I was still interested in the timings so a made a quick check
Installation time difference with cold cache is +~0.3s or ~+10%
(I think it's mostly time to resolve the 8 additional packages - it seems size alone is not the main factor for npx speed)
```
npx @open-wc/create@0.4.0
npx: installed 29 in 3.42s
npx: installed 29 in 3.333s
npx: installed 29 in 3.332s
=> 1008

npx @open-wc/create@0.3.1
npx: installed 21 in 3.077s
npx: installed 21 in 3.014s
npx: installed 21 in 3.065s
=> 914
```

PR #307, opened 2019-03-21
docs: add input checkbox attribute faq

thepassle:


docs/faq/README.md
```diff
@@ -57,4 +57,13 @@ class MyEl extends LitElement {
     }
   }
 }
-```
\ No newline at end of file
+```
+
+### Checkbox's 'checked' state does not reflect a property's value after the checkbox has been clicked
```
daKmoR: that seems really complicated explained 🙈 
it's about that the attribute does not is equal to the property right?
maybe show some code
```js
// <input id="mycheck" type="checkbox" checked>
console.log(mycheck.checked); // true

// removing the attribute does not change it's checked state
mycheck.removeAttribute('checked');
console.log(mycheck.checked); // true

// changeing it's checked state does not change the attribute
mycheck.checked = false;
console.log(mycheck.hasAttribute('checked')); // true
console.log(mycheck.checked); // false
```

daKmoR:
uh that's way awesome now 💪 

thxxx :)

PR #313, opened 2019-03-23
fix(building-rollup): disable http server caching

thepassle:


packages/building-rollup/README.md
```diff
@@ -68,15 +68,17 @@ To illustrate this, here is an example of scripts which our rollup config will *
...
+    "start:build": "http-server dist -o -c-1",
+    "watch:build": "rimraf dist && rollup --watch -c rollup.config.js & http-server dist -o -c-1",
   }
 }
 ```
 - `build` builds your app and outputs it in your `dist` directory
 - `start:build` runs your built app from `dist` directory
 - `watch:build` builds and runs your app, rebuilding when input files change
 
+Note that caching is disabled by default, if you want to enable caching, remove `-c-1` from the `start:build` and `watch:build` scripts.
```
LarsDenBakker: No need for this note I think. Starts to become too much information. People wont use this in production

daKmoR: true - my bad I suggested it....

thepassle: Removed 👍

daKmoR:
oh we probably should also update the generator for it?

wanna do here or in a separate PR?

thepassle:
@dakmor Good one, fixed it in this PR

PR #316, opened 2019-03-24
fix(building-webpack): allow to provide a publicPath

paales:


LarsDenBakker:
Thanks for your contribution. I think the change is ok. What problem are you solving with this?

daKmoR:
quick question why a separate parameter and not part of the options object?

paales:
> quick question why a separate parameter and not part of the options object?

Not really into the webpack conventions so I was in doubt if I should create `option.publicPath` or access `compiler.options.output.publicPath`. What would you suggest @daKmoR?

paales:
> Thanks for your contribution. I think the change is ok. What problem are you solving with this?

We're actually integrating the open-wc stuff into an existing project which requires different paths when the website is actually served.

packages/building-webpack/src/modern-web-webpack-plugin.js
```diff
@@ -80,14 +66,27 @@ function createEntryVariable(entries, development) {
     : ${asArrayLiteral(entries.legacy)};`;
 }
 
-function createPolyfillsLoaders(options) {
+function createPolyfillsLoaders(options, publicPath) {
```
paales: I should probably rename this to `polyfillsPublicDir` makes more sense.

LarsDenBakker:
Ah, that makes perfect sense to add in. 

@daKmoR the separate parameter is because it's not a user provided option, it's an option retreived inside the plugin from the webpack config.

Good to go, you just need to run `npm run format` from the repository global to format your files.

paales:
@LarsDenBakker I'm unable to do so, if I checout the project run npm install and then npm run format I get: `ESLint couldn't find the plugin "eslint-plugin-html".`.
After adding that one: Error: Cannot find module 'babel-eslint'

Not sure what you are expecting, maybe you can run it your self?

daKmoR:
@paales I will do it :hugs: 

can you sign the CLA then I will make sure to release it?

paales:
@daKmoR Signed :)

daKmoR:

PS: sorry for the delay 🙈  better late then never right 😅 🤣

PR #318, opened 2019-03-24
feat(semantic-dom-diff): added option to ignore html in diff

LarsDenBakker:
- Gets rid of dom5 dependency in `semantic-dom-diff` greatly reducing the total size of this library
- Adds option to ignore certain patterns in the diff. See the readme for options.


packages/semantic-dom-diff/README.md
```diff
@@ -6,21 +6,119 @@
...
   <div>bar</div>
-`;
+`);
+
+// use any string comparison tool, for example chai:
+expect(leftTree).to.equal(rightTree);
+```
+
+## Ignoring tags and attributes
+When working with libraries or custom elements there might be parts of the renderd HTML which is random or otherwise outside of your control. On those cases, you might want to ignore certain attributes or tags entirely. This is possible by passing an options object.
```
thepassle:
```suggestion
When working with libraries or custom elements there might be parts of the rendered HTML which is random or otherwise outside of your control. In those cases, you might want to ignore certain attributes or tags entirely. This is possible by passing an options object.
```

packages/semantic-dom-diff/package.json
```diff
@@ -26,17 +26,15 @@
...
   "devDependencies": {
     "@bundled-es-modules/chai": "^4.2.0",
     "@open-wc/testing-karma": "^0.4.15",
     "@open-wc/testing-karma-bs": "^0.2.17",
     "@open-wc/testing-wallaby": "^0.1.12",
     "mocha": "^5.0.0",
     "webpack-merge": "^4.1.5"
+  },
+  "dependencies": {
+    "htmlparser2": "^3.10.1"
```
daKmoR: that will not work via polymer serve/owc-dev-server right?
so we should open an issue and make an estimate module version right?

LarsDenBakker: Indeed, it's a WIP item left to do. I was thinking of pre-bundling within `semantic-dom-diff` as a prepublish step. We can the minify it too to reduce the footprint when running tests. What do you think?

LarsDenBakker:
I added a pre-bundling step for `semantic-dom-diff`. This is needed in order to make sure that it still works in the browser as pure es modules. I also added a minifying step so that it adds minimal footprint to user's tests.

Unfortunately after minify it's still 120kb (non gzip) which is a really unfortunate amount to add to user's tests. https://bundlephobia.com/result?p=htmlparser2@3.10.1 reports 62kb but it doesn't take into account that it has to add polyfills for a lot of nodejs builtins. 

In the next iteration we can try to see if we can find a html parser that's light and more browser oriented. 

I added a `prepublish` hook for npm, that should work right? Does anyone have any experience with that?

Other than that I think we're good to go here. Once this is published, I will expose the options in `chai-dom-equals`.

packages/semantic-dom-diff/package.json
```diff
@@ -24,19 +24,24 @@
     "test:watch": "karma start --auto-watch=true --single-run=false",
     "test:es5": "karma start karma.es5.config.js",
     "test:es5:watch": "karma start karma.es5.config.js --auto-watch=true --single-run=false",
+    "postinstall": "rollup -c rollup.config.js",
+    "prepublish": "rollup -c rollup.config.js",
```
daKmoR:
```suggestion
```

prepublish is deprecated use prepublishOnly

packages/semantic-dom-diff/package.json
```diff
@@ -24,19 +24,24 @@
...
     "prepublishOnly": "../../scripts/insert-header.js"
```
daKmoR:
```suggestion
     "prepublishOnly": "../../scripts/insert-header.js && rollup -c rollup.config.js"
```

LarsDenBakker:
CI failed because of flaky IE11 test.

daKmoR:
@LarsDenBakker is it ready to be released? you wanna release it right?

LarsDenBakker:
Yes we can release it

PR #319, opened 2019-03-25
docs: fix file extension

glitchperfect:
Fix the .js extension to .json so it matches the actual source file


daKmoR:
thhaaankkks ❤️

PR #321, opened 2019-03-26
fix: default webpack generator config and docs

thepassle:
I think this fixes the webpack generator/manual instructions, it doesn't seem to pick up the default options correctly though, need to dive into it more


LarsDenBakker:
Looks good to me. Should we remove WIP?

daKmoR:
LGTM as well 👍

thepassle:
can be merged :)

Issue #284, opened 2019-03-13
[create] npm init @open-wc: 'command failed: command -v yarn'


jorenbroekema:
## Description
This is on node v10.15.3 and npm v6.9.0

I tried to do `npm init @open-wc` and install barebone or starter app. Both fail due to an error when firing `command -v yarn`

### Windows 10
```
Installing dependencies...
This might take some time...
{ Error: Command failed: command -v yarn
'command' is not recognized as an internal or external command,
operable program or batch file.

    at ChildProcess.exithandler (child_process.js:294:12)
    at ChildProcess.emit (events.js:189:13)
    at maybeClose (internal/child_process.js:970:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
  killed: false,
  code: 1,
  signal: null,
  cmd: 'command -v yarn',
  stdout: '',
  stderr:
   '\'command\' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n' }
```

### Linux Ubuntu 16.04
```
Installing dependencies...
This might take some time...
{ Error: Command failed: command -v yarn

    at ChildProcess.exithandler (child_process.js:294:12)
    at ChildProcess.emit (events.js:189:13)
    at maybeClose (internal/child_process.js:970:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
  killed: false,
  code: 127,
  signal: null,
  cmd: 'command -v yarn',
  stdout: '',
  stderr: '' }
```

Here is where it fails https://github.com/open-wc/open-wc/blob/master/packages/create/src/core.js#L107

If I run `command -v yarn` in my bash terminal it works just fine both for windows 10 and linux and prints the installation path of my yarn. So it seems like it not working has something to do with node trying fire it.

I did some googling but couldn't really find the problem, and since the command is literally command it's quite hard to find good search results haha.

Let me know if you guys have some suggestions on how to fix, I'm happy to try to test out suggestions and make MR if it works.

daKmoR:
this is "just" the last step if doing an automatic npm install so everything should be fine... you would only need to do the npm install yourself...

still strange that it fails...

jorenbroekema:
Ah I guess if it's the last step, that's a good workaround for now :P

jorenbroekema:
Btw, perhaps this is unrelated, but I tried doing `npm init @open-wc linting` and, as expected, it gives the same error, but it also does not add the config files 
```
husky.config.js
.eslintignore
[...]

PR #322, opened 2019-03-26
[create] Fix bug with install deps

jorenbroekema:
fixes https://github.com/open-wc/open-wc/issues/284

It also allows the user to select the dependency manager.

Note: There is one problem where the prompt is fired twice, once before installing dependencies, and then at the end it prompts the user again. I am not sure why this is so perhaps one of you guys can look into that, I can't really figure it out.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,11 +19,10 @@ cd open-wc
 Now that you have cloned the repository, run the following commands to set up the development environment.
 
 ```
-npm install
-npm run bootstrap
+yarn install
 ```
 
-Both commands will install packages like eslint which are necessary to make your changes.
+This will download and install all packages needed.
 
 # Making Your Changes
 
--- a/packages/create/src/core.js
+++ b/packages/create/src/core.js
@@ -2,15 +2,12 @@
 
 import fs from 'fs';
 import path from 'path';
-import util from 'util';
-import { exec, spawn } from 'child_process';
+import { spawn } from 'child_process';
 
 import glob from 'glob';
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -20,7 +20,6 @@ Now that you have cloned the repository, run the following commands to set up th
 
 ```
 npm install
```
daKmoR:
```suggestion
yarn install
```

CONTRIBUTING.md
```diff
@@ -20,7 +20,6 @@ Now that you have cloned the repository, run the following commands to set up th
...
-npm run bootstrap
 ```
 
 Both commands will install packages like eslint which are necessary to make your changes.
```
daKmoR:
```suggestion
This will download and install all packages needed.
```

packages/create/src/core.js
```diff
@@ -99,18 +96,13 @@ function _install(command = 'npm', options) {
   });
 }
 
-export async function installNpm(where) {
+export async function installNpm(where, packageManagerChoice) {
```
daKmoR: can we also move the "decision" to the generator so this is
```js
export async function installNpm(where, command) {
```

I think that's easier to understand

jorenbroekema: Not sure if I understood what you meant fully with move the "decision" but I changed it

daKmoR: that was exactly what I had in mind 💪

daKmoR:
wooohooo thank you for this contribution ❤️

PR #326, opened 2019-03-27
fix(building-rollup): Add missing dependency

Pupix:


Pupix:
Welp I borked the commit message.

thepassle:
Could you check if this needs to be done in the generator as well? Id like to keep everything in sync :)

Pupix:
The generator uses the `building-rollup` package directly https://github.com/open-wc/open-wc/blob/aa3018db055e138830b8ff6fc31b26f3cff2fbbd/packages/create/src/generators/building-rollup/templates/_package.json#L8 so it should inherit the change without modifying anything.

thepassle:
Oki, just making sure. Thanks! LGTM.

I can't merge, @daKmoR  ?

LarsDenBakker:
What is clone used for? I dont see it used.

LarsDenBakker:
Oh its in the dom5 fork .

PR #336, opened 2019-03-29
feat(testing-karma): simplify/improve karma config

LarsDenBakker:
Updates the karma config:

- Update to reflect latest webpack config updates. Browser support based on `.browserslistrc`, babel config based on `.babelrc`

- Only run coverage when `--coverage` flag is passed. Update recommendation to do this only when not watching. The reason for this is that instrumenting for coverage is slow and adding the coverage reporter causes all logs to be logged twice in the terminal. This is a known issue in karma.

- Mocha HTML reporter is configured, so that when you visit http://localhost:9876/debug.html you get a nice BDD UI (just like WCT did).

- Errors during compiling are passed a JSON objects with a message property. This causes unreadable error logs, and lots of spam when importing the wrong file. Formatting the error helps a lot here.

- Static files are correctly served from the current working directory.

- Opened a PR at https://github.com/webpack-contrib/karma-webpack/pull/404 to prevent karma-webpack and karma logging the same error twice.

Note: This is a breaking change.


LarsDenBakker:
I've updated the init script as well, but I'm not sure how to test it until the versions are released. Any ideas?

LarsDenBakker:
This is also a breaking change that needs to be applied across the monorepo, not sure how that should be handled. Should/can we do the update within the same MR as making these changes?

thepassle:
You can test the generator by cding to the `create` package and just running `start: npm run build && node ./dist/create.js`. Choose the karma generator, and it'll output in the same directory

daKmoR:
yes we should do the needed change for all the files across all the packages...

as this is a breaking change would be nice to write a blog post about it right? do we want to use/enable the vuepress blog for it? or create a dev.to open-wc user?

btw. really nice improvement :+1: still needs some cleanup but I really like the direction :hugs:

LarsDenBakker:
I think a blog post in the sense of a 'news post' would make sense. Not a standalone blog post.

Also a `MIGRATION.md` would be good.

LarsDenBakker:
I added `require.resolve` for all plugins we use. This makes it so that modules are always resolved relative to the `testing-karma` package, rather than the package where it's installed. In real life projects this was giving me a lot of issues.

I adjusted all the configs, and standardized all the naming conventions, and it passes the CI. I squashed the commits, so I think we're good to go here.

daKmoR:
I am looking at it now 💪

daKmoR:
this is only testing in chrome even when running the bs version 🙈

PR #338, opened 2019-03-30
fix(semantic-dom-diff): remove dependency on libraries

LarsDenBakker:
- Refactor `semantic-dom-diff` to remove dependency on external libraries. This reduces the size from 160kb (minified) to 7kb (non-minified) :)
- Pass along diff configuration from `chai-dom-equals` to `schemantic-dom-diff`

We need to improve the documentation on this subject, as it's in several places right now. I propose we do that in a separate PR.

There are no breaking changes.


LarsDenBakker:
I added a lot of tests, definitely would be interesting to come up with more scenarios...

packages/semantic-dom-diff/get-diffable-html.js
```diff
@@ -0,0 +1,243 @@
...
+
+    if (node instanceof Element) {
+      printCloseElement(node);
+    }
+  }
+
+  const container = document.createElement('diff-container');
+  container.innerHTML = html;
+
+  const walker = document.createTreeWalker(
```
daKmoR: as I did some adjustments for lit-html on this as well I know that IE doesn' t like that

you will need something like this
```
    // Edge needs all 4 parameters present; IE11 needs 3rd parameter to be null
    const walker = document.createTreeWalker(
        element.content,
        133 /* NodeFilter.SHOW_{ELEMENT|COMMENT|TEXT} */,
        null,
        false);
```

LarsDenBakker: thanks!

Issue #342, opened 2019-04-01
[building-webpack] Index.html script undefined


jorenbroekema:
Currently, using building-webpack, it is not properly bundling my entry point js and linking it in the entry point html. 

I used `npm init @open-wc building-webpack` in the root folder of my app.

Setup:
Windows 7
Node: 10.15.1
NPM: 6.9.0
Yarn: 1.13.0

Folder structure:
```
my-app
| -- dist
    | -- legacy
    | -- polyfills
    | -- index.html
    | -- main.[chunkhash].js
    | -- main.[chunkhash].js.map
| -- packages
    | -- my-app/src
        | -- index.html
        | -- my-app.js 
| -- .browserslistrc
| -- package.json
| -- webpack.config.js
| -- yarn.lock
```

`webpack.config.js`:
```js
const { resolve } = require('path');
const createDefaultConfig = require('@open-wc/building-webpack/modern-and-legacy-config');

// If you don't need IE11 support, use the modern-config instead
// import createDefaultConfig from '@open-wc/building-webpack/modern-config';

module.exports = createDefaultConfig({
  entry: resolve(__dirname, './packages/my-app/src/my-app.js'),
  indexHTML: resolve(__dirname, './packages/my-app/src/index.html'),
});
```
src/my-app.js
```js
import { LionLitElement, html } from 'ing-web/core';

class MyApp extends LionLitElement {
  render() {
    return html`
      <h1>Hello world!</h1>
    `;
  }
}

customElements.define('my-app', MyApp);
```

src/index.html
```html
<html>
  <head></head>
  <body>
    <my-app></my-app>
  </body>
</html>
```

dist/index.html
```html
<html>
  <head><link rel="preload" href="main.0f31193ba01e5fb5d98e.js" as="script"></head>
  <body>
    <my-app></my-app>
  <script src="polyfills\babel.js" nomodule></script><script>undefined</script></body>
</html>
```
^ Especially this one gives unexpected results. 

When running `"http-server dist/ -o"` I get the following console error in latest Chrome:
![image](https://user-images.githubusercontent.com/36734656/55334034-baf6a800-5498-11e9-9894-56d594b2c7fd.png)

Another unrelated buggy thing is that `http-server` is not part of the devDependencies of building-webpack package.json, while it is used in one of the scripts. That should probably also be fixed.

jorenbroekema:
Update: This seems to be a Windows only bug. I tested this just now on Windows 10, same issue where it doesn't build properly.

[...]

PR #344, opened 2019-04-01
fix(building-webpack): support windows file pathes as well

daKmoR:
closes https://github.com/open-wc/open-wc/issues/342

nodes `path.join` uses system separators so on windows it will create urls like `foo\bar.js`... that maybe still works for browser if they are nice and accept/change them... but Terser just says `\b` is not a valid escape thing and breaks...

=> now we force `/` even on windows

### DIFF
--- a/packages/building-webpack/src/modern-web-webpack-plugin.js
+++ b/packages/building-webpack/src/modern-web-webpack-plugin.js
@@ -54,6 +54,11 @@ const legacyError = error(
   'You are running webpack-dev-server with a legacy build, run the dev server with the --legacy flag for this browser or run a production build.',
 );
 
+function pathJoin(...args) {
+  // enfore / also on windows => as it is used for web pathes
+  return path.join.apply(null, args).replace(/\\(?! )/g, '/');
+}
+
 function createEntryVariable(entries, development) {
   if (development) {
     return `var entries = 'noModule' in HTMLScriptElement.prototype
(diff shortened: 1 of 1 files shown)


packages/building-webpack/src/modern-web-webpack-plugin.js
```diff
@@ -54,6 +54,11 @@ const legacyError = error(
   'You are running webpack-dev-server with a legacy build, run the dev server with the --legacy flag for this browser or run a production build.',
 );
 
+function pathJoin(...args) {
```
LarsDenBakker: I suggest naming this browserPathJoin or something like that. Also, why do we need to call apply here?

packages/building-webpack/src/modern-web-webpack-plugin.js
```diff
@@ -54,6 +54,11 @@ const legacyError = error(
   'You are running webpack-dev-server with a legacy build, run the dev server with the --legacy flag for this browser or run a production build.',
```
LarsDenBakker: I think the solution is correct but I am not sure if we need to apply it in all places. I think we only need this in places where we ship to the browser, right? Any code that runs in mode still needs os specific separators. Are you sure we got this covered correctly?

LarsDenBakker:
Great find, thanks!

packages/building-webpack/src/modern-web-webpack-plugin.js
jorenbroekema: I tested this code on Windows 7 just now and it works ;).

daKmoR: all the pathes changed are only used to finally end up in the frontend... so I don't see a path where we should not have applied it?

packages/building-webpack/src/modern-web-webpack-plugin.js
daKmoR: `path.join takes` a list of n argument. can you `forward` n arguments without using apply?

LarsDenBakker: Yeah I think you can just do `path.join(...args)`

packages/building-webpack/src/modern-web-webpack-plugin.js
LarsDenBakker: Ok, thanks for testing. I was reviewing on my phone so it was hard to tell.

PR #347, opened 2019-04-02
feat(building-webpack): enable single entry point html file

daKmoR:
- Allows to use a single `input` HTML File for a webpack config
- This means that the same `index.html` file can be used with rollup/webpack/polymer server/owc-dev-server
- is backward compatible as there is a new config `input` if that one is used it overrides the entry and indexHTML
- uses pretty much the same code as found in rollup (parse5, dom5)
- parse5/dom5 could probably be ditched as we only need to take care of script tags (or we could try to align with the rollup config and offer a index.html to webpack/rollup config tool)
- adds all script src urls as a webpack entry point
- removes all script tags before passing it on to `HtmlWebpackPlugin`

=> this is a POC to show the idea  
=> I think this is really promising and we should promote it as the default way of using it (even deprecating `entry & indexHTML`)

### DIFF
--- a/docs/building/README.md
+++ b/docs/building/README.md
@@ -13,6 +13,19 @@ Building is a necessary optimization when shipping apps to production. Generally
 
 Our recommended setups address the first four points but refrain from introducing non-standard features. We think it's best to avoid those, as they might lock your source code to a specific build setup which can be hard to get away from at a later time. Javascript is evolving fast, your code can quickly become out of sync with new developments.
 
+## Using `index.html` as an entry point
+Our configs for `rollup` & `webpack` are unique in that sense that they take a single html file as an entry point.
+Doing so allows you to work with the same entry point no matter if you use
+- webpack
+- rollup
+- owc-dev-server
+- polymer serve
+- etc
+
+This means you can easily compare how different setups affect your app size and loading time.
+This allows you to easily switch between build configurations; if you want to switch to webpack for a while, your entrypoint will be exactly the same.
+It also means you can use `owc-dev-server` for your regular development, and then use either `rollup` or `webpack` to prepare your application for your production environment.
+
 ## Rollup
 We recommend [Rollup](https://rollupjs.org/) for building front-end projects. Rollup is convenient to use and gets out of your way. It is easy to understand what's going on. Quite a relief in a world of complex javascript tooling.
 
--- a/packages/building-webpack/modern-config.js
+++ b/packages/building-webpack/modern-config.js
@@ -5,9 +5,14 @@ const TerserPlugin = require('terser-webpack-plugin');
 
 const development = !process.argv.find(arg => arg.includes('production'));
 
+const prefix = '[@open-wc/building-webpack/modern-config]:';
+const { queryAll, predicates, getAttribute } = require('./dom5-fork/index.js');
+const { readHTML } = require('./src/utils.js');
+
 const defaultOptions = {
   indexHTML: './index.html',
   entry: './index.js',
+  htmlEntryPoint: false,
 };
 
 module.exports = userOptions => {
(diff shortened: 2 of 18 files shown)


packages/building-webpack/modern-config.js
```diff
@@ -67,6 +92,10 @@ module.exports = userOptions => {
             },
           },
         },
+        {
+          test: /\.html$/,
```
LarsDenBakker: Make sure to only match the actual `index.html` rather than any `.html` file.

packages/building-webpack/modern-config.js
```diff
@@ -16,6 +20,27 @@ module.exports = userOptions => {
     ...userOptions,
   };
 
+  if (typeof options.input === 'string' && options.input.endsWith('index.html')) {
```
LarsDenBakker: What if we base this off the `indexHTML` property? So if `indexHTML` is set and `entry` is not set, we extract it from the `indexHTML` file?

LarsDenBakker:
This is really nice! A very minimal approach, I hadn't thought of it like that. I think we should go with this approach.

Indeed we could create an 'extract modules' tool for which we probably don't need a big parser library.

packages/building-webpack/modern-config.js
daKmoR: thought about it as well but that would be a breaking change... as it would result in different defaults...
```js
options = { indexHTML: 'foo.html' }

// result now
{ index:HTML: 'foo.html', entry: 'index.js' }
// result if we do change
{ index:HTML: 'foo.html', entry: '<what we find in indexHTML>' }
```

LarsDenBakker: This would work, right:

```
if (options.indexHTML && !options.entry) {
  options.entry = extractModulesFromHTML(options.indexHTML);
}
```

This way our config can have a great default, but also offer a more 'traditional' or 'custom' approach where needed.

packages/building-webpack/modern-config.js
daKmoR: only add the loader if needed and only match the input file 💪

packages/building-webpack/modern-config.js
daKmoR: as discussed entry is always defined as we merge the default options before and we would like to align with buildling-rollup where it's called input as well

docs/building/README.md
```diff
@@ -13,6 +13,19 @@ Building is a necessary optimization when shipping apps to production. Generally
...
+Our configs for `rollup` & `webpack` are unique in that sense that they take a single html file as an entry point.
+Doing so allows you to work with the same entry point no matter if you use
+- webpack
+- rollup
+- owc-dev-server
+- polymer serve
+- ...
+
+This means you can easily compare how different setups effect your app size and loading time.
+So this truely means you can use `owc-dev-server` for your regular development then use `rollup` to prepare it to be put on a server.
```
thepassle:
```suggestion
So this truly means you can use `owc-dev-server` for your regular development, and only leverage build tools like `rollup` to prepare it to be hosted on a server.
```

docs/building/README.md
```diff
@@ -13,6 +13,19 @@ Building is a necessary optimization when shipping apps to production. Generally
...
+And if at some point you need to make an emergency fix that requires a dependecy that is not available as an es module just switch to `webpack ` for a few days. Make sure to request an es module version or make a temporary fork and switch back to `rollup` for optimal performance.
```
thepassle:
```suggestion
And if at some point you need to make an emergency fix that requires a dependency that is not available as an es module, you can easily switch to `webpack` for a few days. (Make sure to request an es module version or make a temporary fork and switch back to `rollup` for optimal performance!)
```

thepassle: Also, rollup can handle commonjs right? Isnt this a bit of a weird example? As a developer im not very attracted to the idea of having to switch between build tools (even with open-wc recommendations/configs)

docs/building/README.md
```diff
@@ -13,6 +13,19 @@ Building is a necessary optimization when shipping apps to production. Generally
...
 Our recommended setups address the first four points but refrain from introducing non-standard features. We think it's best to avoid those, as they might lock your source code to a specific build setup which can be hard to get away from at a later time. Javascript is evolving fast, your code can quickly become out of sync with new developments.
 
+## Using `index.html` as an entry point
+Our configs for `rollup` & `webpack` are unique in that sense that they take a single html file as an entry point.
+Doing so allows you to work with the same entry point no matter if you use
+- webpack
+- rollup
+- owc-dev-server
+- polymer serve
+- ...
```
thepassle:
```suggestion
- etc
```

docs/building/README.md
```diff
@@ -13,6 +13,19 @@ Building is a necessary optimization when shipping apps to production. Generally
...
+Doing so allows you to work with the same entry point no matter if you use
+- webpack
+- rollup
+- owc-dev-server
+- polymer serve
+- etc
+
+This means you can easily compare how different setups affect your app size and loading time.
+This allows you to easily switch between build configurations; if you want to switch to webpack for a while, your entrypoint will be exactly the same.
+It also means you can use `owc-dev-server` for your regular development then use `rollup` or `webpack` to prepare your application for your production environment.
```
thepassle:
```suggestion
It also means you can use `owc-dev-server` for your regular development, and then use `rollup` or `webpack` to prepare your application for your production environment.
```

PR #355, opened 2019-04-06
feat(building-webpack): make custom webpack plugin public

LarsDenBakker:
- Make webpack plugin public so that it can be used without the default config
- Correctly reference options inside webpack plugin


daKmoR:
LGTM 👍 

pls rebase or update + squash 🤗

PR #356, opened 2019-04-06
fix(chai-dom-equals): ignore style tag in shadow dom by default

daKmoR:


packages/chai-dom-equals/chai-dom-equals.js
```diff
@@ -100,9 +100,9 @@ export const chaiDomEquals = (chai, utils) => {
   const domEquals = _super => function handleDom(value, ...args) {
     // @ts-ignore
     if (utils.flag(this, 'shadowDom')) {
-      const expectedHTML = getDiffableHTML(value, args[0]);
+      const expectedHTML = getDiffableHTML(value, { ignoreTags: ['style'], ...args[0] });
```
LarsDenBakker: this does not merge user-provided `ignoreTags`

daKmoR: yeah it overrides... which would be good I would say...
```js
expect(el).shadowDom.to.equal('...'); // ignores style tag
expect(el).shadowDom.to.equal('...', { ignoreTags: ['foo'] }); // does not ignores style tag
// e.g. you will need to ignore it yourself if you want to keep ignoring
expect(el).shadowDom.to.equal('...', { ignoreTags: ['foo', 'style'] });
```

LarsDenBakker: I think that's a confusing API, and it's an extra thing to explain/document. Maybe the better approach is to just ignore style, svg and script tags entirely and not just the content. Directly in `semantic-dom-diff` (basically the same thing we had before). That way at least it's straightforward

daKmoR: true let's ignore it globally for now... if there is a need to override these default we can look at it when the need arises...

PR #357, opened 2019-04-07
Feat/snapshot testing

LarsDenBakker:
- adds support for testing with snapshots in `testing-karma` 
- merges `chai-dom-equals` into `semantic-dom-diff`
- renamed chai plugin to `chai-dom-diff`
- adds support for testing with snapshots in `chai-dom-diff`

Snapshot workflow:

In your test:
```
describe('my-element', () => {
  it('renders state A correctly', () => {
    const el = fixture(elementInStateA);
    expect(el).shadowDom.to.equalSnapshot();
  });

  it('renders state B correctly', () => {
    const el = fixture(elementInStateB);
    expect(el).shadowDom.to.equalSnapshot();
  });
});
```

Snapshot is saved as: `__snapshots__/my-element.md` containing state A and B.

On each test run, it checks if the fixture matches the snapshot. Otherwise it fails.

If your dom updated on purpose, you need to run: `yarn update-snapshots`. 

For more details, see the updated readmes of `testing-karma` and `semantic-dom-diff`.

This change has a large DX element, so please test it and let me know what you think :)


packages/create/src/generators/testing-karma/templates/_package.json
```diff
@@ -3,7 +3,9 @@
     "test": "karma start --coverage",
     "test:watch": "karma start --auto-watch=true --single-run=false",
     "test:legacy": "karma start --legacy --coverage",
-    "test:legacy:watch": "karma start --legacy --auto-watch=true --single-run=false"
+    "test:legacy:watch": "karma start --legacy --auto-watch=true --single-run=false",
+    "update-snapshots": "karma start --update-snapshots",
```
daKmoR:
```suggestion
    "test:update-snapshots": "karma start --update-snapshots",
```

I think it's test specific so it should have the test prefix right?

LarsDenBakker: Good point, I've adjusted it everywhere

LarsDenBakker:
@daKmoR I made the changes to ignore style, script and svg tags by default here.

PR #361, opened 2019-04-09
Feat/prevent snapshot warnings

LarsDenBakker:
- prevent filename warnings from snapshot filenames
- don't warn about unused snapshots if you are using `grep` to run individual files


LarsDenBakker:
@daKmoR I made sure to include all new files in the `package.json` this time :)

packages/testing-karma/default-config.js
```diff
@@ -48,6 +48,7 @@ module.exports = config => ({
     require.resolve('karma-snapshot'),
     require.resolve('karma-mocha-snapshot'),
     require.resolve('karma-chrome-launcher'),
+    require.resolve(require.resolve('./src/snapshot-filename-preprocessor.js')),
```
daKmoR: why is a nested `require.resolve(require.resolve(` needed?

LarsDenBakker: It's not :D I didn't have my morning coffe when I wrote that, I think.

LarsDenBakker:
merged into https://github.com/open-wc/open-wc/pull/387

PR #365, opened 2019-04-10
docs(owc-dev-server): improve various bits

djlauk:
As I already had the editor open to add a note on the package name (see [previous discussion on slack](https://polymer.slack.com/archives/CE6D9DN05/p1554801186272300)), I couldn't help but felt like adding and tweaking a bit here and there.


packages/owc-dev-server/README.md
```diff
@@ -33,26 +33,41 @@ Please note that the owc-dev-server *only resolves bare module specifiers*, and
...
 }
+```
 
-// bash
+This allows you to execute owc-dev-server like this:
+```bash
+# run package.json script 'start'
 npm run start
+# or
+yarn run start
```
daKmoR:
```suggestion
yarn start
```

yarn does not need the run

packages/owc-dev-server/README.md
```diff
@@ -33,26 +33,41 @@ Please note that the owc-dev-server *only resolves bare module specifiers*, and
...
+```
 
-// or via npx
+### Execute with npx
+You can also start owc-dev-server through `npx`. This allows you to easily try out different [command line options](#command-line-options). Starting owc-dev-server through `npx` will even work if you did not previously install it ([see above](#usage)).
```
LarsDenBakker: can you use npx without installing a module? will that install it on the fly?

daKmoR: jup that's the idea behind npx 🤗 

however we had some path problems with `owc-dev-server` when it was used via npx... not 100% if we have them all fixed 🙈

packages/owc-dev-server/README.md
djlauk: I know that the `run` is optional. It's my personal preference, though.
This way, if `yarn` ever comes up with a new sub-command that happens to be the same as one of my scripts, I won't have to change my habits -- or worse: documentation.

packages/owc-dev-server/README.md
djlauk: Nope, it's not fixed yet for owc-dev-server. Just verified with a fresh app (using `polymer init --name polymer-3-starter-kit`). Sorry, I assumed wrongly without trying it first. 🙈
I will adjust the parts *usage* and *execute with npx* accordingly.

djlauk: FYI: I created #366 so I can reference it in the docs.

thepassle:
Can this be merged?

packages/owc-dev-server/README.md
```diff
@@ -33,26 +33,45 @@ Please note that the owc-dev-server *only resolves bare module specifiers*, and
...
+### Execute with npx
+You can also start owc-dev-server through `npx`. This allows you to easily try out different [command line options](#command-line-options).
+```bash
+# run ad-hoc via npx
 npx owc-dev-server
 ```
 
-#### Static files
+::: warning
+If you didn't install owc-dev-server locally (as a dependency, [see above](#usage)), running `npx owc-dev-server` will start owc-dev-server, but it **will not work** properly and report errors. See [this issue](https://github.com/open-wc/open-wc/issues/366) for details.
```
daKmoR: can we rephrase that? maybe something like this
```suggestion
`owv-dev-server` only works properly when installed locally. 
```

coming to think of it maybe we should remove all the `npx` stuff and e.g. better explain how to add a script and then only use `npm run start` or so - what do you think?

djlauk: Good point. Should definitely get some thought...

After all, what I loved so much about the open-wc docs, was getting clear recommendations how to set things up. And I learned a lot from the "manual setup" parts of the documentation, too. Ever since I have a better grasp on how these tools interact. (There's still much to learn for me, though.)

So... how about I give it another shot and remove all mentions of npx?

daKmoR: sounds good too me :hugs:  @lars do you agree?

LarsDenBakker:
This PR dragged on for a long time, really sorry about that :( 

We've completely overhauled the dev server now: https://github.com/open-wc/open-wc/pull/520 if we want to make improvements let's do it on the readme of that project.

Again sorry and thanks for the work you did on this.

PR #375, opened 2019-04-13
WIP: chore: add documentation link to scaffold

jlengrand:
* add mocha to dev deps to avoid requiring it in the global scope


package.json
```diff
@@ -45,6 +45,7 @@
     "husky": "1.1.2",
     "lerna": "3.4.3",
     "lint-staged": "^8.1.0",
+    "mocha": "^6.1.3",
```
daKmoR: I just ran this in a new folder
```bash
git clone git@github.com:open-wc/open-wc.git
cd open-wc
yarn install
yarn test
```

and there was no need for me to add mocha as a dependency?

what version of yarn do you have?

if we need (can't reproduce this need) to add another dependency then it would probably be this
```
"@open-wc/testing": "file:./packages/testing",
```

jlengrand: None needed. Let me see how to fix this CLA issue now. I did the first two commits with my ING account and github doesn't like it.

Might be just as fast fixing it yourself :).

daKmoR: I think it's better to learn how to fix it 👍 
so you have 3 commits now
```bash
# remove them all but keep the file changes
git reset --soft HEAD~3

# make a new commit with the correct user
git commit -m 'fix(create): add documentation link to scaffold'

# if you want you can also specify a specific author
git commit -m 'fix(create): add documentation link to scaffold'  --author="Author Name <email@address.com>"

# or if you want to change the author of an existing commit
git commit --amend --author="Author Name <email@address.com>"
```

jlengrand:
Hum, prettier fails on me now.

package.json
jlengrand: Well, I commit quite often with my ING account, so I actually preferred adding my work email to my account. That fixed it as well. Thanks for taking time to provide info though <3

jlengrand:
Do note that I have not run the change locally. Not sure it looks as desired.

packages/create/src/generators/app/index.js
```diff
@@ -44,7 +44,8 @@ const AppMixin = subclass =>
       const { scaffold } = await qoa.prompt([
         {
           type: 'interactive',
-          query: 'What would you like to scaffold?\nNote: Content will be written in a new folder.',
+          query:
+            'What would you like to scaffold?\n You can read more abbout your options here : https://open-wc.org/developing/create.html#usage\nNote: Content will be written in a new folder. ',
```
thepassle:
```suggestion
            'What would you like to scaffold?\n You can read more about your options here : https://open-wc.org/developing/create.html#usage\nNote: Content will be written in a new folder. ',
```

daKmoR:
I am already working on https://github.com/open-wc/open-wc/issues/376 so this whole flow might be quite different 🙈 

this sort of was the kick in the butt I needed to finally pick it up - so already thx for that 🤗 
however, this might mean that we will not merge this 🙈

packages/create/src/generators/app/index.js
jlengrand: Good one thanks <3

jlengrand:
No biggie. Closing. Have fun with #376

PR #382, opened 2019-04-13
docs: add LitElement-friendly .babelrc definition to webpack build page

ZempTime:
Spent a little trying to figure out the right `.babelrc` configuration for typescript by trying different manual combinations as specified in the linked babel docs. These are order and option-sensitive. Then I copied https://github.com/Polymer/lit-element/blob/master/.babelrc and things began to work.

If you're trying to use `LitElement` in typescript files with `@open-wc`-generator powered configs, this is what you'll need.

### DIFF
--- a/packages/building-webpack/README.md
+++ b/packages/building-webpack/README.md
@@ -220,6 +220,23 @@ We recommend using the babel typescript plugin. Add this to your `.babelrc`:
 
 This the fastest method, as it strips away types during babel transformormation of your code. It will not perform any type checking though. We recommend setting up the type checking as part of your linting setup, so that you don't need to run the typechecker during development for faster builds.
 
+
+<details>
+  <summary>Supporting `LitElement`'s `@customElement` and `@property`</summary>
+
+  > Due to the way Babel handles [decorators](https://babeljs.io/docs/en/babel-plugin-proposal-decorators) and class properties, you'll need to specify the plugins in a specific order with specific options. Here's what you'll need:
+  ```javascript
+  {
+    "presets": ["@babel/preset-typescript"],
+    "plugins": [
+      ["@babel/plugin-proposal-decorators", { "decoratorsBeforeExport": true }],
+      "@babel/plugin-proposal-class-properties"
+    ]
+  }
+
+  ```
+</details>
+
 #### 2. Plugin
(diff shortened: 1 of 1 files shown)


ZempTime:
Sorry, I'm gonna re-open this with a better named commit.

daKmoR:
> Sorry, I'm gonna re-open this with a better named commit.

if you want to change a commit you can just rename you commit and do a force push
```bash
# change message of last commit
git commit --amend -m 'your new commit message'
# update PR
git push --force
```

@ZempTime

ZempTime:
Thanks. Forgot that was a possibility.

Screen Shot 2019-04-13 at 12 27 35 PM

Do I need to do anything else to get that picked up in this pr? Seems like [it should update](https://stackoverflow.com/questions/32346624/what-happens-if-you-force-push-to-a-branch-with-an-existing-pull-request), but let me know if there's anything else I should do if this is a pr you'd want to accept.

ZempTime:
Also, what's proper form here? Should I have opened an issue asking if this was wanted, first?

thepassle:
Should we add a note about using decorators? 
https://open-wc.org/building/building-webpack.html#common-extensions

> Some extensions add non-native syntax to your code, which can be bad for maintenance longer term. We suggest avoiding adding plugins for using experimental javascript proposals or for importing non-standard module types.

daKmoR:
@thepassle you mean we should add the warning for every plugin? if needed we can do so in a separate PR 👍

daKmoR:
thxxxx ❤️ 

fyi: I change the commit message to `fix(building-webpack): add decorator friendly .babelrc example` as it this will release a new version (to update the npmjs page) and with every release we should have at least one commit showing up in the changelog (and docs does not)

thepassle:
Well its a bit strange if we say "we dont recommend using non-native syntax", but then go ahead and in a different place give guides on how to add non-native syntax. Thats gonna send mixed messages. We should at least put a little disclaimer/notification; "We do not recommend non-standard syntax(link to rationale?), however, if you do want to use it... etc etc"

daKmoR:
fully agree - it should be super transparent that we definitely do not recommend non-native syntax 👍

PR #387, opened 2019-04-14
feat(testing-karma): add esm based karma configuration

LarsDenBakker:
I've been battling with getting karma to be less noisy and to report correct files and line numbers one error. I couldn't get this working properly with webpack and rollup.

This takes a different approach, not doing any bundling for tests and only doing the minimal source code transformation to get the tests running.

**Benefits:**
- Serves native es modules, debug the code you wrote
- The browser understands the code we serve much better
- Terminal output is much cleaner
- Errors are reported with correct file + line number reporting
- Clicking on stacktraces in the browser shows the actual files
- Alt + clicking on stacktraces in the terminal jumps to the actual file (in VS code at least)
- Overall simpler and less magic
- Good baseline for future work, such as when chrome supports import maps we can use this in our tests.

**How it works:**
Essentially the only thing we do is run js code through babel to resolve bare imports and to add istanbul coverage instrumentation. Unfortunately when your tests hit the browser all your module imports will 404 because karma does not allow serving arbitrary files. To work around this we install a custom karma middleware, which loads the files.

**speed**
This approach is not necessarily faster for single test runs. For watch mode it is definitely faster though, as not all code needs to be completely re-bundled and re-compiled and we can use karma's watch system to only re-run changed tests. 

For large projects it might be slower due to large waterfall of requests because of HTTP 1. For future iterations we can investigate running the tests with HTTP2. 

**legacy**
In our default config, we use the `legacy` flag to switch between the esm/karma-webpack setup. For legacy runs, we still use karma-webpack.


LarsDenBakker:
Note to self: we need to tell users to add 'type:module' to their files (or maybe do a check in our plugin)

LarsDenBakker:
@daKmoR this is ready for review now. I still need to fix typings and clean up code / add docs. But I'd like to see if this setup works for you.

I've merged the changes into the `testing-karma` package, with a fallback to the existing `karma-webpack` setup when running with the legacy flag.

As mentioned before, this approach is not necessarily faster for single test runs. For watch mode it is potentially faster, as not all code needs to be completely re-bundled and we can use karma's watch system to only re-run changed tests. 

For large projects it might be slower due to large waterfall of requests because of HTTP 1. For future iterations we can investigate running the tests with HTTP2. 

As discussed, this also sets the stage for using import maps. When we have those we can skip babel for running without test coverage which will greatly improve speed. For test coverage we still need some transform.

packages/karma-esm/middleware.js
```diff
@@ -0,0 +1,53 @@
+const fs = require('fs');
+const path = require('path');
+const esm = require('./src/karma-esm');
```
bennypowers: consider using `karmaEsm` for variable name, as in the future you might want to use `esm` package for interop

packages/karma-esm/src/forked-babel-plugin-bare-import-rewrite.js
```diff
@@ -0,0 +1,163 @@
+// @ts-nocheck
+/* eslint-disable */
+'use strict';
+
+/**
+ * Fork of babel-plugin-bare-import-rewrite:
```
bennypowers: maybe use `patch-package` for this?

packages/karma-esm/src/middleware.js
```diff
@@ -0,0 +1,75 @@
+const fs = require('fs');
+const path = require('path');
+const esm = require('./karma-esm');
+
+/**
+ * Middleware for serving es modules. Karma doesn't serve any files not specified
+ * beforehand. This is extremely inconvenient when working with es modules, as many
+ * files are requested on the fly.
```
bennypowers: will this work for components that dynamically import from constructed urls?

like, imagine

```js
connectedCallback() {
  import(`https://cdn.components.thing/${this.someAttr}/component.js`)
    .then(this.doWithModule)
}
```

LarsDenBakker: Yes, because these requests will never reach the web server as the browser will request it from the internet.

packages/karma-esm/src/forked-babel-plugin-bare-import-rewrite.js
LarsDenBakker: I'd rather spend the effort making a PR to the babel plugin. I already did a few PRs, but I'd like to get this change merged so that people can start using it.

packages/karma-esm/middleware.js
LarsDenBakker: Good idea, made the changes.

thepassle:
Should we keep up the naming conventions? eg `testing-karma-esm`?

bennypowers:
> Should we keep up the naming conventions? eg `testing-karma-esm`?

I'm of the general opinion that es modules should be treated as the standard, so, all other things being equal, I'd prefer `testing-karma` for esm and `testing-karma-legacy` for cjs

LarsDenBakker:
`karma-esm` is the plugin to make karma work with es modules. I can't call the plugin just `karma` :)

`testing-karma` uses `karma-esm` by default, and uses `karma-webpack` if the legacy flag is passed.

daKmoR:
- set version of karma-esm to 0.1.0 (breaking change release will bump it to 1.x)
- made it 2 commits
- 1 breaking for testing-karma & karma-esm
- 1 feat to upgrade to latest testing-karma (=> so it's breaking for testing)

It will result in the following version bumps:
```
 - @open-wc/chai-dom-equals: 0.11.7 => 0.12.0
 - @open-wc/create: 0.9.5 => 0.10.0
 - @open-wc/karma-esm: 0.1.0 => 1.0.0
 - @open-wc/semantic-dom-diff: 0.10.7 => 0.11.0
 - @open-wc/testing-helpers: 0.8.10 => 0.9.0
 - @open-wc/testing-karma-bs: 1.0.6 => 1.1.0
 - @open-wc/testing-karma: 1.1.2 => 2.0.0 
 - @open-wc/testing: 0.11.7 => 0.12.0
```

so `@open-wc/testing`  & `@open-wc/testing-karma` require a manual update 

I think that looks good 👍 @LarsDenBakker pls do a final check before merging 🤗

daKmoR:
oh sorry forgot to change the author of the commits 🙈 pls do so - or I can do it tomorrow evening

Issue #390, opened 2019-04-16
Building generator is broken.


jorenbroekema:
https://github.com/open-wc/open-wc/tree/master/packages/building-webpack  installation instructions are wrong. Upgrade -> Building -> Webpack is not a thing. You can only select Upgrade -> Building, and then nothing happens.

also, `npm init @open-wc building-webpack` used to work 2 weeks ago but now I get:

```
$ npm init @open-wc building-webpack
npx: installed 29 in 15.997s
C:\<user>\AppData\Roaming\npm-cache\_npx\15292\node_modules\@open-wc\create\node_modules\command-line-args\dist\index.js:1345
        throw err
        ^

UNKNOWN_VALUE: Unknown value: building-webpack
    at commandLineArgs (C:\<user>\AppData\Roaming\npm-cache\_npx\15292\node_modules\@open-wc\create\node_modules\command-line-args\dist\index.js:1342:21)
    at Object.<anonymous> (C:\<user>\AppData\Roaming\npm-cache\_npx\15292\node_modules\@open-wc\create\dist\generators\app\index.js:72:48)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (C:\<user>\AppData\Roaming\npm-cache\_npx\15292\node_modules\@open-wc\create\dist\create.js:12:14)
```
Additionally the manual setup doesn't work out of the box either:
```
npm i -D @open-wc/building-webpack webpack http-server webpack-dev-server
```
because it should say `webpack-cli` instead of `webpack` .

Actually I found another issue:
```js
const { resolve } = require('path');
const createDefaultConfig = require('@open-wc/building-webpack/modern-and-legacy-config');

// If you don't need IE11 support, use the modern-config instead
// import createDefaultConfig from '@open-wc/building-webpack/modern-config';

module.exports = createDefaultConfig({
  input: path.resolve(__dirname, './src/index.html'),
});
```
This is the code snippet from the manual setup from the docs, the very first line should be `const path = require('path');`

[...]

PR #394, opened 2019-04-16
fix(building-webpack): correct manual install steps

daKmoR:
closes https://github.com/open-wc/open-wc/issues/390

### DIFF
--- a/packages/building-webpack/README.md
+++ b/packages/building-webpack/README.md
@@ -18,14 +18,14 @@ npm init @open-wc
 
 1. Install the required dependencies:
 ```bash
-npm i -D @open-wc/building-webpack webpack http-server webpack-dev-server
+npm i -D @open-wc/building-webpack webpack webpack-cli webpack-dev-server http-server
 ```
 
 2. Create a file called `webpack.config.js` and pass in your app's js entry point and index.html.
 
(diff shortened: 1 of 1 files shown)


packages/building-webpack/README.md
```diff
@@ -18,7 +18,7 @@ npm init @open-wc
 
 1. Install the required dependencies:
 ```bash
-npm i -D @open-wc/building-webpack webpack http-server webpack-dev-server
+npm i -D @open-wc/building-webpack webpack-cli http-server webpack-dev-server
```
LarsDenBakker: Webpack itself is still also needed

packages/building-webpack/README.md
```diff
@@ -32,7 +32,7 @@ const createDefaultConfig = require('@open-wc/building-webpack/modern-and-legacy
 // import createDefaultConfig from '@open-wc/building-webpack/modern-config';
 
 module.exports = createDefaultConfig({
-  input: path.resolve(__dirname, './src/index.html'),
+  input: resolve(__dirname, './src/index.html'),
```
LarsDenBakker: I do think its path.resolve?

LarsDenBakker: Oh i see now. We should adjust the require, we never destructure it in any other examples.

Issue #385, opened 2019-04-14
[docs] add non standard syntax warning


thepassle:
https://open-wc.org/building/building-webpack.html#_1-babel

- Add non standard syntax warning: "We do not recommend non-standard syntax(link to rationale?), however, if you do want to use it... etc etc"
- remove backtics
- change: "Supporting `LitElement`'s **`@customElement` and `@property`**" to "Supporting LitElement's **decorators**"

LarsDenBakker:
There is already a warning above this section:
```
Some extensions add non-native syntax to your code, which can be bad for maintenance longer term. We suggest avoiding adding plugins for using experimental javascript proposals or for importing non-standard module types.
```

Agreed we should change the name. I think it can be even more generic saying something like 'supporing decorators'.

PR #402, opened 2019-04-23
docs(building-webpack): add non standard syntax warning

thepassle:
fixes https://github.com/open-wc/open-wc/issues/385

### DIFF
--- a/packages/building-webpack/README.md
+++ b/packages/building-webpack/README.md
@@ -223,7 +223,11 @@ This the fastest method, as it strips away types during babel transformormation 
 
 
 <details>
-  <summary>Supporting `LitElement`'s `@customElement` and `@property`</summary>
+  <summary>Supporting decorators</summary>
+
+::: warning
+Please note that our default recommendations do not support the use of decorators as adding non-native syntax to your code can be bad for maintenance longer term. If you do, however, want to add decorators, you can follow this section.
+:::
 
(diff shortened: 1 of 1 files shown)


packages/building-webpack/README.md
```diff
@@ -223,7 +223,11 @@ This the fastest method, as it strips away types during babel transformormation
 
 
 <details>
-  <summary>Supporting `LitElement`'s `@customElement` and `@property`</summary>
+  <summary>Supporting decorators</summary>
+
+::: warning
+Please note that our default recommendations do not support the use of decorators as adding non-native syntax to your code can be bad for maintenance longer term. If you do, however, want to add decorators, you can follow this section.
```
LarsDenBakker: We already have the same warning on top of this section. Do you think we should print it twice? Maybe we should restructure this section to differentiate between non-standard syntax and 'build jobs' like copy?

thepassle: I think if we're going to show people how to add non standard syntax to their code, we should add a disclaimer right there in that place that says we dont recommend it. Maybe we can make it smaller? Like:

```
::: warning
Please note that decorators will add [non standard syntax](https://open-wc.org/building/building-webpack.html#common-extensions) to your code.
```

thepassle:
Shortened the message to:

```
::: warning
Please note that decorators will add [non standard syntax](https://open-wc.org/building/building-webpack.html#common-extensions) to your code.
```

PR #405, opened 2019-04-27
docs(linting): suggest to add eslint dev dependency on eslint manual setup

blikblum:
Needed to run lint scripts

### DIFF
--- a/packages/eslint-config/README.md
+++ b/packages/eslint-config/README.md
@@ -15,7 +15,7 @@ This is part of the default [open-wc](https://open-wc.org/) recommendation
 :::
 
 ## Manual
-- `yarn add --dev @open-wc/eslint-config`
+- `yarn add --dev eslint @open-wc/eslint-config`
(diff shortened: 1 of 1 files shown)


daKmoR:
`@open-wc/config-eslint` has a dependency on `eslint` so it should not be needed in the root? 

https://github.com/open-wc/open-wc/blob/master/packages/eslint-config/package.json#L22

am I missing something? 🤔

blikblum:
> am I missing something?

After i installed @open-wc/config-eslint and tried to run the lint script and got "eslint not found" error. In windows 10

thepassle:
Just tried this on my windows 10 machine, i can't reproduce, its working as intended for me and I dont have to separately install eslint

LarsDenBakker:
Transitive dependencies dont work in that way with on npm we have seen that before. If you require or use something it needs to be a direct dependency. Otherwise npm can scope packages if multiple dependencies need eslint for example.

blikblum:
Sorry the exact step i did was to install @open-wc/prettier-config. When i saw @open-wc/eslint-config in the eslint config file suggested in prettier lint page i think that was a dependency from @open-wc/prettier-config which is not true.

Closing this.

But should not @open-wc/eslint-config be a dependency of  @open-wc/prettier-config?

PR #408, opened 2019-04-28
fix(eslint-config): loosen up rules for test and stories files

daKmoR:
disable
- `no-console` => useful especially in stories to show actions
- `no-unused-expressions` => mostly for tests to use `.to.be.true;`
- `class-methods-use-this` => mostly for tests when mocking

for test and stories files...

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -19,7 +19,7 @@ module.exports = {
   },
   overrides: [
     {
-      files: ['**/test/**/*.js', '**/stories/**/*.js'],
+      files: ['**/test/**/*.js', '**/demo/**/*.js', '**/stories/**/*.js'],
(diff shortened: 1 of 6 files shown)


packages/eslint-config/index.js
```diff
@@ -17,4 +17,14 @@ module.exports = {
       },
     ],
   },
+  overrides: [
+    {
+      files: ['**/test/**/*.js', '**/stories/**/*.js'],
```
LarsDenBakker: Can we add `'**/demo/**/*.js'` here?

LarsDenBakker:
```suggestion
      files: ['**/test/**/*.js', '**/demo/**/*.js', '**/stories/**/*.js'],
```

PR #431, opened 2019-05-03
chore(create): remove plain browser tests

LarsDenBakker:
We moved away from using mocha directly in the browser, instead advising to use karma in debug mode as the developer experience is better.

If you agree, we should remove it from the init as well.


LarsDenBakker:
@daKmoR would like your opinion on this one

daKmoR:
I agree the index.html is no longer needed... all your testing needs should be covered by using test or test:watch

Issue #426, opened 2019-05-03
Improve developer ergonomics around `.browserslistrc`


LarsDenBakker:
Our build configs expect a `.browserslistrc` to be present, but fails silently a user doesn't have one. If there is none, it will default to compiling to es5 which breaks many assumptions of our build output.

Ideally we just throw an error if the file isn't there, but there are different ways to define a `..browserslistrc`, such as in your babel config or package.json. 

It would also be good to just set the latest two chrome/firefox/safari/edge as a default, since that's what most people need anyway.

PR #437, opened 2019-05-07
Feat/browserslist

LarsDenBakker:
Fixes https://github.com/open-wc/open-wc/issues/426

- Devs no longer need to always provide a browserslist configuration. If none is given, the default is taken.
- Updated instructions on how to add your own browserslist configuration
- Add warning if adding IE11 in your browserslist configuration (this would cause a broken es5 build as we don't load es2015 polyfills)


thepassle:
LGTM

PR #441, opened 2019-05-08
FIX: install new testing-karma with generator

tpluscode:
The testing generator installs `1.x` version of `@open-wc/testing-karma` which does not properly handle the new syntax for karma files.

It causes IE11 tests to fail silently unless that package is updated to 2+


daKmoR:
thxxx :heart:

PR #443, opened 2019-05-08
feat: use latest eslint-plugin-lit

thepassle:


LarsDenBakker:
Should we also configure their new rules they added?

thepassle:
we use the recommended:
```
  extends: ['@open-wc/eslint-config', 'eslint-config-prettier', 'plugin:lit/recommended']
```

PR #447, opened 2019-05-12
feat(owc-dev-server): add watch mode

LarsDenBakker:
A minimal implementation for reloading the browser on file changes. Uses [chokidar](https://github.com/paulmillr/chokidar) for watching files, and injects as small script to open a [Server sent events](https://www.html5rocks.com/en/tutorials/eventsource/basics/) connection to the server to be notified about updates.

This does not work in IE11 or Edge, but these are not browsers typically used to develop with.

I like this better than browser sync, because browser sync does a lot more than just reload the browser on file change. It also syncs multiple browsers on the same web server, so that scrolling and clicking is synced. For example to do large scale testing in a large device lab.

I refactored the code a bit so the diff looks larger than it is.


thepassle:
Very nice :) LGTM!

packages/owc-dev-server/owc-dev-server.js
```diff
@@ -126,26 +33,92 @@ if (modulesDir) {
...
+function serveIndexHTMLFallback(req, res, next) {
+  if (!req.url.includes('.')) {
+    serveIndexHTML(req, res, next);
+  } else {
+    next();
+  }
+}
+
+/** Adds any served files to file watcher, reloading the browser on change */
+if (watch) {
```
bennypowers: Docs say that watch requires appIndex, so cli should exit with an error if it is not set

LarsDenBakker: Great point

PR #448, opened 2019-05-12
chore: split babel compiler code from karma-esm

LarsDenBakker:
split out code from `karma-esm` into `building-utils` so that it can be used by `owc-dev-server`


daKmoR:
`express-transform-bare-module-specifiers` expects outdated types or `lru-cache` types are not fully up to date?

```
node_modules/express-transform-bare-module-specifiers/dist/compile-middleware.d.ts:7:45 - error TS2694: Namespace 'LRUCache' has no exported member 'Cache'.

7 export declare const babelCompileCache: LRU.Cache<string, string>;
```

what to do? find an alternative? fork?

LarsDenBakker:
I don't understand why we are type checking dependencies. This shouldn't be the case.

LarsDenBakker:
I was able to fix this by adding the outdated types to the package we use bare module specifier transform.

PR #452, opened 2019-05-12
chore: fix favicon

thepassle:
`npm run vuepress:build` didnt pick up the favicon file 😒


docs/.vuepress/config.js
```diff
@@ -112,7 +112,7 @@ module.exports = {
   plugins: ['@vuepress/google-analytics'],
   ga: 'UA-131782693-1',
   head: [
-    ['link', { rel: 'icon', type: 'image/png', href: '/favicon.png' }],
+    ['link', { rel: 'icon', type: 'image/png', href: '/public/favicon.png' }],
```
daKmoR: preview says no 😅 

Icon is here:
https://deploy-preview-452--open-wc-org.netlify.com/favicon.png

so placing it in public was a good idea... but that path does not need to be reflected in this href

thepassle: Doh.

PR #455, opened 2019-05-13
fix: add syntax highlighting

newbornfrontender:


LarsDenBakker:
Thanks for your PR. I think this is a script that should be evaluated, it's not part of  the visible readme. But something seems to be going wrong @daKmoR any idea?

daKmoR:
thx but this seems to be working as intended for our url

https://open-wc.org/building/building-rollup.html#import-css-files-in-lit-html

you see the script on github as you can't comment it out for github 🙈  or at least I could not find a way... (there may be other solutions like "hacking" vuepress generation of the edit page button... so if anyone want's to take a stab at it 😬)

However, I'm afraid we can not accept this 🙈 still a good reminder 👍  that on github it's not looking as nice as it should be 🙈

PR #460, opened 2019-05-17
chore(demoing-storybook): add knob exports

jorenbroekema:
Storybook has a couple of knobs that are cool but not re-exported in open-wc, so here's a MR :).


daKmoR:
thaaannkk you ❤️

PR #461, opened 2019-05-19
feat(owc-dev-server): add HTTP2 option

LarsDenBakker:
Adds option to serve with HTTP2 rather than HTTP1. This will improve speed while developing locally, as we are serving unbundled which can lead to large request waterfalls on HTTP1.

It also make `owc-dev-server` suitable for giving a more realistic representation of performance of a fully built application.

HTTP2 requires HTTPS which makes this a bit more complex. I found https://www.npmjs.com/package/selfsigned which lets us generate a certificate from node, but the user can provide their own certificate as well. The generated certificate is cached for performance, and to avoid repeatedly needing to accept the certificate in the browser. 

I am very curious to see if this works on different OSs, browsers etc.

For now the main performance benefit is request multiplexing, HTTP1 is limited to only a few concurrent requests. We can later look into server push, but I don't know if that's worth the effort for development.


daKmoR:
tried it via Windows 10 WSL

I can start it and it opens the browser and I get the "insecure" page thing - when I say "open anyways" then `owc-dev-server` just stops with the following... 

I tried to debug but it seems to be an internal error of spdy?

```
$ yarn start
yarn run v1.15.2
$ ./owc-dev-server.js --app-index demo/index.html --http2 --open --watch
owc-dev-server started on https://localhost:8080
  Serving files from './'.
  Opening browser on '/demo/'
  Using history API fallback, redirecting requests to 'demo/index.html'
  Using HTTP2 and HTTPS
internal/buffer.js:788
class FastBuffer extends Uint8Array {}
^

RangeError: Invalid typed array length: -4095
    at new Uint8Array (<anonymous>)
    at new FastBuffer (internal/buffer.js:788:1)
    at Handle.onStreamRead [as onread] (internal/stream_base_commons.js:144:17)
    at Stream.<anonymous> (/mnt/c/html/open-wc/open-wc/node_modules/handle-thing/lib/handle.js:88:10)
    at Stream.emit (events.js:193:15)
    at endReadableNT (/mnt/c/html/open-wc/open-wc/node_modules/spdy-transport/node_modules/readable-stream/lib/_stream_readable.js:1077:12)
    at processTicksAndRejections (internal/process/next_tick.js:76:17)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

any idea? what I could try?

LarsDenBakker:
That's unfortunate. Which version of node are you using?

LarsDenBakker:
Looks to be related to this issue https://github.com/spdy-http2/node-spdy/issues/350

daKmoR:
```
$ node --version
v11.7.0
```

LarsDenBakker:
Looking at the linked issue, it breaks on node 11 and it sounds like they're not going to fix it. Back to the drawing board then. Thanks for testing!

LarsDenBakker:
Superseded by https://github.com/open-wc/open-wc/pull/520

Issue #463, opened 2019-05-20
Generated demo app uses incorrect property for the sample component attribute 'heading'


harunhasdal:
The demo app uses `header` while the component declares `heading`

https://github.com/open-wc/open-wc/blob/4ec95baa487b9718d9143b7ac81540f887800af3/packages/create/src/generators/wc-lit-element/templates/static/demo/index.js#L6

https://github.com/open-wc/open-wc/blob/4ec95baa487b9718d9143b7ac81540f887800af3/packages/create/src/generators/wc-lit-element/templates/_MyEl.js#L16

daKmoR:
thank you for fixing it :+1:

PR #464, opened 2019-05-20
fix(create): use correct attribute name in demo (#463)

harunhasdal:
Fix for #463


daKmoR:
thaannkkk you :heart:

PR #467, opened 2019-05-25
Fix/css minify

LarsDenBakker:
CSS was not properly minified in some cases. Passing an explicit flag fixes this.


daKmoR:
LGTM 🤗

PR #468, opened 2019-05-25
Fix/template literal

LarsDenBakker:
Safari 12 has a bug where it garbage collects template literals in some conditions. This is not a problem for most use cases, and lit-html handles it properly. It's also already fixed in safari 12.1. 

This bug causes `preset-env` to compile template literals to an array of strings, which is a shame for the modern build. My proposal is disabling this manually for now.

See https://github.com/Polymer/lit-html/issues/575 for more info.


LarsDenBakker:
@daKmoR @thepassle @Westbrook @bennypowers any opinions on this?

LarsDenBakker:
Yes that's a good point

bennypowers:
For context, caniuse.com/usage-table lists usage for safari (macOS)

version|usage
-----|-----
12|1.19%
12.1|0.81

it does not list any usage for iOS Safari 12, only 12.1. Is there an iOS Safari 12?

PR #473, opened 2019-06-01
feat: include all static modules in `entryModules`

### LINKED ISSUE #472: [building][rollup] Modules imported to main module not appended at `rel=preload`
When building for modern and legacy and relying on dynamic imports, it is possible to for Rollup to create a chunk that is required in both the main module and the dynamically imported module, in this case the chunk is not registered as `isEntry` and will not be added to the HTML via `<link rel="preload" />`.

*Reproduction steps:*
1. Clone https://github.com/Westbrook/minesweeper/tree/2bec691f54708e3bb6eb83755139ffbd144c9a20
2. Install with `yarn`
3. Build `yarn build`
4. See lack of secondary `<link rel="preload" />`

*Expected results:*
The `<link rel="preload" />` is available and ensures the earliest request for this asset.

*Actual results:*
There is no `<link rel="preload" />` and the developer needs to add it manually for optimal performance.

Westbrook:
This will benefit first load experience by not requiring the browser to traverse the module graph before initiating the request of these files.

fixes #472

### DIFF
--- a/packages/building-rollup/plugins/rollup-plugin-modern-web/rollup-plugin-modern-web.js
+++ b/packages/building-rollup/plugins/rollup-plugin-modern-web/rollup-plugin-modern-web.js
@@ -228,8 +228,19 @@ module.exports = (_pluginConfig = {}) => {
     // Injects generated module paths into index.html
     generateBundle(outputConfig, bundles) {
       const entryModules = Object.keys(bundles)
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
You need to include the commit scope so that the changelog for the right packages gets updated. I changed it to:

`feat(building-rollup): include all static modules in `entryModules` (#473)`

Issue #435, opened 2019-05-04
modern-and-legacy-config not save correct index.html - missing preload and polyfills


mjanos1:
Hi,

this issue is happen for my project and for clean instalation of webpack created by https://github.com/open-wc/open-wc/tree/master/packages/building-webpack

This happen very randomly, but almost everytime when run build on docker, and in some case when run build under win 10.
```
"build": "webpack --mode production",
```

When I try to debug this issue, html-webpack-plugin do correct changes for index.html, but it is not saved in dist/index.html.
What I can observe, ```modern-and-legacy-config.js``` randomly start build with legacy build  ```createConfig(options, true)``` and then with modern build ```createConfig(options, false)```, in this case ```index.html``` was correctly build.
In ```modern-and-legacy-config.js```  is ```[createConfig(options, false), createConfig(options, true)]```

project config:

webpack.config.js:
```
/* eslint-disable import/no-extraneous-dependencies */

const { resolve } = require('path');
const createDefaultConfig = require('@open-wc/building-webpack/modern-and-legacy-config');
const merge = require('webpack-merge');
const CopyPlugin = require('copy-webpack-plugin');

const defaultConfiguration = createDefaultConfig({
  indexHTML: resolve(__dirname, './src/index.html'),
  entry: resolve(__dirname, './src/index.js'),
});

const [defaultModernConfig, defaultLegacyConfig] = Array.isArray(defaultConfiguration)
  ? defaultConfiguration
  : [defaultConfiguration, defaultConfiguration];

const customConfig = {
  devServer: {
    contentBase: resolve(__dirname, 'dist'),
    historyApiFallback: true,
    https: true,
  },
  plugins: [
    new CopyPlugin([
//copy fonts and css to assets folder
    ]),
  ],
};


module.exports = [
  merge(defaultModernConfig, customConfig),
  merge(defaultLegacyConfig, customConfig),
];
```
src/index.html:
```
<!doctype html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
    <meta name="Description" content="">
    <base href="/">
<!-- Import css files from assets folder -->
</head>
<body>
    <div id="auth"></div>
    <div id="breadcrumbs"></div>
    <div id="root"></div>
    <div id="splash">
        <div class="splash-logo-wrap">
            <svg>
<!-- SVG detail -->
           </svg>
        </div>
    </div>
</body>

</html>
[...]

PR #474, opened 2019-06-02
feat: utils and webpack plugin for an index.html entrypoint

LarsDenBakker:
This is part of a larger refactoring. I'm extracting all common code between different build related projects so that they can be reused. For example `building-rollup`, `building-webpack` and `owc-dev-server` all need to extract modules from a `index.html`. 

This also separate the `modern web plugin` as `webpack-index-html-plugin`, so that it can be used separately by power users (I actually want to reuse it separately myself for my project at work).

For more information you can check the readme of `webpack-index-html-plugin`

In this PR I am only updating `building-webpack`, later on I will update the other projects.

Fixes https://github.com/open-wc/open-wc/issues/435

Fixes https://github.com/open-wc/open-wc/issues/475

Fixes https://github.com/open-wc/open-wc/issues/435


LarsDenBakker:
This is ready for review

packages/webpack-index-html-plugin/webpack-index-html-plugin.js
```diff
@@ -0,0 +1,129 @@
...
+      }
+
+      if (config.multiIndex.variations.some(v => !v.name)) {
+        throw createError(`Some variations are missing a name property.`);
+      }
+    }
+  }
+
+  apply(compiler) {
+    // @ts-ignore
```
daKmoR: added this ignore... is this if needed? (it is nowhere set... does it comes from webpack directly?)

daKmoR:
made some changes and a force push 😬 
- rebased
- made it 2 commits
- use babel 7.3.3 version everywhere
- add some eslint ignore for demos
- set version of the webpack plugin to 0.0.1 (will be bumped to 0.1.0 when released)
- added one comment for a ts-ignore (but that is not blocking)

very nice work 💪 

so LGTM => feel free to merge 🤗 

PS: it seems like something broke vuepress :/ it can't find some of the core-js things... let's fix that in a separate MR... so homepage will not update until that is fixed 🙈

daKmoR:
hmmm scratch that remark for `vuepress`... it works fine on master 😱  
sooo looking at the changes I am all about - "whaatt the hell could break it?" 🤔

packages/building-utils/package.json
```diff
@@ -8,23 +8,35 @@
...
   "files": [
     "src",
+    "index-html",
+    "dom5-fork",
     "*.js"
   ],
   "dependencies": {
-    "@babel/core": "^7.4.4",
+    "@babel/core": "^7.3.3",
+    "@types/deepmerge": "^2.2.0",
```
daKmoR: > This is a stub types definition. deepmerge provides its own type definitions, so you do not need this installed.
?

daKmoR: I removed it

daKmoR:
ok, I restored the `yarn.lock` file from master and then did an install which now only add the new dependencies...
=> doing that restored vuepress
=> and also the [preview](https://deploy-preview-474--open-wc-org.netlify.com/) is back 🤗 

so the lock currently saves our ass 🙈  at some point we will need to update everything and make sure it still works... but not today 😬

packages/webpack-index-html-plugin/webpack-index-html-plugin.js
LarsDenBakker: Yes apply comes from webpack

Issue #476, opened 2019-06-03
[webpack-import-meta-loader] Not all backslashes are transformed on Windows


tmsns:
When using the webpack loader to rewrite `import.meta`, not all backslashes are being rewritten. This is aspecially breaking on Windows. The issue seems to be in [webpack-import-meta-loader.js#L23](https://github.com/open-wc/open-wc/blob/master/packages/webpack-import-meta-loader/webpack-import-meta-loader.js#L23): `.replace('\\', '/');` should become: `.replace(/\\/g, '/');`

Didn't have the time yet to setup an environment and create a test though.. :-/ Hope someone can help out here. :-)

LarsDenBakker:
I think rather it should be: `.replace(path.sep, '/');`? Are you able to test this?

tmsns:
We can use `path.sep`, but without the global flag, only the first occurrence of the separator will be replaced.

LarsDenBakker:
`new RegExp(path.sep, 'g')` ? :)

I don't have a windows machine, and we don't run nodejs tests in the CI. So these things are a big blind spot for us, which is definitely a problem we need to find a solution for because this path issue keeps coming up

LarsDenBakker:
Sorry I slightly misunderstood, the original code was already only handling backslashes you just wanted to make it global which is indeed correct!

The original code is actually wrong though because on unix I could have a file or folder with a backslash in it and it would be rewritten to a forward slash. So let's change it to `new RegExp(path.sep, 'g')`

PR #477, opened 2019-06-03
fix(webpack-import-meta-loader): rewrite all backslashes

LarsDenBakker:
Fixes https://github.com/open-wc/open-wc/issues/476

### DIFF
--- a/packages/webpack-import-meta-loader/src/to-browser-path.js
+++ b/packages/webpack-import-meta-loader/src/to-browser-path.js
@@ -1,0 +1,5 @@
+const path = require('path');
+
+module.exports = function toBrowserPath(filePath, _path = path) {
+  return filePath.replace(new RegExp(_path.sep === '\\' ? '\\\\' : _path.sep, 'g'), '/');
+};
--- a/packages/webpack-import-meta-loader/test/to-browser-path.test.js
+++ b/packages/webpack-import-meta-loader/test/to-browser-path.test.js
@@ -1,0 +1,14 @@
+const { expect } = require('chai');
+const toBrowserPath = require('../src/to-browser-path');
+
+describe('toBrowserPath', () => {
+  it('replaces all path windows separators', () => {
+    // @ts-ignore
+    expect(toBrowserPath('foo\\bar\\buz', { sep: '\\' })).to.equal('foo/bar/buz');
+  });
+
+  it('preserves all unix path separators', () => {
+    // @ts-ignore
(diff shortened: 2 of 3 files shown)


tmsns:
Wanted to add a specific test for it:

```
  it('Replaces all backslashes', async () => {
    const stats = await compiler('caseB\\caseBsub\\caseBsub.js', rules);
    const caseBsub = stats.toJson().modules[0].source;

    expect(caseBsub).to.equal(
      `${'' +
        "export const bar = new URL('./', ({ url: `${window.location.protocol}//${window.location.host}/caseB/caseBsub/caseBsub.js` }).url);"}${newLine}`,
    );
  });
```

But Karma doesn't want to find the test. I did see the exclusion in the Karma config [karma.conf.js#L13](https://github.com/tmsns/open-wc/blob/master/karma.conf.js#L13), but also the with grep function, I can't find it.. :-/
`yarn run test --grep packages/webpack-import-meta-loader/test/webpack-import-meta-loader.test.js`

LarsDenBakker:
These tests are run in node, not in in the browser. That's why they're excluded. You can run them by moving into the `packages/webpack-import-meta-loader` and running `npm run test` or `yarn test`

tmsns:
> These tests are run in node, not in in the browser. That's why they're excluded. You can run them by moving into the `packages/webpack-import-meta-loader` and running `npm run test` or `yarn test`

😅 🙈 God... I guess it's been a long day..

tmsns:
Hmmm.. ok.. The test is not working how I expected it to work. `compiler` is using webpack, so backslashes are already transformed in the tests. Don't know if it'll be doable to test the specific behavior. 🤔 

On a sidenote, by using `import { EOL as newLine } from 'os';` the test expects the files to be using the specific EOL chars of the system. I'm using git's `core.autocrlf=false` so no auto conversion.. hence, breaking tests. 🙈

LarsDenBakker:
Does the change work properly for you locally though? Will look into a proper setup for testing these things cross platform later.

tmsns:
Normally yes, as I tested the `/\\/g` variant earlier today by changing it in `node_modules` directly. So, I expect it to work just fine. 👍

LarsDenBakker:
Made a test and checked with @thepassle that it's working

tmsns:
Yeah, the mocked `path` module will do the trick! Of course, there is no guarantee that `toBrowserPath` is actually used in the plugin, but it's always tricky when using mocks. [proxyquire](https://www.npmjs.com/package/proxyquire) seems to be helpful.

PR #487, opened 2019-06-10
Fix(building-rollup): add proper typescript support

kasperstorgaard:
I played around with the rollup setup for typescript, and after a lot of frustration, discovered that the rollup setup doesn't really support typescript (the demo code is just es6 with classes).

To reproduce this, add a type to the demo file, like this: `render(): TemplateResult {`,
and then run the ts-babel demo: `npm run build:ts-babel`.
You will then get the following error:
`[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)`.

To fix this, all that is needed is to add `extensions` option to `node-resolve` and `babel` plugins.

Not sure what the best approach would be for adding them, but thinking options?

TODO: docs when an approach has been picked

### DIFF
--- a/packages/building-rollup/README.md
+++ b/packages/building-rollup/README.md
@@ -289,7 +289,7 @@ You also need to specify `.ts` in the `extensions` option, for babel and node to
 ```
 const configs = createDefaultConfig({
   input: './index.html',
-  extensions: ['.js', '.ts'],
+  extensions: ['.js', '.mjs', '.ts'],
 });
 ```
 (keep `.js` in there, since node will want to resolve javascript files in node_modules)
--- a/packages/building-rollup/README.md
+++ b/packages/building-rollup/README.md
@@ -285,6 +285,14 @@ We recommend using the babel typescript plugin. Add it to your babel config file
   ],
 }
 ```
+You also need to specify `.ts` in the `extensions` option, for babel and node to properly recognize ts files:
+```
+const configs = createDefaultConfig({
+  input: './index.html',
+  extensions: ['.js', '.ts'],
+});
+```
+(keep `.js` in there, since node will want to resolve javascript files in node_modules)
 
 This is the fastest method, as it strips away types during babel transformation of your code. It will not perform any type checking though. We recommend setting up the type checking as part of your linting setup, so that you don't need to run the typechecker during development for faster builds.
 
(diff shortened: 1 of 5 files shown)


thepassle:
Did this https://open-wc.org/building/building-rollup.html#support-typescript not work for you?

kasperstorgaard:
No it didn't work, but it took me a long time to figure out why :)
Just add a type to the render method and test it out.
`render(): TemplateResult {`

kasperstorgaard:
somewhat related to
https://github.com/rollup/rollup-plugin-babel/issues/255

LarsDenBakker:
Looks like a good change then. Thanks for updating the demo as well, sorry we didn't cover this properly.

Could you update the readme to cover this? Also could you add `.mjs` to the defaults? I'm not sure if `resolve` or `babel` takes these in by default but I want to make sure we don't ignore those.

LarsDenBakker:
Also do a `npm run format` to make the CI pass

kasperstorgaard:
working on it, but it seems that the `ts` demo does not need the same options, so that might be slightly confusing? (it actually breaks if I add the option and you specified it for that demo)

packages/building-rollup/README.md
```diff
@@ -285,6 +285,14 @@ We recommend using the babel typescript plugin. Add it to your babel config file
   ],
 }
 ```
+You also need to specify `.ts` in the `extensions` option, for babel and node to properly recognize ts files:
+```
+const configs = createDefaultConfig({
+  input: './index.html',
+  extensions: ['.js', '.ts'],
```
LarsDenBakker:
```suggestion
  extensions: ['.js', '.mjs', '.ts'],
```

packages/building-rollup/demo/ts-babel/rollup.config.js
```diff
@@ -2,4 +2,5 @@ const createDefaultConfig = require('../../modern-and-legacy-config');
 
 module.exports = createDefaultConfig({
   input: './demo/ts-babel/index.html',
+  extensions: ['.js', '.ts'],
```
LarsDenBakker:
```suggestion
  extensions: ['.js', '.mjs', '.ts'],
```

LarsDenBakker:
Nice catch on the `DEFAULT_SUGGESTIONS` 👍 you're right we only need this for the babel-typescript setup in the readme.

kasperstorgaard:
anything else need for this PR? Is the wording fine for the readme?

daKmoR:
thaaankkk you for this contribution ❤️

PR #491, opened 2019-06-13
fix(building-rollup): make ts builds work for modern ts-babel

kasperstorgaard:
Recently there was a change to fix `ts-babel` builds https://github.com/open-wc/open-wc/pull/487
That change was only affecting `modern-and-legacy`, but not `modern`.

It should have been part of the first pr, but might also highlight some code duplication between the two 😄


LarsDenBakker:
Thanks. You're right they're duplicated, but that's intentional to keep the configs readable. I'd like any person to come in any copy paste it for their own use.

kasperstorgaard:
Ok, that makes sense

PR #496, opened 2019-06-14
fix(polyfills-loader): patch unpkg resolution error by specifying 'module'

atla5:
problem found by @daKmoR when bugfixing unpkg imports [[codepen](https://codepen.io/daKmoR/pen/ydeaQQ)]

### DIFF
--- a/packages/polyfills-loader/package.json
+++ b/packages/polyfills-loader/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@open-wc/polyfills-loader",
-  "version": "0.3.2",
+  "version": "0.3.1",
   "main": "polyfills-loader.js",
   "module": "polyfills-loader.js",
   "description": "Load web component polyfills using dynamic imports",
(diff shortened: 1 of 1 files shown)


packages/polyfills-loader/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "@open-wc/polyfills-loader",
-  "version": "0.3.1",
+  "version": "0.3.2",
```
daKmoR: no need to bump version - CI takes care of it

```suggestion
  "version": "0.3.1",
```

LarsDenBakker:
What are you using the polyfills loader for? We generally don't have it in any of our recommendations anymore.

atla5:
I'm likely just following a previous recommendation. I've been working on a repo consuming it for some time now and didn't get the memo this was outmoded/unnecessary. 

is there anything replacing it? and if so, what's the new suggested makeup of the `main` entrypoint file collecting all of your web components?

daKmoR:
still makes sense to fix it... but yeah we should deprecate it as well

LarsDenBakker:
@atla5 we add the polyfills in the build, so you just write your code as you normally would.

Nothing wrong with the polyfills loader though, we just found a more efficient way to do it. You can take a look at `npm init @open-wc` and see how it looks and see if it works for you.

Issue #429, opened 2019-05-03
[init] difference between starter app and application is unclear


LarsDenBakker:
In init we prompt users to choose between `Starter app` and `Application` but the difference between them is not clear.

Can we come up with a clearer distinction? Maybe `App boilerplate` or `Basic app` and `Example app` ?

thepassle:
I agree, or just go back to using `Barebone app` and `Starter app` like we had? I think that was pretty clear

LarsDenBakker:
Still need to do something here

thepassle:
I'll take a look at this somewhere this week

PR #497, opened 2019-06-15
fix: improve clarity of init

thepassle:
Fixes https://github.com/open-wc/open-wc/issues/429

### DIFF
--- a/packages/create/README.md
+++ b/packages/create/README.md
@@ -49,6 +49,14 @@ npm init @open-wc
 
 ### Available scaffold generators:
 
+- `Web Component`<br/>
+  This generator scaffolds a starting point for a web component. We recommend using this generator when you want to develop and publish a single web component.
+  <br/>
+
+- `Basic Application`<br/>
+  This generator scaffolds a minimal application for quick prototyping, and does not come with any linting, testing, or building configs.
+  <br/>
+
 - `Starter App`<br/>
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
I see now that basic app doesnt come with any build/test etc. Whats the point then?

thepassle:
Its easy for quick prototyping I guess, also works nicely when using with pika cdn, though I add it manually. I use the barebone app quite a lot actually. Maybe also interesting with the import maps stuff coming up?

LarsDenBakker:
Ah I see. I feel that it adds a lot of confusion though. Starter app gives you choices, right? You could toggle everything off?

thepassle:
Thats true yeah. In that case it may be fine to get rid of it, it made more sense to have it before we had the option toggles in the cli. Im fine with removing it, dont know if master chief @dakmor has any strong feelings about it?

LarsDenBakker:
I'll merge this now, we should revisit the init later.

thepassle:
Yeah we should iron it out a bit more. Im fairly occupied this week (buying a house/moving), so maybe the week after we can have a discussion about it.

I think the starter-app should go. The ‘Application’ without anything selected is basically the ‘Barebones app’ and with everything selected its just the ‘Starter App’. But lets have a chat and make sure.

PR #498, opened 2019-06-15
feat: tool to generate an import map

daKmoR:
ToDo:
- [x] Linting
- [x] Remove quick hack reading package.json with require
- [x] Fix failing tests
- [x] Ask if resolution version can't be found
- [x] If asked save selected version to `package.json`
- [x] Better question? (currently just versions - what would be a good context?)
- [x] Better readme


packages/import-maps-generate/README.md
```diff
@@ -0,0 +1,104 @@
...
+If you only want to try it out once to see what it will generate you can simply do so via:
+
+```bash
+npx @import-map/generate
+```
+
+## Configuration
+
+There may be times where you'll want to apply overrides. For example, if you're using the built-in `std:kv-storage` module and the polyfill for browsers that don't support it, the `kv-storage-polyfill` will be in your yarn.lock, and will be included in your import map. But what you'll actually want is:
+
```
LarsDenBakker: It is not clear here how to solve this problem

daKmoR: jup still working on this 🙈

thepassle: The JSON example down here shows how to solve it right?

packages/import-maps-generate/README.md
```diff
@@ -0,0 +1,104 @@
...
+  // ...
+  "importMap": {
+    "deletes": ["kv-storage-polyfill", "kv-storage-polyfill/"]
+  }
+}
+```
+
+### Overrides
+
+If you want to override a dependency in the importMap with a custom value (eg: you want to get something from a CDN instead), <!-- TODO: better example? --> you can add an `overrides` property:
```
thepassle: I need to find a better example for this probably

daKmoR:
in my current understanding, the import map is "parsed" for every url... e.g. relative/root urls in the "value" will do the following

an import map on `http://example.com/foo/bar` with
- `'lit-html': './node_modules/lit-html/lit-html.js'`
  - http://example.com/foo/bar/node_modules/lit-html/lit-html.js
- `'lit-html': '/node_modules/lit-html/lit-html.js'`
  - http://example.com/node_modules/lit-html/lit-html.js

so for our case the "root" url is the way to go... as we will expect the `node_modules` always at the "host/root" and not in a specific `foo/bar` page folder

LarsDenBakker:
When the URL is `http://example.com/foo/bar` `/foo/bar` is only prepended if there is no `<base href="">`. you need a base element if you want to resolve URLs properly.

For example:

```
<html>
  <head>
    <base href="/foo">
    <script src="bar1.js"></script>
    <script src="./bar2.js"></script>
    <script src="/bar3.js"></script>
  </head>
</html>
```

will request:
- `/foo/bar1.js`
- `/foo/bar2.js`
- `/bar3.js`

This works the same for anchors, imports, images etc.

Respecting the base URL is important for apps which aren't deployed on the root of a domain to work.

packages/import-maps-generate/src/generate.js
```diff
@@ -0,0 +1,22 @@
+import fs from 'fs';
+import path from 'path';
+import commandLineArgs from 'command-line-args';
+import { generateFromYarnLock } from './generateFromYarnLock';
+import injectToHtmlFile from './injectToHtmlFile';
+
+const optionDefinitions = [{ name: 'inject-to' }];
```
LarsDenBakker: do we need to support this? eventually you can just point to the sourcemap file from your `index.html`, temporarily maybe it's easier to handle this in `owc-dev-server` maybe?

packages/import-maps-generate/src/findPathToVersion.js
```diff
@@ -0,0 +1,87 @@
...
+
+  if (rootVersion !== version) {
+    const parents = await findAllParentDeps(depName, deps, packageJson, targetPath);
+
+    for (let i = 0; i < parents.length; i += 1) {
+      const parent = parents[i];
+      const parentPath = require.resolve(parent, {
+        paths: [targetPath],
+      });
+      const subVersionPath = require.resolve(`${depName}/package.json`, {
```
LarsDenBakker: need to use `path.join` here

LarsDenBakker:
Looks good guys, great job !

packages/import-maps-generate/src/generate.js
daKmoR: by the spec it is highly encouraged to inline the map at the top in the html as it is blocking ALL other module requests from even starting - so I would argue such a feature is quite needed

LarsDenBakker: wondering if that should not be an optimization in the build? I just like to keep it in a separate file and keep a single source of truth

LarsDenBakker: since we recommend running import map generation on postinstall, this creates a lot of git diffs/conflicts. Especially when you clone a new project and don't have everything configured correctly yet. This happens a lot already with yarn/package lockfiles.

keeping it in a separate file helps a lot with that

packages/import-maps-generate/README.md
daKmoR: updated readme => example is to use a (temporary) local fork for a hotfix

packages/import-maps-generate/README.md
daKmoR: explained it more in detail hope it's more clear now

packages/import-maps-generate/src/findPathToVersion.js
daKmoR: are you sure? can't resolve handle pathes on its own?

packages/import-maps-generate/src/generate.js
daKmoR: a import map is "locked" to the lock file... so as long as you do not change the lock file the import map will not change... even if you generate it multiple times

so I am not sure what you mean by that?

packages/import-maps-generate/src/findPathToVersion.js
LarsDenBakker: on windows it's a different path delimiter, right?

packages/import-maps-generate/src/generate.js
LarsDenBakker: lockfiles are also notorious for changing when not needed, often due to misconfiguration but this happens a lot when random people clone repositories they want to contribute to. since we inject the import maps into the user's index.html I'm worried people end up unintentionally changing the index.html in their merge request, adding a source of confusion.

We can try it out with this though and see how it works out in practice, but a separate file just seems a lot cleaner to me.

packages/import-maps-generate/src/findPathToVersion.js
daKmoR: jup but it can handle it... at least it work if I run it in windows 10 powershell
but I found another thing... I use path.join somewhere and that generates pathes with `\` in the "values" and they then get escaped so it's `\\` on windows..

long story short it creates "invalid" import maps on windows 😭

daKmoR: so I am not sure should the import map we generate be normalized for linux? I guess? or should we normalize the pathes when we parse the import-map?

let's postpone this discussion for later... let's say linux pathes only for now 😢 

shall I add an info to the readme?

daKmoR: or open an issue?

daKmoR: added it to readme for now

packages/import-maps-generate/src/generate.js
daKmoR: let's keep it in beta until we know how what is the way forward

PR #499, opened 2019-06-15
feat: add postprocessimportmap

thepassle:
TODO:
- ~Change overrides to spread operator~
- ~Implement scopes~
- ~Implement deletes~


daKmoR:
was just reading some parts of the spec again... and I saw this
https://github.com/WICG/import-maps/blob/master/spec.md#merging-import-maps

maybe we should "just" support this + delete? e.g. not "deep" merging of scopes? would certainly make it simpler as it would just 2 spread merges... + logic for deletion...

but again maybe people need more control when creating an import map then when merging 2...

thepassle:
So following [this](https://github.com/WICG/import-maps/blob/master/spec.md#merging-import-maps)

Our config in package.json would look like:
```json
  importmap: {
    deletes: ['a', 'b'],
    overrides: {
      'std:kv-storage': ['std:kv-storage', '/node_modules/kvs-polyfill/index.mjs'],
    },
    merge: {
      'c': 'foo'
    }
  }
```

?

daKmoR:
```js
  "importMap": {
    "deletes": {
        "imports": ["a", "b"],
        // "scopes" and "scopeImports" is also possible
    },

    "overrides": {
       "imports": {
          "std:kv-storage": ["std:kv-storage", "/node_modules/kvs-polyfill/index.mjs"],
          "c": "foo"
       }
       // "scopes" is also possible
    },
  }
```

- I would say `importMap` as other `package.json` keys like `devDependencies` are also camelCased. 
- and "merge" is not needed as it's basically the same as "overrides"? or we should call overrides => mergeReplace or merge or mergeWithOverride, ...?

thepassle:
Why not?

```json
  "importMap": {
    "deletes": ["a", "a/", "b"]
    },

    "overrides": {
      "std:kv-storage": ["std:kv-storage", "/node_modules/kvs-polyfill/index.mjs"],
      "c": "foo",
      "c/": "foo/"
    },
  }
```

daKmoR:
how can you delete or override scopes then?

for example deletion: https://github.com/open-wc/open-wc/pull/499/files#diff-2e109ad19bb252e6d38876cfe42de032R134

or overriding scopes: https://github.com/open-wc/open-wc/pull/499/files#diff-2e109ad19bb252e6d38876cfe42de032R33

thepassle:
@daKmoR PTAL, I think that should be correct now

daKmoR:
@thepassle uhhh that looks really nice 🤗 (sorry for the delay I somehow missed it 🙈 )

just one thing - I think we should call it `importMap` in the `package.json` config (to stay consistent with all the other camelCased properties)

daKmoR:
thxxxx 🤗 

*I will do the `importMap` change in the `feat/importMapCreate` branch*

Issue #501, opened 2019-06-18
Fetch polyfill doesn't work on IE11


Viniter:
Since version 3.0.0 package `whatwg-fetch` contains ES module `export` statements. The package also contains UMD build, which should be included in the legacy build instead. See discussion to this issue: https://github.com/github/fetch/issues/656

Since IE11 doesn't support ES modules, this kind of defeats the purpose of the polyfill. It essentially breaks the legacy build, which according to [docs](https://open-wc.org/building/building-webpack.html#browser-support) should target IE11.

The project's maintainer also recommends to...
> Parse the `package.json` of whatwg-fetch to determine the file referenced by the `main` property, and serve that to browsers.

I'm using the `building-webpack` package and it's `modern-and-legacy-config`, but I have the feeling this might affect the rollup builds as well.

LarsDenBakker:
Thanks we will look into this

Viniter:
Turns out it's a pretty quick fix, I have taken the liberty of opening a PR. It seems to work in my use case.

LarsDenBakker:
Thanks for the PR

PR #502, opened 2019-06-18
fix(building-utils): change fetch polyfill script

Viniter:
Switch fetch polyfill script to use the UMD build (see issue #501)
Fix hashes in tests to match the new polyfill script
Rename `create-index-html` test to match the test suite name

fixes issue #501


LarsDenBakker:
Thanks!

PR #505, opened 2019-06-19
feat: package to parse and resolve import maps

daKmoR:
code is a copy of the reference implementation with some adoption
- parse to string instead of javascript new URL (anticipate support for node paths)
- use es modules, mocha, chai instead of require, jest, jasmine

still todo:
- [x] actually, support paths (for build tools) besides URLs
- [x] documentation
- [x] add thanks to reference implemenation
- [x] clean up history
- [x] final name? `@import-maps/process`? => renamed to `@import-maps/resolve`

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -10,7 +10,7 @@ module.exports = config => {
         {
           pattern: config.grep
             ? config.grep
-            : 'packages/!(webpack-import-meta-loader|create|building-utils|webpack-index-html-plugin|import-maps-generate)/test/*.test.js',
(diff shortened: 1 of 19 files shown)


LarsDenBakker:
What about calling this `@import-maps/resolve`?

packages/import-maps-process/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@import-maps/process",
```
LarsDenBakker: what about calling this `@import-maps/resolve`

daKmoR: will not be the full story but probably waaayy easier to understand - I would say sounds good 👍

daKmoR: renamed to `@import-maps/resolve`

PR #510, opened 2019-06-22
feat(building-rollup): add rollup-index-html-plugin

LarsDenBakker:
- Introduces `rollup-plugin-index-html`, which replaces `modern-web` rollup plugin which was inlined in the `building-rollup` project.

- Both `building-webpack` and `building-rollup` now have a index-html plugin, which shares most of it's code in `@open-wc/building-utils/index-html`.

- `@open-wc/building-utils/index-html` updated to support module and systemjs output

- `@open-wc/building-utils/index-html` updated to handle dynamic import polyfills

- `@open-wc/building-utils/testing-helpers/snapshots` added to enable testing of build output files, our build plugins are now properly tested

- General docs cleanup

This is not a breaking change, as the public interface of `@open-wc/building-rollup` did not change.


packages/rollup-plugin-index-html/TODO.md
```diff
@@ -0,0 +1,2 @@
+- prefetch sync imports in entrypoints
+- update webpack plugin.
```
daKmoR: should that file still be there?

LarsDenBakker: nope, removed it

PR #518, opened 2019-06-26
docs: improve importmap readme

thepassle:


packages/import-maps-generate/README.md
```diff
@@ -77,7 +94,11 @@ you can archive that via an override in your `package.json`
 }
 ```
 
-which will result in the following import map.
+::: warning
+Note that if you apply overrides, you may also need to specify deletions for the generated scope in the importmap.
```
daKmoR: you will need to override the `directory` map right? these are not to be confused with scopes

```
imports: {
   'foo/', '/path/to/foo/', // "directory map" (no clue if there is a name for it 🙈 )
}
scopes: {
   '/path/to/script.js': { // scope to provide a specific import map (which overrides) for a specific path/url
     'foo/': '/path/to/fork/
  }
}
```

packages/import-maps-generate/README.md
```diff
@@ -89,46 +110,39 @@ which will result in the following import map.
...
 
+Overrides may be useful for:
+- Polyfilling
+- Fixing a dependency with a local fork
+- Getting a dependency from a CDN instead
+
 ### Deletes
 
-You can apply deletions to the generated importMap by adding a `importMap` property to your package.json:
+You can apply deletions to the generated importmap by adding a `deletes` property to your package.json:
```
daKmoR: I think it should be called `importMap` everywhere right?

thepassle: Actually should be "[import map](https://github.com/WICG/import-maps#the-import-map)" in running text. Should be `importMap` when referring to the property in package.json.

Fixed it

packages/import-maps-generate/README.md
thepassle: Lets go with directory map for now yeah :)

thepassle: It's just called a `package` mapping: https://github.com/WICG/import-maps#packages-via-trailing-slashes

packages/import-maps-generate/README.md
```diff
@@ -6,17 +6,17 @@ This will allow you to generate a flat [import-map](https://github.com/WICG/impo
 It allows you to fix the "nested" npm problem for front end development.
 
 ::: warning
-Only yarn.lock is supported for now
+Currently only yarn.lock is supported
 :::
 
 ::: warning
-This is still in early beta (with windows pathes not supported yet)
+This is still in early beta (windows paths are not supported yet)
```
bennypowers:
```suggestion
This is still in early beta (Windows paths are not supported yet)
```

packages/import-maps-generate/README.md
```diff
@@ -6,17 +6,17 @@ This will allow you to generate a flat [import-map](https://github.com/WICG/impo
 It allows you to fix the "nested" npm problem for front end development.
 
 ::: warning
-Only yarn.lock is supported for now
+Currently only yarn.lock is supported
```
bennypowers:
```suggestion
Currently, only yarn.lock is supported
```

packages/import-maps-generate/README.md
```diff
@@ -6,17 +6,17 @@ This will allow you to generate a flat [import-map](https://github.com/WICG/impo
...
 :::
 
 ::: warning
-This is still in early beta (with windows pathes not supported yet)
+This is still in early beta (windows paths are not supported yet)
 :::
 
 ## Features
 - generates a flat import-map
```
bennypowers:
```suggestion
- Generates a flat import-map
```

packages/import-maps-generate/README.md
```diff
@@ -6,17 +6,17 @@ This will allow you to generate a flat [import-map](https://github.com/WICG/impo
...
-- prompts when there are not automatically resolveable conflicts
-- supports resolutions, overrides, deletions via the `importMap` property in package.json
+- prompts the user when there are version conflicts that are not automatically resolvable
```
bennypowers:
```suggestion
- Prompts the user when there are version conflicts that are not automatically resolvable
```

packages/import-maps-generate/README.md
```diff
@@ -6,17 +6,17 @@ This will allow you to generate a flat [import-map](https://github.com/WICG/impo
...
+- supports resolutions, overrides and deletions via the `importMap` property in package.json
```
bennypowers:
```suggestion
- Supports resolutions, overrides, and deletions via the `importMap` property in package.json
```

packages/import-maps-generate/README.md
```diff
@@ -6,17 +6,17 @@ This will allow you to generate a flat [import-map](https://github.com/WICG/impo
...
 - supports yarn workspaces (monorepos)
```
bennypowers:
```suggestion
- Supports yarn workspaces (monorepos)
```

packages/import-maps-generate/README.md
```diff
@@ -27,7 +27,7 @@ Install:
 yarn add @import-maps/generate
 ```
 
-Generation of the importMap should happen after you install dependencies, so you can add the script to your postinstall hook in `package.json`:
+Generation of the import map should happen after you install dependencies, so you can add the script to your postinstall hook in `package.json`:
```
bennypowers:
```suggestion
You should generate the import map after you install dependencies by adding the script to your `postinstall` hook in `package.json`:
```

packages/import-maps-generate/README.md
```diff
@@ -43,41 +43,69 @@ npx @import-map/generate
...
+    },
+    "deletes": [],
+    "resolutions": {}
+  }
+}
+```
+
+### Overrides
+
 There may be times where you'll want to apply overrides. For example, if you're using the built-in `std:kv-storage` module and the polyfill for browsers that don't support it, the `kv-storage-polyfill` will be in your yarn.lock, and will be included in your import map like so.
```
bennypowers:
```suggestion
There may be times where you'll want to override certain imports. For example, if you're using the built-in `std:kv-storage` module along with `kv-storage-polyfill` for nonsupporting browsers, then `kv-storage-polyfill`  will be in your lockfile. The generated import map will look like this:
```

packages/import-maps-generate/README.md
```diff
@@ -43,41 +43,69 @@ npx @import-map/generate
...
+        "kv-storage-polyfill": "/node_modules/kv-storage-polyfill/index.js",
+        "kv-storage-polyfill/": "/node_modules/kv-storage-polyfill/"
+      }
+    }
   }
 }
 ```
 
-however what you actually want is to
+however what you actually want is to:
```
bennypowers:
```suggestion
This, however, is not what you really want. Instead, you'd prefer to use the built-in module where supported and fall back to the polyfill if needed.
```

packages/import-maps-generate/README.md
```diff
@@ -43,41 +43,69 @@ npx @import-map/generate
...
 - use the built-in module if supported
 - use the polyfill as a fallback
```
bennypowers:
```suggestion
```

packages/import-maps-generate/README.md
```diff
@@ -43,41 +43,69 @@ npx @import-map/generate
...
+        "kv-storage-polyfill/": "/node_modules/kv-storage-polyfill/"
+      }
+    }
   }
 }
 ```
 
-however what you actually want is to
+however what you actually want is to:
 - use the built-in module if supported
```
bennypowers:
```suggestion
```

packages/import-maps-generate/README.md
```diff
@@ -43,41 +43,69 @@ npx @import-map/generate
...
 - use the polyfill as a fallback
 
-you can archive that via an override in your `package.json`
+you can achieve that via an override in your `package.json`:
```
bennypowers:
```suggestion
You can achieve that via an override in your `package.json`'s `importMap` configuration:
```

packages/import-maps-generate/README.md
```diff
@@ -89,46 +117,56 @@ which will result in the following import map.
...
+      "scopes": {
+        "lit-html/": "/path/to/lit-html/"
+      }
+    }
+  }
+}
+```
+
+
 ### Deletes
```
bennypowers: `overrides` and `resolutions` are nouns, but `deletes` is a verb. Perhaps this key should be `deletions`?

packages/import-maps-generate/README.md
```diff
@@ -89,46 +117,56 @@ which will result in the following import map.
...
 ```
 
-### Overrides
+### Resolutions
 
-If you want to override a dependency in the importMap with a custom value you can add an `overrides` property.
-This can be useful
-- to fix a dependency with a local fork (hopefully temporarily)
-- to get a dependency from a CDN instead
+There may be times where you have conflicting versions. For example; `lit-html@0.14.0` and `lit-html@1.1.0` are both present, and you only want `lit-html@1.1.0`. When this happens, it will prompt the user to ask to solve the resolution. Alternatively, you can specify resolutions in your package.json instead.
```
bennypowers:
```suggestion
There may be times where you have conflicting versions of the same package in your `node_modules`. For example, one package may depend on `lit-html@0.14.0` and another on `lit-html@1.1.0`. In that case, the import map generator will prompt the user to pick a particular version to use canonically.

Alternatively, you can specify your own resolutions in your package.json.
```

packages/import-maps-generate/README.md
thepassle: Why remove it?

packages/import-maps-generate/README.md
thepassle: I agree, though this requires changes in the code as well so I propose to do it in a separate PR

thepassle: Actually nvm, I changed it in this PR

PR #519, opened 2019-06-29
feat(rollup-plugin-index-html): support import maps

daKmoR:
- fixes some missing feature in import-maps/resolve

ToDo:
- [x] Read import map from index.html (currently hard coded in test)
- [x] Remove import map from html output?
- [x] Fix failing test in import-maps/resolve
- [x] Cleanup code (console, lint, ...)
- [x] Cleanup history and make nice commits

### DIFF
--- a/packages/building-utils/index-html/extract-resources.js
+++ b/packages/building-utils/index-html/extract-resources.js
@@ -5,6 +5,8 @@ const { queryAll, predicates, getAttribute, remove } = require('../dom5-fork');
  * @typedef {object} ExtractResult
  * @property {import('parse5').ASTNode} indexHTML the index file, with resources removed
  * @property {string[]} jsModules paths to js modules that were found
+ * @property {string[]} inlineImportMaps content of inline import maps that were found
+ * @property {string[]} importMapPaths paths to import map json files that were found
  */
 
 /**
@@ -21,7 +23,11 @@ function extractResources(htmlString) {
 
   const scripts = queryAll(indexHTML, predicates.hasTagName('script'));
   const moduleScripts = scripts.filter(script => getAttribute(script, 'type') === 'module');
+  const importMapScripts = scripts.filter(script => getAttribute(script, 'type') === 'importmap');
+
   const jsModules = [];
+  const inlineImportMaps = [];
+  const importMapPaths = [];
 
   moduleScripts.forEach(moduleScript => {
     const src = getAttribute(moduleScript, 'src');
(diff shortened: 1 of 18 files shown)


packages/rollup-plugin-index-html/rollup-plugin-index-html.js
```diff
@@ -73,6 +76,36 @@ module.exports = (pluginConfig = {}) => {
       return result.rollupOptions;
     },
 
+    resolveId(source, importer) {
+      const { rootDir } = localPluginConfig;
+      const relativeSource = source.replace(rootDir, '');
+
+      const basePath = importer ? importer.replace(rootDir, `${rootDir}::`) : `${rootDir}::`;
+      if (!importMapCache) {
+        // const mapString = fs.readFileSync(path.join(currentDir, 'import-map.json'), 'utf-8');
```
daKmoR: @LarsDenBakker do you have an idea how to best get the import-map string from the index.html?

also we might need an option to remove the import-map from the generated index.html?

packages/building-utils/index-html/extract-resources.js
```diff
@@ -21,7 +23,11 @@ function extractResources(htmlString) {
 
   const scripts = queryAll(indexHTML, predicates.hasTagName('script'));
   const moduleScripts = scripts.filter(script => getAttribute(script, 'type') === 'module');
+  const importMapScripts = scripts.filter(script => getAttribute(script, 'type') === 'importmap');
+
   const jsModules = [];
+  const importMaps = [];
+  const importMapPaths = [];
```
LarsDenBakker: I think we need to be a bit clearer around the naming conventions. Maybe `inlineImportMaps` and `importMapPaths` ? And similarly we will need to rename `jsModules` to `inlineModules` and `modulePaths` but we can do that later.

What do you think?

LarsDenBakker:
Should we scope this to the rollup plugin?

packages/building-utils/index-html/extract-resources.js
daKmoR: sounds like a good idea 🤗

daKmoR:
what do you mean by "experimental" flat?
if something changes with the import maps then this package will not require changes as that logic is in `@import-maps/resolve`

LarsDenBakker:
If people use the rollup plugin to set up import maps and either the import map spec or our resolve plugin changes, we break their build. Since import maps are still a work in progress, should we enable this by default or require an explicit opt in for now?

packages/building-utils/index-html/extract-resources.js
daKmoR: changed

daKmoR:
puh you mean like... even if there is an import map in the indx.html we should not respect it unless you start rollup with a specific flag?

imho I doubt there will be big changes... also if you don't have an import map in your index.html it's not used... so I'm not sure if adding a flag makes it much better? it will certainly make it another extra step people need to do/learn... so if it can be avoided I would like that

I mean we could ask in the import-maps repo if they still have something "big" in mind they wanna change?

Overall I think the usage of import-maps will probably not explode overnight 🙈 

Verdict:
I would prefer no flag but if preferred I would be ok with it 🤗

LarsDenBakker:
Yes, yeah let's just do it.

PR #525, opened 2019-07-01
feat(rollup-plugin-index-html): enable inline module scripts

bennypowers:
Not sure if this is the best approach (fs.writeFileSync), but here's an ostensibly working implementation of inline importmapped modules for rollup-plugin-index-html for your consideration


LarsDenBakker:
Looks like a good approach, but I want to think about this a bit and look at the use cases.

We need to keep the logic inside `rollup-plugin-index-html` to a minimum, and keep most of the logic in `building-utils`. It needs to work for the webpack plugin too, and es dev server (which is in a sperate PR but also uses these building-utils functions).

packages/rollup-plugin-index-html/src/process-entry-html.js
```diff
@@ -16,15 +17,27 @@ function processEntryHtml(inputConfig) {
...
+    jsModules,
+    inlineScripts,
+    inlineImportMaps,
+    indexHTML: outputIndexHTML,
+  } = extractResources(indexHTMLString);
+  const input = jsModules.map(p => path.join(indexFolder, p));
+  if (inlineScripts.length) {
+    inlineScripts.forEach(script => {
+      const inlineScriptPath = path.join(indexFolder, `inline.${md5(script)}.js`);
+      fs.writeFileSync(inlineScriptPath, script);
```
LarsDenBakker: rollup will already output whatever is in the input, so we don't need to write it ourselves.

packages/rollup-plugin-index-html/src/process-entry-html.js
```diff
@@ -16,15 +17,27 @@ function processEntryHtml(inputConfig) {
...
+  const {
+    jsModules,
+    inlineScripts,
+    inlineImportMaps,
+    indexHTML: outputIndexHTML,
+  } = extractResources(indexHTMLString);
+  const input = jsModules.map(p => path.join(indexFolder, p));
+  if (inlineScripts.length) {
+    inlineScripts.forEach(script => {
+      const inlineScriptPath = path.join(indexFolder, `inline.${md5(script)}.js`);
```
LarsDenBakker: rollup will hash the final script output, you're now hashing the unbuilt content. I think we can just use the script's array index as id. (`inline-${i}.js`)

LarsDenBakker:
Can you add a test fixture of an inline script + non-inline script + legacy build? :)

packages/rollup-plugin-index-html/test/integration.test.js
```diff
@@ -33,6 +33,12 @@ async function testSnapshot(name, configFilePath) {
     });
   }
 
+  if (name.includes('inline-entry')) {
```
LarsDenBakker: `testSnapshots` below already does the same I tink

packages/rollup-plugin-index-html/src/process-entry-html.js
bennypowers: we need to write here because we're *creating* the input

LarsDenBakker: Ouch... I hope there is a cleaner way for thag

packages/rollup-plugin-index-html/src/process-entry-html.js
bennypowers: fine by me

packages/rollup-plugin-index-html/test/integration.test.js
bennypowers: this whole block was committed by mistake

packages/rollup-plugin-index-html/src/process-entry-html.js
bennypowers: unless there's some way to pass script content as a string to input, but I think it only takes a list of pathnames

bennypowers:
@LarsDenBakker in channel we talked about 
- can we assume the user really wants us to extract inline scripts to files? should we instead leave some comment in the dom with a hash of the script and pass that to the plugin, which will replace the inline content with the built version?
- doing more work in build-utils and less in the rollup-plugin, specifically, I imagine you're proposing writing inline modules as files in `@open-wc/building-utils/index-html/extract-resources.js` instead of in rollup?

packages/rollup-plugin-index-html/src/process-entry-html.js
bennypowers: could be we could defer all that work and handle it all in generateBundle or writeBundle, which is probably The Right Way™️, but also more tightly couples the thing to rollup

LarsDenBakker:
After thinking about it, I think we can transform inline module scripts by default and add an option to turn that behavior off. 

I'd prefer not to write any inline modules to files at all, so I guess that takes moving more logic to building-utils out of the equation... I'll check if there is a rollup way for it

LarsDenBakker:
@bennypowers I've pushed changes to avoid writing to temporary file.

I think this is good to go now, can you rebase on master and make this into one commit?

PR #549, opened 2019-07-05
fix: subproperty bug

thepassle:
When applying an override, it adds an extra/unnecessary `.imports` property to the import map

Screen Shot 2019-07-05 at 16 41 40

After applying this fix:

Screen Shot 2019-07-05 at 17 16 53


LarsDenBakker:
Can you scope the commit to the package? And make the commit msg explain which bug it solves?

thepassle:
done

PR #550, opened 2019-07-05
feat(eslint-config): add import extensions linting rule

thepassle:
BREAKING CHANGE: imports other than bare module specifiers now require their extensions to be explicitly set in the import, eg: `import foo from './bar.js';`

### DIFF
--- a/packages/create/src/generators/app/index.js
+++ b/packages/create/src/generators/app/index.js
@@ -2,7 +2,7 @@
 import prompts from 'prompts';
 import commandLineArgs from 'command-line-args';
 import { executeMixinGenerator } from '../../core.js';
-import LintingMixin from '../linting';
+import LintingMixin from '../linting/index.js';
 import TestingMixin from '../testing/index.js';
 import DemoingStorybookMixin from '../demoing-storybook/index.js';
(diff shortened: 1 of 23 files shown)


daKmoR:
I would suggest the following commit message
```
BREAKING CHANGE: imports now require a file extension
Every import that is not a "pure" bare import requires a file extension
\```js
// before
import '../my-el';
import '@open-wc/testing/index-no-side-effects';
import '@open-wc/testing';
// after
import '../my-el.js';
import '@open-wc/testing/index-no-side-effects.js';
import '@open-wc/testing';
\```
```

also in a separate commit you will probably need to fix all those imports now 🙈

daKmoR:
uh can we also update the readme for it 🙏

daKmoR:
imho if we do a breaking change the we should also update to the latest Airbnb and see if we want to "break" more stuff as well...

potentially every change is breaking here 😭

LarsDenBakker:
Is this good to go?

thepassle:
@daKmoR was thinking about any other breaking changes we may want to do while we're at it, he said he recalled wanting to change something else, but couldnt remember what :p

So if we don't have any other breaking changes we want to do, this should be good to go.

PR #553, opened 2019-07-08
feat(es-dev-server): add typescript support

LarsDenBakker:
- Add `--file-extensions` flag output to babel node resolve plugin
- Add docs on how to use typescript
- Improve 'advanced' usage readme

### DIFF
--- a/packages/es-dev-server/package.json
+++ b/packages/es-dev-server/package.json
@@ -34,6 +34,7 @@
     "start:no-watch": "yarn build && node dist/cli.js --app-index demo/import-map/index.html --open --compatibility all",
     "start:babel": "yarn build && node dist/cli.js --app-index demo/babel/index.html --open --watch --http2 --babel",
     "start:no-babel": "yarn build && node dist/cli.js --app-index demo/babel/index.html --open --http2 --watch",
+    "start:typescript": "yarn build && node dist/cli.js --app-index demo/typescript/index.html --file-extensions .ts --node-resolve --open --watch --http2 --babel",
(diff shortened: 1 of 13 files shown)


yarn.lock
```diff
@@ -10630,6 +10637,11 @@ lit-html@^1.0.0:
   resolved "https://registry.yarnpkg.com/lit-html/-/lit-html-1.0.0.tgz#3dc3781a8ca68a9b5c2ff2a61e263662b9b2267b"
   integrity sha512-oeWlpLmBW3gFl7979Wol2LKITpmKTUFNn7PnFbh6YNynF61W74l6x5WhwItAwPRSATpexaX1egNnRzlN4GOtfQ==
 
+lit-html@^1.0.0-rc.2:
```
daKmoR: how come?

packages/es-dev-server/package.json
```diff
@@ -78,6 +80,7 @@
     "babel-preset-typescript": "^7.0.0-alpha.19",
     "chai": "^4.2.0",
     "koa-proxies": "^0.8.1",
+    "lit-element": "2.0.0",
```
daKmoR: any particular reason we have pinned versions here?

LarsDenBakker: Yes a very silly reason that I want to make sure it ends up in the node_modules of es-dev-server, so that they're within the scope of the web server.

I know that's pretty fragile, but it's just for the demo of the dev server.

An alternative can be to always serve from the root of the monorepo.

daKmoR: hihi ok... good enough I would say 😁

Issue #548, opened 2019-07-05
[es-dev-server] inject polyfills on all index html


LarsDenBakker:
Currently we inject polyfills only on the `app-index`, so we require the app index to be set. This is a shame, because it prevents workflows which might have multiple index files.

I think we can use some heuristics to determine if we're serving a index file and inject the polyfills.

Issue #526, opened 2019-07-02
webpack modern-legacy-config, partial html is not valid as template


askbeka:
We have use case where supplied html is not full html, but just a a partial that is used in server to construct full html.

webpack build does not work with this format right now, I haven't checked it for rollup

bennypowers:
Are you passing this partial to the index-html plugin?

askbeka:
@bennypowers updated title, I passing a path to input property for config

LarsDenBakker:
We parse the HTML in order to inject elements into it. In what sense is your HTML partial? Is it still valid HTML?

askbeka:
it is valid html, but woithou html, body or head tags like
```html
<x-component></x-component>
<script type="module" src="./my-component.js"></script>
```

LarsDenBakker:
That's an interesting use case...

Are you expecting polyfills etc. to be injected here, or do you just need a bundled JS output?

askbeka:
Yes, I need polyfills too
Can I do so by providing a template to index html plugin configuration?

LarsDenBakker:
Yep, then providing the template should work out. Let me know if it works for you.

LarsDenBakker:
Closing this as there is no actionable item. Providing the template manually should work here.

PR #555, opened 2019-07-08
feat(es-dev-server): don't require index.html for transform middleware

LarsDenBakker:
closes https://github.com/open-wc/open-wc/issues/548

### DIFF
--- a/packages/es-dev-server/src/middleware/compatibility.js
+++ b/packages/es-dev-server/src/middleware/compatibility.js
@@ -59,8 +59,8 @@ export function createCompatibilityMiddleware(cfg) {
   const { compatibilityMode: compatMode } = cfg;
   /** @type {Map<string, string>} */
   const polyfills = new Map();
-  /** @type {{ body: string, lastModified: string }} */
-  let cachedIndexHTML;
+  /** @type {Map<string, { body: string, lastModified: string }>} */
+  const cachedHTMLs = new Map();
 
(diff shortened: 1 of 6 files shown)


packages/es-dev-server/src/middleware/compatibility.js
```diff
@@ -94,77 +94,98 @@ export function createCompatibilityMiddleware(cfg) {
...
-      // if there were any importmaps, they were extracted. re-add them as an importmap shim
-      indexHTML = injectImportMaps(indexHTML, resources, 'importmap-shim');
-
-      if (!modernOnly) {
-        // if we need to support legacy browsers, also add systemjs-importmap
-        indexHTML = injectImportMaps(indexHTML, resources, 'systemjs-importmap');
-
-        // inject systemjs resolver which appends a query param to trigger es5 compilation
-        indexHTML = indexHTML.replace('</body>', `${systemJsLegacyResolveScript}</body>`);
+      if (!indexHTMLString.includes('<html') && !indexHTMLString.includes('</html>')) {
```
Westbrook: Want to double check the intersection of this test with this issue: https://github.com/open-wc/open-wc/issues/526

LarsDenBakker: This code doesn't touch the webpack plugin. Here we just need to check if it's an `index.html` file being served in order to determine whether we can inject polyfills in there.

PR #556, opened 2019-07-08
feat(semantic-dom-diff): strip some empty attrs

### LINKED ISSUE #554: [semantic-dom-diff] Should strip empty class attributes
`semantic-dom-diff` Should strip empty `class` attrs by default.

The only semantic meaning would be if someone puts a `[class] { /* ... */ }` rule in their css,
… but in my opinion, that person has it coming to them :wink:.

Empty `class` is a common symptom of `classMap`: You add some classMap to your code, but the case only obtains for one of your tests, everything still fails because of an empty `class` attribute which has otherwise no effect

Users can of course add `<div class>` to expected DOM, but this is 
1. a hassle
2. not semantic

Users could also use `{ ignoreAttributes: ['class'] }`, but this would strip *all* class attributes, where there might be meaningful static or dynamic class lists elsewhere in the tested HTML.

In short, `class=""` is *not semantic*, therefore `semantic-dom-diff` should strip it

@daKmoR suggested in slack a new `{ stripEmptyAttributes: ['class'] }` option with `class` in the defaults. Theoretically, all global, non-boolean attrs can go in by default. But in practise that probably only includes `class` and `id`.

bennypowers:
Adds a `stripEmptyAttributes` option, which defaults to `['class', 'id']`,
which when encountering an empty instance of that attr
(i.e. `class=""`), will remove it from the resulting string.

affects: @open-wc/semantic-dom-diff

Fixes #554


LarsDenBakker:
Looks good! Can you update the readme as well?

daKmoR:
can we maybe call it "stripEmptyAttributes"...  
- better be explicit to avoid confusions (especially for options that do not need to be used very often)

Issue #485, opened 2019-06-07
Testing Typscript


hipstersmoothie:
Hey, I love how easy this was all to set up. Converting to typescript was even a breeze for the most part. 

However, I can't seem to get karma working with my typescript. I've tried configuring `karma-typescript` myself but have failed for the last few hours. 

I would really like to see some docs for setting up tests for typescript. The docs are great in most other places, but lack here.

hipstersmoothie:
Findings:

If i use karma-typescript with the following config things kinda start to work

```js
const createDefaultConfig = require('@open-wc/testing-karma/default-config');
const merge = require('webpack-merge');

module.exports = config => {
  config.set(
    merge(createDefaultConfig(config), {
      frameworks: ['karma-typescript'],
      reporters: ['karma-typescript'],
      preprocessors: {
        '**/*.ts': ['esm', 'karma-typescript']
      },
      files: [
        {
          pattern: config.grep ? config.grep : 'src/**/*.ts',
          type: 'module'
        }
      ]
    })
  );

  return config;
};
```

This will error on line 3 of the following test file

```ts
import { html, fixture, expect } from '@open-wc/testing';

import './visually-hidden';

describe('<visually-hidden>', () => {
  it('allows property heading to be overwritten', async () => {
    const el = await fixture(html`
      <visually-hidden></visually-hidden>
    `);

    expect(el.heading).to.equal('different heading');
  });
});
```

I can fix this be adding `.ts` to the import (that breaks my editor plugin though lit-plugin). But it still doesn't build. I am presented with the following error that I don't know how to fix.

```sh
08 06 2019 11:44:14.252:INFO [compiler.karma-typescript]: Compiling project using Typescript 3.5.1
08 06 2019 11:44:18.114:INFO [compiler.karma-typescript]: Compiled 4 files in 3849 ms.
08 06 2019 11:44:18.700:ERROR [karma-server]: Error: Invalid syntax in bundle: 'import' and 'export' may only appear at the top level (23:0)
in /var/folders/kd/4dz51_q566sg7y02l6gwc8zr9mjqwm/T/karma-typescript-bundle-30316F5TV217rQ9xW.js
at line 23, column 0:

... import { TemplateResult } from 'lit-html';
import  ...

Possible fix: configure karma-typescript to compile es6 modules with karma-typescript-es6-transform

    at Validator.validate (/Users/alisowski/Documents/visually-hidden/node_modules/karma-typescript/dist/bundler/validator.js:20:23)
[...]

Issue #484, opened 2019-06-07
Support running karma-esm from a subdir


nicojs:
Hi again 👋 

I've dived into the karma-esm world and found the _actual issue_ why Stryker isn't working with the this plugin.

It seems that it isn't possible to run karma-esm from within in a sub dir. I've created [an example repo here](https://github.com/nicojs/karma-esm-from-a-subdir).

## Repro

```
git clone git@github.com:nicojs/karma-esm-from-a-subdir.git
cd karma-esm-from-a-subdir
npm i
cd foo/bar
npx karma start
```

This error is logged:

```
Uncaught TypeError: Failed to resolve module specifier "@open-wc/testing-helpers/index.js". Relative references must start with either "/", "./", or "../".
  at http://localhost:9876/context.html:0:0
```

## Explanation

My dir structure is:

```
root
├── foo
│   └── bar
│       ├── node_modules (symlink -> ../../node_modules)
│       ├── karma.conf.js
│       ├── src
│       │   └── wc-test.js
│       ├── test
│       │   └── wc-test.test.js
│       └── karma.conf.js
└── node_modules
```

I figured out that a babel plugin is used to transform `import` statements to their true server path here: 

https://github.com/open-wc/open-wc/blob/01445cfbde3d720f61e0d354fa35f32abbb91dcb/packages/karma-esm/src/forked-babel-plugin-bare-import-rewrite.js#L92-L119

This piece of code transforms `import { html, fixture, expect } from '@open-wc/testing'` to `import { html, fixture, expect } from '../../../node_modules/@open-wc/testing'`. This path is correct relative to the test file on the server. The problem is that the browser cannot make sense of it. The file that loads it has this URL: `base/test/wc-test.test.js`, so the url would be `base/test/../../../node_modules/@open-wc/testing` which is not supported. 

## Proposed solution

I've been able to fix this issue by allowing for symlinked node modules. I did this by changing two lines of code:

1. 

https://github.com/open-wc/open-wc/blob/01445cfbde3d720f61e0d354fa35f32abbb91dcb/packages/karma-esm/src/forked-babel-plugin-bare-import-rewrite.js#L79

Change to `result` (remove `require.resolve`)

2. 

https://github.com/open-wc/open-wc/blob/01445cfbde3d720f61e0d354fa35f32abbb91dcb/packages/karma-esm/src/forked-babel-plugin-bare-import-rewrite.js#L71

Change to `preserveSymlinks: true`

[...]

Issue #481, opened 2019-06-07
Use "basePath" to resolve node_modules


nicojs:
Hi there 👋 love this initiative. This issue is about the testing helpers and karma plugin. 

As far as I can see, running karma-esm plugin only supports resolving node_modules relative to the current working directory:

https://github.com/open-wc/open-wc/blob/56dfc680a767871a4d5ca12716dc899820154e77/packages/karma-esm/src/middleware.js#L51

Karma has something called a basePath: https://karma-runner.github.io/4.0/config/configuration-file.html. Would it be possible to resolve node_modules from the basePath instead of cwd?

It would be nice to be able to run karma inside a sub-dir. Without this, we cannot support it with [Stryker, the JavaScript mutation testing framework](https://stryker-mutator.io). See https://github.com/stryker-mutator/stryker/issues/1573

LarsDenBakker:
This is something we want to support. I am working on some other things right now, so I dont know when I can get to this. Open to PRs

nicojs:
Ok, I will try to open a PR. I've got some experience with Karma plugins and middlewares. 

I might choose a different solution with regards to the basePath, since that might have some undesired consequences.

Thanks for the fast reply.

LarsDenBakker:
I have some upcoming changes to our karma setup which will incorporate this

LarsDenBakker:
This was implemented in https://github.com/open-wc/open-wc/pull/562 !

PR #562, opened 2019-07-09
feat(karma-esm): enable testing with modules on all browsers

LarsDenBakker:
Relies on https://github.com/open-wc/open-wc/pull/553 and https://github.com/open-wc/open-wc/pull/563

- Update `karma-esm` to enable testing with modules on all browsers, including IE11. 
- Update testing-karma to use new version of karma-esm
- Remove legacy testing-karma config with karma-webpack
- Add import map option, which shims/polyfills where necessary. we need to discuss the workflow here though, since we can't read it from an index.html

`karma-esm` now spins up `es-dev-server` to handle code compilation, and uses `@open-wc/building-utils/index-html` to set up polyfills etc. just like the rest of our tools.

Unfortunately it was a bit more work than I'd hoped, we have to set up a custom karma test runner html and run a somewhat elaborate loader script to load all the polyfills, libs and run the tests. This is perhaps something to improve on later, but it works and it's not something users notice.

To see it in action, look at `/packages/karma-esm/demo/` and run the test commands from `karma-esm`.

Closes https://github.com/open-wc/open-wc/issues/485
Fixes https://github.com/open-wc/open-wc/issues/484, https://github.com/open-wc/open-wc/issues/481

TODO:

- [x] add ability to exclude some files from babel compile, large libs like mocha, sinon and chai are slow to compile and don't need to be processed for modern browsers
- [x] update documentation
- [x] watch served files
- [x] test on windows


LarsDenBakker:
Should we consider this a breaking change? The API didn't change, but a lot changed under the hood.

It does break for people using a custom webpack configuration, should we consider that part of the public API?

packages/testing-karma/default-config.js
```diff
@@ -1,6 +1,115 @@
...
+
+    // fallback: resolve any karma- plugins
+    'karma-*',
+  ],
+
+  frameworks: ['esm', 'mocha', 'snapshot', 'mocha-snapshot', 'source-map-support'],
+
+  esm: {
+    coverage,
+    compatibility: compatibility ? 'all' : 'none',
```
daKmoR: why only support "all" or "none" while the es-dev-server supports more?

daKmoR:
woohooo - that really looks nice 🎉 all puzzle pieces are falling into place 🤗 

imho it should definitely be a breaking change... especially as we are deleting 2 files which have been probably used as entry points `esm-config.js`, `legacy-config.js`.

and I would say we should also take that chance to export all configs via a `index.js` e.g. end-user api should be
```js
// old
const createDefaultConfig = require('@open-wc/testing-karma/default-config.js');

// new
const createDefaultConfig = require('@open-wc/testing-karma');
```

and move "specific" configs into `src` so it's more clear what people should use as entry points

packages/testing-karma/default-config.js
LarsDenBakker: I was thinking that for tests the speed is a bit less important than for developing. The primary speed difference between compatibility modes is whether or not babel has to be used, which is very quickly the case for code coverage.

But we could also just make it a flag like es-dev-server, that does make it consistent.

Remember that this is just the default config, `karma-esm` supports all the flags and people could turn it on in their config.

LarsDenBakker:
True you could have imported the separate configs. Let's make it a non-default export then:
```
const { createDefaultConfig } = require('@open-wc/testing-karma');
```

that way we have a road forward for other types of configs later on.

LarsDenBakker:
Added another TODO: test on windows

packages/testing-karma/default-config.js
LarsDenBakker: Made it an enum flag instead.

LarsDenBakker:
Tested everything on windows and resolved latest change requests/TODOs. 

Split this change up into two separate breaking commits @daKmoR please check if the commit messages look correct.

This is ready for merging now!

packages/karma-esm/TODO.md
```diff
@@ -0,0 +1,2 @@
+- don't run babel on mocha etc. it's now passed to es dev server
```
daKmoR: we probably want to remove this file at some point right?

LarsDenBakker: Yeah that file is no longer needed! I've removed

daKmoR:
version updates do not look that good 🙈 

```
 - @open-wc/building-rollup: 0.8.4 => 0.8.5
 - @open-wc/building-utils: 2.4.1 => 3.0.0
 - @open-wc/building-webpack: 2.3.1 => 2.3.2
 - @open-wc/chai-dom-equals: 0.12.20 => 1.0.0
 - es-dev-server: 1.6.3 => 2.0.0
 - @open-wc/karma-esm: 1.1.5 => 2.0.0
 - rollup-plugin-index-html: 1.4.2 => 2.0.0
 - @open-wc/semantic-dom-diff: 0.13.0 => 1.0.0
 - @open-wc/testing-helpers: 1.0.11 => 2.0.0
 - @open-wc/testing-karma-bs: 1.1.16 => 2.0.0
 - @open-wc/testing-karma: 2.0.16 => 3.0.0
 - @open-wc/testing: 1.0.13 => 2.0.0
 - @open-wc/webpack-index-html-plugin: 1.2.6 => 2.0.0
```


ok I now made separate commits for the breaking changes and the updates so we get the version updates we need 🤗 

```
 - @open-wc/building-rollup: 0.8.4 => 0.8.5
 - @open-wc/building-utils: 2.4.1 => 2.4.2
 - @open-wc/building-webpack: 2.3.1 => 2.3.2
 - @open-wc/chai-dom-equals: 0.12.20 => 0.12.21
 - es-dev-server: 1.6.3 => 1.6.4
 - @open-wc/karma-esm: 1.1.5 => 2.0.0
 - rollup-plugin-index-html: 1.4.2 => 1.4.3
 - @open-wc/semantic-dom-diff: 0.13.0 => 0.13.1
 - @open-wc/testing-helpers: 1.0.11 => 1.0.12
 - @open-wc/testing-karma-bs: 1.1.16 => 1.1.17
 - @open-wc/testing-karma: 2.0.16 => 3.0.0
 - @open-wc/testing: 1.0.13 => 1.0.14
 - @open-wc/webpack-index-html-plugin: 1.2.6 => 1.2.7
```

Issue #459, opened 2019-05-16
Full support for monorepositories


bashmish:
## Description

I want to use open-wc testing configs in this [monorepository](https://github.com/ing-bank/lion) using lerna and using a single test page context (basically run tests once for the entire repo instead of running it per package).

## Issues

Some challenges I'm facing right now:

1. wallabyjs config does not work out of the box and requires a preprocessor (I introduced it [here](https://github.com/ing-bank/lion/commit/f0046ef1d785b761e53f898b45025b5fafa9c71e)) to prevent double loading of the same ES modules
2. when switching to new testing config which also uses native ES modules same problem happens to karma that happened to wallabyjs

Example of a module which gets loaded twice:
- `import { LionFieldset } from '@lion/fieldset';` (e.g. [here](https://github.com/ing-bank/lion/blob/49f18a42ed70aa62c1290eaf07368d2eeed1b788/packages/radio-group/src/LionRadioGroup.js#L1))
- `import { LionFieldset } from './src/LionFieldset.js';` (e.g. [here](https://github.com/ing-bank/lion/blob/9688007123debe900ba599976d1f48d356609d07/packages/fieldset/lion-fieldset.js#L1))

It causes the double registration of the same custom element.

```
Error: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry
      at node_modules/@lion/fieldset/lion-fieldset.js:41:16HeadlessChrome 74.0.3729 (Mac OS X 10.14.4) ERROR
```

So the first path got rewritten from bare module specifier to a relative path and for the LionFieldset we'll have 2 different paths in the same page context:
- `import { LionFieldset } from '../../../node_modules/@lion/fieldset/index.js'` (which uses `'./src/LionFieldset.js'` unde the hood)
- `import { LionFieldset } from './src/LionFieldset.js';`

You see now that the symlinks which lerna creates are not handled correctly since a browser while loading an ES module has no idea that these different paths point at the same location on the file system. That makes sense since a browser just makes a request to a web server and caches modules by final URLs, not by the paths resolved on the file system level.

## Proposed solution

Would be nice to have a reusable preprocessor exposed in `open-wc` that  automatically fixes double loading of all monorepository packages.

[...]

PR #567, opened 2019-07-10
fix: allows symlinks in node_modules

manolakis:
Having dependencies as symlinks (because of Lerna or npm link for example) in a monorepo causes some fails in unit tests and Storybook because those dependencies are not detected as external dependencies because the absolute path is outside the node_modules and then they are not properly processed (as for example transpiling them).

Enabling the symlinks those problems are fixed and sibling projects are detected as external dependencies.

On the other hand, the function `defineCE` has been refactored because the counter only works when we had just one file in the same location. With symlinks we have several locations and that caused some tag name collisions, so I refactored it to obtain a random number instead a counter and also checks if the random number is available before defining the component.

### DIFF
--- a/packages/building-utils/forked-babel-plugin-bare-import-rewrite.js
+++ b/packages/building-utils/forked-babel-plugin-bare-import-rewrite.js
@@ -64,19 +64,16 @@ function absResolve(importPath, sourceFileName, pluginOptions = {}) {
     return importPath;
   }
 
-  const result = resolve.sync(importPath, {
+  return resolve.sync(importPath, {
     basedir: basedirResolve(importPath, sourceFileName, pluginOptions),
     extensions: pluginOptions.extensions || ['.mjs', '.js', 'json'],
     moduleDirectory: pluginOptions.resolveDirectories || 'node_modules',
-    preserveSymlinks: false,
+    preserveSymlinks: true,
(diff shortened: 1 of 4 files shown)


packages/building-utils/forked-babel-plugin-bare-import-rewrite.js
```diff
@@ -64,19 +64,16 @@ function absResolve(importPath, sourceFileName, pluginOptions = {}) {
...
     moduleDirectory: pluginOptions.resolveDirectories || 'node_modules',
-    preserveSymlinks: false,
+    preserveSymlinks: true,
     packageFilter(packageJson) {
       packageJson.main = packageJson.module || packageJson['jsnext:main'] || packageJson.main;
       return packageJson;
     },
   });
-
-  // use node's resolve to convert symlinks to real paths
```
LarsDenBakker: this was actually added on purpose to fix https://github.com/open-wc/open-wc/issues/459

the intention is to resolve symlinks to their actual paths so that we don't get duplicate imports

manolakis: Ok, I thought that this was added on purpose but I didn't find why. I'm going to close the PR

PR #573, opened 2019-07-11
feat(semantic-dom-diff): add Assert API to Chai plugin

### LINKED ISSUE #537: [semantic-dom-diff] Support Chai assert API
`semantic-dom-diff`'s Chai plugin supports the `expect` and `should` APIs, but not `assert.

I prefer to use the `assert` API over the other two because it reads a bit more clearly.

[`chai-as-promised`](https://github.com/domenic/chai-as-promised) [supports the Assert API](https://github.com/domenic/chai-as-promised/blob/39bfd34988132e443d703c8547fef4e2fbfa7a21/lib/chai-as-promised.js#L314-L355). This may not be the exact approach you want to take, but it is useful to see how others in the community have implemented this. It could be more idiomatic to add methods like `assert.lightDomEqual`, `assert.shadowDomNotEqual`, etc.

I'm happy to implement this if one of the maintainers could validate the use case and provide feedback on the `assert.lightDom.equal` vs `assert.lightDomEqual` question.

AdamVig:
Closes #537.


daKmoR:
thank you for this contribution 🤗

PR #576, opened 2019-07-13
fix missing '/' in legacy/[main].js in win32

stevenao:
In a non-posix environment, the directory separator used is uri incompatible.


LarsDenBakker:
Thank you!

PR #583, opened 2019-07-14
feat(create): move to a single app generator

thepassle:


thepassle:
To do: fix className to tagName in storybook -> index.stories.js

https://polymer.slack.com/archives/CE6D9DN05/p1563341672214000

LarsDenBakker:
Looking good... some thoughts:

Should we add a `(recommend)` marker behind the recommended options. For example rollup over webpack?

Should we add a 'pick all defaults' option?

daKmoR:
if we have more then one option adding a `(recommend)` seems like a good approach... we only have it for building right?

thepassle:
If we add a ‘pick all defaults’ option that basically leads to having a starter-app again though :p

LarsDenBakker:
Hehe yes I understand the irony there. Maybe I was just slow, but I didn't understand that difference in the current setup. Saying that it's the setup with all defaults makes it clear I think, and we can add it for both apps and components.

PR #605, opened 2019-07-19
feat(demoing-storybook): add a11y addon to package and generator

Westbrook:
- update `@open-wc/demoing-storybook` to include the a11y addon
- include these additions in the config of `create/demoing-storybook`


LarsDenBakker:
Looks good to me! @daKmoR any thoughts?

Westbrook:
Tested the updated in the generator using `yarn link` to point to the updated package and things look copacetic to me. Does the system pump out that new version on its own? Or do I need to bump version numbers somewhere?

LarsDenBakker:
Its automatic

Issue #608, opened 2019-07-21
[es-dev-server] Create a createServer separate from startServer


usergenic:
Right now I'd love to wire up a configured `es-dev-server` as a middleware but I can't do that because the big `startServer` function does everything including the `listen()` call.

Would be nice if this was designed to be used as a library as well as dev-server launcher.  Recommend a `createServer` or `setupServer` or `buildServer` function, maybe that gets delegated to by `startServer`.

LarsDenBakker:
Yeah, makes sense I'll add that in.

We can break it up into creating the middlewares, creating the server and starting the server.

LarsDenBakker:
There are some more aspects to consider when programmatically using the dev server, for example when I created `karma-esm` I can't let `es-dev-server` take care of my index.html and have to do all the polyfill work and loading libraries like systemjs myself.

It would be good to gather some use cases, and think of some good generic solutions. Right now the dev server takes quite some assumptions that might not always be valid.

PR #613, opened 2019-07-22
feat(es-dev-server): allow using the server as a library

LarsDenBakker:
Splits up `server.js` into separate parts:
- createConfig
- createMiddlewares
- createServer
- startServer

This will make it much easier to compose the dev server, and use it as a library.

Also adds documentation on how to use the dev server as a library in the advanced section of the docs.

Fixes https://github.com/open-wc/open-wc/issues/608


LarsDenBakker:
Right now programmatically using the server uses a different config object than the command line args config. I want to align these first before merging.

LarsDenBakker:
This is ready for review, please check if the added docs make any sense!

LarsDenBakker:
@bennypowers these titles reflect the functions they're calling, I thought it'd be nice for findability

bennypowers:
> @bennypowers these titles reflect the functions they're calling, I thought it'd be nice for recognition.

I was specifically referring to 

> @bennypowers these titles reflect the functions they're calling, I thought it'd be nice for recognition.

I meant specifically `Command line flags and configuration` => `Command Line Flags and Configuration`. I agree about e.g. `# \`createServer\``

LarsDenBakker:
Ah yes, good catch. I've updated the readme!

packages/es-dev-server/README.md
```diff
@@ -374,6 +380,96 @@ Compatibility mode enables bundle-free development with features such as es modu
...
+`startServer` creates and starts the server, listening on the configured port. It opens the browser if configured and logs a startup message.
+
+Returns the koa app and a node http or http2 server.
+
+```javascript
+import Koa from 'koa';
+import { createConfig, startServer } from 'es-dev-server';
+
+const config = createConfig({ ... });
+const { app, server } = startServer(config, fileWatcher);
```
kevinmpowell: @LarsDenBakker is there a callback that can be run after `startServer` is successful? My use case is for Visual Regression Testing. I need to:

1. Spin up the test server and wait for it to be ready
2. Seed my testing framework with the server's url/port and run the tests against the test server
3. Shut down the test server after my testing framework has finished running.

LarsDenBakker: I'm doubting whether it's not better to use `createServer` in that case. That way, you're in full control when it starts. What do you think?

kevinmpowell: It's not about being in control of when it starts, it's about knowing when the server is ready to receive requests. See Browsersync's `init` command. https://www.browsersync.io/docs/api#api-init

PR #614, opened 2019-07-22
docs: added io gui to component libraries doc

arodic:


daKmoR:
uh another nice library :) thhxx 🤗

Issue #617, opened 2019-07-23
AMP Doesn't work in Shadow DOM


Westbrook:
In the component libraries page, currently, AMP (with the presumption that its project Bento would free AMP from AMP pages) is listed as a more general web component solution. However, if you check out https://stackblitz.com/edit/awc6fe you'll see that AMP doesn't work inside of Shadow DOM and as such, I've removed it from the docs.

PR #618, opened 2019-07-23
docs: remove amp for lack of support in shadow dom

Westbrook:
Fixes: #617 

Remove AMP for lack of shadow dom support from the library listings.


daKmoR:
sad but true...

Issue #616, opened 2019-07-22
[building-*] should have a main which exports all configs


daKmoR:
currently
```js
import createDefaultConfig from '@open-wc/building-rollup/modern-config';
```

should be
```js
import { createConfig, createCompatibilityConfig  } from '@open-wc/building-rollup';
```

so should have a "main" in the package.json for an `index.js` exporting the functions

names are still open for discussion.
=> same for building-webpack

daKmoR:
thx @Westbrook

PR #619, opened 2019-07-23
#616 Update building tools with `main` entry point.

Westbrook:
- adds `main` to rollup
- add `main` to webpack
- **asks** for naming approval

TO DO:
- [ ] add usage in the generator for rollup
- [ ] add usage in the generator for webpack

### DIFF
--- a/packages/building-rollup/README.md
+++ b/packages/building-rollup/README.md
@@ -32,10 +32,11 @@ npm i -D @open-wc/building-rollup rollup rimraf es-dev-server
 
 2. Create a `rollup.config.js` file and pass in your app's `index.html`:
 ```javascript
-import createDefaultConfig from '@open-wc/building-rollup/modern-config';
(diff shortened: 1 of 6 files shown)


daKmoR:
looks good to me (including the names) 👍 

before merging we should update the readmes - generators can/should be done afterwards (as you can't properly test it before it's released)

LarsDenBakker:
Looks great! Should we maybe rename createConfig to createDefaultConfig? Im doubting a bit.

Can you name the main entry point after the package name for consistency?

LarsDenBakker:
We probably need to update the readme as well

Westbrook:
Updated the naming here and expanded the docs.

If we're good on the naming, I've got some generator additions that I can push up as well!

LarsDenBakker:
Looks good to me!

daKmoR:
woop woop 🎉

PR #635, opened 2019-07-25
#616 Finish entry point updates by updating generator

Westbrook:
- update deps and config for rollup
- update config for webpack

I meant to get this into the last PR for this issue. Working on the right communication practices as I get up to speed on the repo.

### DIFF
--- a/packages/create/src/generators/building-webpack/templates/static/webpack.config.js
+++ b/packages/create/src/generators/building-webpack/templates/static/webpack.config.js
@@ -7,6 +7,6 @@ const { createDefaultConfig } = require('@open-wc/building-webpack');
 //   input: path.resolve(__dirname, './index.html'),
 // });
 
-module.exports = createConfig({
+module.exports = createDefaultConfig({
   input: path.resolve(__dirname, './index.html'),
 });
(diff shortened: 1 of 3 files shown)


Westbrook:
This should be good to go if anyone’s ready for a review!

packages/create/src/generators/building-webpack/templates/static/webpack.config.js
```diff
@@ -1,8 +1,11 @@
 const path = require('path');
-const createDefaultConfig = require('@open-wc/building-webpack/modern-config.js');
+const { createDefaultConfig } = require('@open-wc/building-webpack');
```
LarsDenBakker: does webpack need a version bump too?

LarsDenBakker:
We can squash this to 1 commit on merge, right?

Westbrook:
Yes, I _love_ squash.

Webpack is post v1.0.0 so when we make feature releases, it will just take them…yay! 🌮 Happy to make a manual change if you’d prefer to be explicit: https://github.com/open-wc/open-wc/blob/master/packages/create/src/generators/building-webpack/templates/_package.json#L7

Discussed separately doing a v1.0.0 party on the features that we have that are ready for that on Slack, but I’m not 100% which ones are to push that initiative forward.

LarsDenBakker:
Ah, yes you're right. We can keep it as is. Happy to bump everything to 1.x.y, I'm not a huge fan of 0.x.y releases unless we're at a very early stage.

PR #641, opened 2019-07-25
fix(create): set custom element in root of the package as entrypoint

jorenbroekema:
The entrypoint for the custom-element should be in the root of the package. I think I changed it everywhere correctly but WIP to confirm that nothing broke.


daKmoR:
has been move and slightly adjusted to https://github.com/open-wc/open-wc/pull/645

PR #648, opened 2019-07-26
remove weird script tag that doesn't reference anything

SampsonCrowley:


LarsDenBakker:
This script tag is necessary for our docs website. Unfortunately there isn't a nice way to hide it from the github readme :(

SampsonCrowley:
@LarsDenBakker can you elborate on how it's used? maybe there's a better way to handle the docs website build?

LarsDenBakker:
I actually don't know precisely what it does, I think @daKmoR  knows

daKmoR:
it's so people can easily edit the page...

If you go for example to
https://open-wc.org/developing/es-dev-server.html
you see a [edit-this-page](https://github.com/open-wc/open-wc/edit/master/packages/es-dev-server/README.md) link at the bottom.

this is a feature of vuepress. However as we are using symlinks to link the readme files of packages into the `docs` folder (which is the only folder vuepress respects) then your edit link would lead you to this edit page 
https://github.com/open-wc/open-wc/edit/master/docs/developing/es-dev-server.md

e.g. not very convenient => that is why we rewrite the link with this script
yes it's a workaround as ideally a vuepress plugin or a post process thing would fix the link or vuepress would accept "external" (e.g. not in the docs folder) files... but yeah someone would need to make it 🙈

SampsonCrowley:
@daKmoR so does the vuepress plugin have access to modify the repo directly? or how does it actually save the updates? are they submitted as pull requests?

daKmoR:
we just call `vuepress build docs` or simply `npm run vuepress:build` this outputs a folder which can then be statically served.

actually netlify does this under the hood and that is what you see when you visit http://open-wc.org

SampsonCrowley:
I think I'll close this then, and put creating a plugin  on the back-burner for when I get some free time

Issue #646, opened 2019-07-26
[create] move common stuff into a separate common generator


daKmoR:
we have quite a difference for app and web component now - for example, content is way different for package.json or one has an editorconfig the other does not

e.g. they share a lot of "common" stuff like
- package.json
- editorconfig
- gitignore
- gitattribute
- LICENCE
- ...

Issue #644, opened 2019-07-26
[create] move as much config into package.json


daKmoR:
e.g. if possible move all configs like
- commitlint
- eslint
- eclint
- husky
- ...

into the package.json - to not have such a huge scattering of small files in the root of your new project

LarsDenBakker:
Ah, that's a great idea!

PR #657, opened 2019-07-27
feat(create): move configs into package.json, add common-repo generator

daKmoR:
- Remove CommitLint from default linting package
- Uses prettier `--check` instead of own message 
- Prettier and eslint now use .gitignore instead of own ignore file
- Adds a CommonRepoMixin (closes https://github.com/open-wc/open-wc/issues/646)
- Moves lots of configs into package.json (closes https://github.com/open-wc/open-wc/issues/644)

goes from
```
├── foo-bar/
│   ├── .editorconfig
│   ├── .eslintignore
│   ├── .eslintrc.js
│   ├── .gitattributes
│   ├── .gitignore
│   ├── .prettierignore
│   ├── commitlint.config.js
│   ├── foo-bar.js
│   ├── husky.config.js
│   ├── index.js
│   ├── karma.bs.config.js
│   ├── karma.conf.js
│   ├── LICENSE
│   ├── package.json
│   ├── prettier.config.js
│   └── README.md
````

to
```
├── foo-bar/
│   ├── .editorconfig
│   ├── .gitignore
│   ├── foo-bar.js
│   ├── index.js
│   ├── karma.bs.config.js
│   ├── karma.conf.js
│   ├── LICENSE
│   ├── package.json
│   └── README.md
```


bennypowers:
my only concern is that it will make tweaking configs harder for interested developers later on.

daKmoR:
> my only concern is that it will make tweaking configs harder for interested developers later on.

correct, this was/is also my concern 🙈 however, people who need to tweak these kind of things are usually more advanced and should be able to make it happen non the less.

on the other side people who would like to start their first project are really overwhelmed by all these extra files...

so yes it is a balancing act - but I think it's better to by default make it less confusing

Issue #634, opened 2019-07-24
[karma-esm] remove @types/deepmerge


daKmoR:
when doing `yarn install`  on a project just initialized I do get

> warning @open-wc/testing-karma > @open-wc/karma-esm > @types/deepmerge@2.2.0: This is a stub types definition. deepmerge provides its own type definitions, so you do not need this installed.

seems we can just get rid of it?

LarsDenBakker:
Yes seems like it

PR #659, opened 2019-07-28
fix: remove @types/deepmerge dependency

LarsDenBakker:
Fixes https://github.com/open-wc/open-wc/issues/634


daKmoR:
I would recommend to make them fixes so they show up in the changelog (always good for dependency changes)

LarsDenBakker:
I've changed it, but is there some way to make chore end up in the changelog? These aren't really bugfixes.

Issue #656, opened 2019-07-27
[testing-karma] testing leaves process open that requires access to folder


daKmoR:
Steps to reproduce
```bash
$ npm init @open-wc
# select web component -> testing => name: foo-bar
$ cd foo-bar
$ npm i
$ npm run test
# tests runs successfully
$ cd ../
$ rm -fr foo-bar/
rm: cannot remove 'foo-bar/': Permission denied
```

System info:
```
$ node --version
v12.5.0

$ npm --version
6.9.0

$ cat /proc/version:
It is a WSL (Windows Subystem Linux)
Linux version 4.4.0-18922-Microsoft (Microsoft@Microsoft.com) (gcc version 5.4.0 (GCC) ) #1000-Microsoft Fri Jun 14 14:27:00 PST 2019

$ lsb_release -a:
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.3 LTS
Release:        16.04
Codename:       xenial
```

daKmoR:
@LarsDenBakker any ideas?

LarsDenBakker:
It could be that the file watcher for browser reloads stays active or doesnt release the file locks to the os even when the process is killed.

I will try to reproduce this on mac

LarsDenBakker:
I don't get this error on mac, can you try how it goes with https://github.com/open-wc/open-wc/pull/662 ?

SampsonCrowley:
I'd be willing to bet this is a windows-only problem. same way you cant move files or parent folders around that are open in windows, but you can in linux/BSD based OSes (I get bit with this one every single time I shutdown my lixux box and open windows, because I always forget you can't move open files).

the cause in windows would be a still-open file locking it from being modified

daKmoR:
closed via https://github.com/open-wc/open-wc/pull/690

PR #662, opened 2019-07-28
feat(es-dev-server): more robust file watcher cleanup

LarsDenBakker:
Potentially fixes https://github.com/open-wc/open-wc/issues/656 @daKmoR can you test?


daKmoR:
didn't help 😭 

I think it's not filewatcher related as I get the same even when using "just" `yarn test` and not `yarn test:watch`

PR #672, opened 2019-07-30
fix(rollup-config): avoid default indexHTML options override

### LINKED ISSUE #669: [rollupConfig] error trying to import a customPolyfill
Related with the issue #647 I was trying to add a custom polyfill, for example:
```
indexHTMLPlugin: {
    polyfills: {
      customPolyfills: [
        {
          name: 'requestAnimationFrame',
          test: "'requestAnimationFrame' in window",
          path: path.resolve(__dirname, './requestAnimationFrame.js')
        }
      ]
    }
```
This configuration works partially but the problem is that the original `indexHTML` configuration is overrided due to the next lines:
https://github.com/open-wc/open-wc/blob/master/packages/building-rollup/modern-config.js#L44
and
https://github.com/open-wc/open-wc/blob/master/packages/building-rollup/modern-and-legacy-config.js#L53
Those lines override the `polyfills` object because `customPolyfills` is inside `polyfills`. In my case deleting those lines solves my problem and I can create a PR with those changes but I'm not totally sure if you use them for something or is simply a bug.

Tansito:
This PR tries to fix the override of default indexHTML options when you add new configuration through indexHTMLPlugin when you use modern and legacy rollup configuration.

Close #669

### DIFF
--- a/packages/building-rollup/modern-and-legacy-config.js
+++ b/packages/building-rollup/modern-and-legacy-config.js
@@ -37,6 +37,7 @@ function createConfig(_options, legacy) {
     },
     plugins: [
       indexHTML({
+        ...(options.indexHTMLPlugin || {}),
         // tell index-html-plugin that we are creating two builds
         multiBuild: true,
         // tell index-html-plugin whether this is the legacy config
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
Thanks for the contribution!

PR #678, opened 2019-08-01
feat: auto generate types

daKmoR:
an awesome proof of concept by @runem 🎉  https://github.com/runem/js-to-ts-declaration
applied to the testing-helpers

- still a draft to verify if this is is what we want
- are the types correct?
- should this be a real tool? or do we want to wait until tsc supports it directly? e.g. do we want to publish/maintain it? or just for us internally? 
- ... 

e.g. this is a starting point for a discussion 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "site:build": "npm run vuepress:build",
     "site:start": "npm run vuepress:start",
     "build": "lerna run build",
+    "generate:types": "node scripts/generate-declarations.js",
     "test": "karma start --coverage",
     "test:watch": "karma start --auto-watch=true --single-run=false",
     "test:update-snapshots": "karma start --update-snapshots",
--- a/packages/testing-helpers/src/litFixture.d.ts
+++ b/packages/testing-helpers/src/litFixture.d.ts
@@ -1,0 +1,16 @@
+/**
+ * Setups an element synchronously from the provided lit-html template and puts it in the DOM.
+ *
+ * @template {Element} T - Is an element or a node
+ * @param {import('lit-html').TemplateResult} template
+ * @returns {T}
+ */
+export declare function litFixtureSync(template: any): any;
+/**
+ * Setups an element asynchronously from the provided lit-html template and puts it in the DOM.
+ *
+ * @template {Element} T - Is an element or a node
+ * @param {import('lit-html').TemplateResult} template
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "site:build": "npm run vuepress:build",
     "site:start": "npm run vuepress:start",
     "build": "lerna run build",
+    "generate:types": "node scripts/generate-declarations.js",
```
bennypowers: This looks amazing.

I only wish this would be an unscoped package instead of a local script

packages/testing-helpers/src/litFixture.d.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export declare function litFixtureSync(template: any): any;
+/**
+ * Setups an element asynchronously from the provided lit-html template and puts it in the DOM.
+ *
+ * @template {Element} T - Is an element or a node
+ * @param {import('lit-html').TemplateResult} template
+ * @returns {Promise<T>}
+ */
+export declare function litFixture(template: any): {};
```
daKmoR: @runem do you have an idea why this is `any` and not `TemplateResult`?
probably an issue within ts itself I guess?

benjamind: Adding moduleResolution to the ts compiler options seems to have helped a little for me, I now get the TemplateResult as the return type...

```
const program = createProgram({
  rootNames: fileNames,
  options: {
    module: ModuleKind.ESNext,
    moduleResolution: ModuleResolutionKind.NodeJs,
    target: ScriptTarget.ESNext,
    allowJs: true,
    sourceMap: false,
    declaration: true,
    noEmitOnError: false,
  },
```

benjamind: Upon further investigation TemplateStringArray is also not apparently resolving, and so resulting in an any type for the `html` declaration. I suspect the default libs are not being correctly loaded?

runem: @benjamind, @daKmoR I just updated https://github.com/runem/js-to-ts-declaration to overwrite the compiler host in a more gentle way that preserves default libs. This should hopefully solve a couple of problems 🤞

daKmoR:
closed in favour of https://github.com/open-wc/open-wc/pull/697

PR #679, opened 2019-08-01
feat: use tsc to generate typings

daKmoR:
alternative to https://github.com/open-wc/open-wc/pull/678

using a pre-alpha version of `tsc`

- typings look better 👍  
- however jsDoc gets lost 😢 
- exports are invalid (missing `from '../path/to/file.js'`)


packages/testing-helpers/src/fixture.d.ts
```diff
@@ -0,0 +1 @@
+export { fixture, fixtureSync };
```
daKmoR: exports are broken 😭

benjamind:
Very promising though, despite the exports being broken. We're so close!

justinfagnani:
At the rick of derailing a PR conversation, can I ask why the open-wc source just doesn't use TypeScript directly?

daKmoR:
woop woop new version at https://github.com/microsoft/TypeScript/pull/32372#issuecomment-517510909 and it seems to work perfectly now 🎉

daKmoR:
@benjamind could you verify if the types are now fully helpful? 🤗

benjamind:
I can verify that I pulled in the `@openwc/testing-helpers` package from this branch and it compiled against my typescript project without issue! Types appear to be strongly applied, and I managed to remove all of my hand crafted .d.ts files for that package.

I then added `@openwc/semantic-dom-diff` to the compilation and got some usable-ish files. Since its used as a chai plugin i think we need some additional types there to correctly expose its functionality on the Chai Assertion namespace, but I think thats just a thing thats lacking from the jsdoc and not an issue with this tsc generation.

Overall I think this is an excellent approach (second only to native Typescript which would be my preference), but this gives us very usable output for Typescript and I think will mean we can after some more testing remove the allowJs approach from our recommended typing configuration which is a great win for Typescript users.

LarsDenBakker:
@justinfagnani we're not doing that because we want to maintain a low barrier to entry for people to contribute to the project.

justinfagnani:
@LarsDenBakker I understand that concern, but from my experience as a maintainer of a lot of TypeScript-based projects I've found that TS has possibly more often than not made it _easier_ for contributors because of the excellent tooling that helps newcomers navigate the codebase and catch errors early. We've seen comments like "that was easier/more helpful than I expected". I'm personally hesitant to contribute to non-TS projects these days :)

daKmoR:
closed in favor of https://github.com/open-wc/open-wc/pull/697

PR #680, opened 2019-08-02
Allow subclassers to pass custom generator names

jorenbroekema:
Before this change, a subclasser making their own scaffolder would see the "for the same scaffold, use the following command:" displaying `npm init @open-wc` without the ability to change it to the name he wants to register the scaffolder as. Allowing this means they don't have to fork the core stuff for making custom generators / scaffolder outside of open-wc domain :)


daKmoR:
awesome - thx 💪

Issue #582, opened 2019-07-14
[monorepo] add postinstall build


LarsDenBakker:
We're starting to use es modules in our node packages now, we should add a postinstall build command so that they get compiled when updating.

Issue #511, opened 2019-06-22
[ci] run node tests in CI


LarsDenBakker:
currently we run only browsers tests in the CI, we should also run the node tests for tools

daKmoR:
we do run `test:ci` so if you want the ci to run your node tests just add a `"test:ci": "npm run test"`
https://github.com/open-wc/open-wc/blob/master/.circleci/config.yml#L33

LarsDenBakker:
I think that's more for specific tests that should run 'only' in the CI. Ideally the CI just runs `yarn test` so that you can also do this locally, right? currently `yarn test` only runs the browser tests.

daKmoR:
The idea is to have `yarn test:ci` for node and `yarn test` for karma.

you can run both of them locally 👍

LarsDenBakker:
what about `yarn test:node` and `yarn test:browser`, and `yarn:test` which runs both browser and node? that's also how we have linting and format

daKmoR:
sure we can change that 🤗 it will mean to release every package we have (for the package.json change)

let's do it after the release of all the open stuff? rollup plugin, import maps stuff

LarsDenBakker:
Yeah, sure.

PR #691, opened 2019-08-04
fix: cleanup package.json scripts

daKmoR:
- split tests in browser/node and run them both via test (closes https://github.com/open-wc/open-wc/issues/511)
- add post install build (closes https://github.com/open-wc/open-wc/issues/582)
- removes no longer needed `test/index.html` files
- removes individual karma setups in packages (use global setup instead)


LarsDenBakker:
There are conflicts now :(

PR #697, opened 2019-08-05
feat: add type definition files for testing

daKmoR:
this really looks like it should work and is minimally invasive...

however, the types for chai do not work 😱 
and `@types/chai` does not fit either...

probably because we are using `@bundled-es-modules/chai`  
https://www.npmjs.com/package/@bundled-es-modules/chai  
https://github.com/bundled-es-modules/chai  

anyone who can help to type it? I tried to copy-paste the `@types/chai` index.d.ts but that did not help...

without proper types for chai it complains that `expect` to not correct


daKmoR:
@benjamind any clue how to type `@bundled-es-modules/chai`?

daKmoR:
ok "@bundled-es-modules/chai": "^4.2.2" is now typed 🎉 
needed to copy the original chai type files and do some adjustments

so definition files are now generated correctly with properly reexporting `expect, assert, chai, should` 💪 

so one step forward 👍 

So what happens now is that it complains that ` expect(el).dom` is not a thing that is typed... e.g. `dom` is not defined on a chai expect... => so I guess we will need to add those types to the chai namespace...

anyone know by heart how to do it?

daKmoR:
ok this now takes a slightly different approach - we now use chai directly without `@bundled-es-modules/chai`

we theoretically get the typing for it but because of
```js
export const { expect, should, assert } = window.chai;
```

it results in the following types
```ts
export const expect:any;
export const should:any;
export const assert:any;
```

not super nice but at least the typings for all our helpers are there 💪 
=> we can improve the typings of expect/should/assert at some point

LarsDenBakker:
looks good to me

Issue #698, opened 2019-08-06
[testing-karma] running karma with lerna on default config, results in an error


askbeka:
For some reason it does not happen with open-wc config

when I run
```sh
lerna exec karma start -- -c karma.conf.js
```
I get, which is not very helpful
```sh
01 08 2019 12:56:06.661:ERROR [karma-server]: UncaughtException
```

But when run with limited concurrency,  it works
```
lerna exec --concurrency=1 karma start -- -c karma.conf.js
```
I have hunch, that it might be related to ports.
Karma server by default tries to find open ports before starting server
I do not see explicit setting of a port in open-wc config, but not sure why it does not work

LarsDenBakker:
I've seen this error too and I've been trying to debug it. Concurrency is a good lead, I will check into that.

askbeka:
I have checked running 
```sh
lerna exec karma start -- -c karma.conf.js --logLevel=DEBUG
```

I see that resulting configs all have the same port 9876

askbeka:
works with simple karma config.

askbeka:
Narrowed down to esm-middleware.

Tried to define a port myself using `portfinder` on my config side by:
```js
const portFinder = require('portfinder');

module.exports = config => {
  portFinder.getPortPromise()
    .then((port) => {
      config.set(
        merge(createDefaultConfig(config), {
          port,
          files: [
            // ....
```
But I get
```sh
06 08 2019 15:11:40.180:ERROR [karma-server]: Server start failed on port 8000: Error: No provider for "framework:esm"! (Resolving: framework:esm)
06 08 2019 15:11:40.181:ERROR [karma-server]: Server start failed on port 8001: Error: No provider for "framework:esm"! (Resolving: framework:esm)
```

So finding new port works but, something is wrong in setup

askbeka:
So I believe esm-middleware, should use portfinder or something to start the server

LarsDenBakker:
Thanks for the investigation, I'll see if I can do that tonight. 

I didn't look into / consider karma concurrency yet. Did you do anything special to turn on the concurrency mode, or is it always on by default?

daKmoR:
are you sure you wanna use multiple same browsers at the same time? services like browserstack also do not allow this (e.g. you can run chrome and firefox at the same time but not 2x chrome [unless they are different versions])

[...]

Issue #694, opened 2019-08-05
[es-dev-server] No longer throws when port is in use


jorenbroekema:
My port 8080 is always in use on my machine, and es-dev-server would usually throw an error warning about this. Since patch 1.10.1 this no longer happens, and the process just exits without any notice if the default port is already in use.

Reverting to 1.10.0 fixes it for me. 

Additionally would be cool to enhance it where the server doesn't throw, but sends a prompt to the user asking if it's okay to use port `default + 1` (8080 -> 8081) or something like that, similar to how Storybook's cli does it.

LarsDenBakker:
Indeed I ran into the same. Will fix it.

Grabbing another port would be a good feature to add.

Issue #670, opened 2019-07-30
[es-dev-server] Add server start hook on startServer


LarsDenBakker:
Make `startServer` return some hook to know when the server actually started

PR #699, opened 2019-08-06
Feat/random ports

LarsDenBakker:
If no port is explicitly configured, picks a random port in testing and the dev server. This fixes running multiple instances of karma, and is generally good for stability. 

For the dev server it's really useful when you're working with many different projects.

Also fixes logging uncaught exceptions.

fixes https://github.com/open-wc/open-wc/issues/698
fixes https://github.com/open-wc/open-wc/issues/694
fixes https://github.com/open-wc/open-wc/issues/670

### DIFF
--- a/packages/es-dev-server/README.md
+++ b/packages/es-dev-server/README.md
@@ -39,7 +39,7 @@ npx es-dev-server
 ### Server configuration
 | name                 |  type          | description                                                                |
 | -------------------- | -------------- | -------------------------------------------------------------------------- |
-| port                 | number         | The port to use. Default: 8080                                             |
+| port                 | number         | The port to use, uses a random free port if not set.                       |
 | hostname             | string         | The hostname to use. Default: localhost                                    |
 | open                 | boolean/string | Opens the browser on app-index, root dir or a custom path                  |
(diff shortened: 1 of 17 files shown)


packages/es-dev-server/README.md
```diff
@@ -39,7 +39,7 @@ npx es-dev-server
 ### Server configuration
 | name                 |  type          | description                                                                |
 | -------------------- | -------------- | -------------------------------------------------------------------------- |
-| port                 | number         | The port to use. Default: 8080                                             |
+| port                 | number         | The port to use, uses a random free port if not set.                       |
```
daKmoR: so if I restart the server I need to update my url in the browser? 😭 
also I can no longer bookmark the default url?

or does it use 8080 by default and if not free picks another random free one?

LarsDenBakker: It uses 8000 by default, if that one is taken it picks 8001 etc. So if you restart it will stay on the same port.

LarsDenBakker: It uses 8000 by default, if that one is taken it picks 8001 etc. So if you restart it will stay on the same port.

Issue #664, opened 2019-07-29
[es-dev-server] npm link not handled by the server


jarrodek:
I just linked a component from a local folder using `npm link` command. It can see linked folder in proper location however when I run demo page I see 404 error:
```
GET http://localhost:8080/arc-demo-helper/ApiDemoPage.js net::ERR_ABORTED 404 (Not Found)
```
This is the component I am importing in my demo page. The file is there but it is a link. The setup was working perfectly before making the link so it has to be server.

LarsDenBakker:
This is intentional because in a monorepo the same modules are imported with a symlink and with an import to the actual path. If we preserved the symlinks, it would be imported twice.

At the least we should make this behavior configurable, and maybe preserving symlinks should be the default.

LarsDenBakker:
IN https://github.com/open-wc/open-wc/pull/704 you can configure this behavior

cristinecula:
@LarsDenBakker I am having the same issue on v1.12.0:
I have the following directory structure:
* project/
  * app/
  * some-component/

In `app/` I do `npm link some-component` and, when running the server, the imports are now rewritten to `../../../some-component/index.js`, instead of `../../node_modules/some-component/index.js`.

Note that `project` is not a monorepo.

I noticed #704, so I tried it using es-dev-server v1.11.0, but I ran into other issues (some imports were rewritten wrong).

LarsDenBakker:
With https://github.com/open-wc/open-wc/pull/704 you can configure `preserveSymlink` that should fix this. Unfortunately, after releasing this it broke some other things unexpectedly. 

Working on fixing it ASAP.

PR #704, opened 2019-08-07
feat(es-dev-server): improve node-resolve performance

LarsDenBakker:
The current implementation of node-resolve uses babel. This is wasteful, as we're creating a full AST of each js file just for resolving the imports.

In this implementation we use the lexer from [es-module-shims](https://github.com/guybedford/es-module-shims/blob/master/src/lexer.js). This just extracts information about where imports and dynamic imports are, so that we can resolve them and do a simple string replacement. 

There is discussion on separating the lexer https://github.com/guybedford/es-module-shims/issues/35 but that may take a while, so I included it here until we resolve that.

Depending on the project, this can significantly speed up performance. Running the dev server and importing all of `lodash-es` takes 5200ms to load on master, and it takes 1800ms in this branch. Running large bundled libraries like sinon, chai, axe etc. through babel take much longer so that's a huge boost there.

On my machine, running the `open-wc` tests on master takes 5560ms while on this branch it takes 3670ms. This is without coverage turned on, because that still boots up babel. I also updated `karma-esm` to only run babel on files that need coverage. That brings it down to 4400ms with coverage enabled.

Fixes https://github.com/open-wc/open-wc/issues/664


LarsDenBakker:
Would be good if we can test this one some different projects before merging

LarsDenBakker:
We can remove our local copy when https://github.com/guybedford/es-module-lexer/pull/1 is merged

Issue #739, opened 2019-08-15
[es-dev-server] OOM error after a few reloads


justinfagnani:
I'm using pdf.js via https://github.com/bundled-es-modules/pdfjs-dist which has some very large files, but they're not changing between reloads.

LarsDenBakker:
OOM on the server side? Thats not good, I will investigate

LarsDenBakker:
I'm not able to reproduce this, are you importing just `@bundled-es-modules/pdfjs-dist` ? Do you have any other flags enabled?

justinfagnani:
My command is just `es-dev-server --node-resolve`. Here's my imports and a config setup that causes the worker to load:

```ts
import {LitElement, html, css, property, customElement, PropertyValues} from 'lit-element';
import pdfjs from '@bundled-es-modules/pdfjs-dist';
import viewer from "@bundled-es-modules/pdfjs-dist/web/pdf_viewer";

pdfjs.GlobalWorkerOptions.workerSrc =
    '/node_modules/@bundled-es-modules/pdfjs-dist/build/pdf.worker.min.js';
```

You might have to new the viewer to get the worker to load. Fetching it might simulate that. It only took a few reloads to OOM.

The workaround is: `es-dev-server --node-resolve --babel-exclude=**/@bundled-es-modules/pdfjs-dist/build/*`

LarsDenBakker:
I wasn't able to get the worker running, but I'm getting out of memory errors when importing `'@bundled-es-modules/pdfjs-dist/build/pdf.worker.js';` as well.

It runs out of memory during babel compilation when creating source maps, if I turn off source maps in babel it works fine. We might want to turn off source maps, especially for node modules.

With https://github.com/open-wc/open-wc/pull/741 we're no longer using babel for resolving imports, so this that will solve this case as well.

LarsDenBakker:
Source maps are quite nice, especially when people use babel to transform typescript etc. so it'd be a shame to turn them off entirely.

After some playing around with this, I opted for disabling source maps for libraries over 500kb. That's an arbitrary number, I based it around when babel starts deoptimizing the output code styles. We can always adjust this number later.

justinfagnani:
That sounds fine. Files that large are probably compiled and have their own source map, and possibly the only transform in the server is the specifier rewrites, which won't affect the source map much.

PR #741, opened 2019-08-17
feat(es-dev-server): improve node-resolve performance

### LINKED ISSUE #664: [es-dev-server] npm link not handled by the server

LarsDenBakker:
https://github.com/open-wc/open-wc/pull/704 was merged, but I had to revert as I didn't cover some scenario's properly. That's fixed now in this PR.

The current implementation of node-resolve uses babel. This is wasteful, as we're creating a full AST of each js file just for resolving the imports.

In this implementation we use https://github.com/guybedford/es-module-lexer. This just extracts information about where imports and dynamic imports are, so that we can resolve them and do a simple string replacement.

Depending on the project, this can significantly speed up performance. Running the dev server and importing all of lodash-es takes 5200ms to load on master, and it takes 1800ms in this branch. Running large bundled libraries like sinon, chai, axe etc. through babel take much longer so that's a huge boost there.

On my machine, running the open-wc tests on master takes 5560ms while on this branch it takes 3670ms. This is without coverage turned on, because that still boots up babel. I also updated karma-esm to only run babel on files that need coverage. That brings it down to 4400ms with coverage enabled.

Fixes #664
Fixes https://github.com/open-wc/open-wc/issues/739


packages/es-dev-server/test/middleware/compile.test.js
```diff
@@ -141,14 +69,14 @@ describe('babel middleware', () => {
 
       try {
         const responseHtml = await fetch(`${host}src/fake-module.html`);
-        const responseFoo = await fetch(`${host}src/fake-module.foo`);
+        // const responseFoo = await fetch(`${host}src/fake-module.foo`);
```
daKmoR: commented text?

packages/es-dev-server/test/start-server.test.js
```diff
@@ -1,121 +1,121 @@
...
-});
+// /* eslint-disable no-restricted-syntax, no-await-in-loop */
+// import { expect } from 'chai';
+// import fetch from 'node-fetch';
+// import path from 'path';
+// import { startServer, createConfig } from '../src/es-dev-server.js';
+
+// const host = 'http://localhost:8080/';
+
+// describe('server', () => {
```
daKmoR: full test disabled?

LarsDenBakker: Missed this one, thanks!

packages/es-dev-server/test/middleware/compile.test.js
LarsDenBakker: Missed this one, thanks!

Issue #729, opened 2019-08-13
[testing] chai plugins


noelmace:
Using chai plugins with `@open-wc/testing` is tedious and undocumented.

## root cause & context

Chai and most chai plugins aren't exposed as es modules.

This led to the creation of [@bundled-es-modules/chai](https://github.com/bundled-es-modules/chai) and the following issues:
- [ ] https://github.com/domenic/sinon-chai/issues/138
- [ ] https://github.com/nathanboktae/chai-dom/issues/38
- [ ] _other issues I don't know about go here :wink:_

> Thankfully, there still is some Chai Plugins which support ESM:
> - https://github.com/astorije/chai-immutable
> - https://github.com/hellosmithy/chai-rx
> - https://github.com/bustle/chai-graphql

But these projects don't seem to be actively maintained anymore. So, there isn't any advancement on this right now.

## current options

Developers using @open-wc/testing do not have many good options when using chai plugins.

### A. esm fork

:point_right: https://github.com/nathanboktae/chai-dom/issues/38#issuecomment-519068027

### B. define plugins globally

Most chai plugins (chai-dom, sinon-chai, [chai-datetime](https://github.com/mguterl/chai-datetime)) [fall back](https://github.com/nathanboktae/chai-dom/blob/master/chai-dom.js#L1-L11
) to using window.chai.use when AMD, CommonJS, etc... aren't available:

> Some others, like [chai-http](https://github.com/chaijs/chai-http/blob/master/dist/chai-http.js), only provide a "global import" syntax. We can use those the same way as showen bellow, by using a static import statement and then calling `chai.use`.

So, we can:

1. install the "official" (non esm) packages
`npm i -D chai-dom sinon-chai`

2. import chai from `bundled-esm-modules/chai`, add it to window.chai, and then dynamically import the plugins packages:

https://github.com/noelmace/devcards/blob/cf0fe84deb338c47699b2aa289100b9834a68e84/test/utils/chai-plugins.js#L1-L6

> FYI, window.chai doesn't exist in the imported package if you use static import statements instead, but I don't see any workaround for that.

3. add this file to karma files:

https://github.com/noelmace/devcards/blob/cf0fe84deb338c47699b2aa289100b9834a68e84/karma.owc.conf.js#L8-L9

## proposals for resolution

1. I feel like option B is the best workaround right now. We could add it to the doc. What do you think?
[...]

PR #742, opened 2019-08-17
feat(testing): use chai instead of @bundled-es-modules/chai

daKmoR:
- use chai directly
- allows to easily use every chai plugin (no need for es module)
- add chai-dom to default chai plugins
- add example for sinon-chai

- [ ] Update Readme
- [ ] Decide if feat release is good enough

fixes https://github.com/open-wc/open-wc/issues/729


packages/testing/index.js
```diff
@@ -18,4 +19,5 @@ export {
   elementUpdated,
 } from '@open-wc/testing-helpers/index.js';
 
-export { chai, expect, should, assert } from '@bundled-es-modules/chai';
+// @ts-ignore
+export const { expect, should, assert } = window.chai;
```
LarsDenBakker: does this work well with type checking? or is that solved in the other PR?

packages/testing/test/chaiA11yAxe.test.js
```diff
@@ -3,76 +3,62 @@ import { fixture, expect, assert, html } from '../index.js';
 describe('chaiA11yAxe', () => {
   describe('Expect', () => {
     it('passes axe accessible tests', async () => {
-      const el = await fixture(
-        html`
-          <button>some light dom</button>
-        `,
-      );
+      const el = await fixture(html`
```
LarsDenBakker: isn't this the prettier formatting style?

packages/testing/index.js
daKmoR: yeah I wanna rebase the typing branch on this one and see if it "works/helps" before merging this

packages/testing/test/chaiA11yAxe.test.js
daKmoR: prettier allows for both styles 👍  you will need to roughly format it like this so it keeps the auto format - imho 2 lines less is worth the effort

packages/testing/index.js
LarsDenBakker: Ok, LGTM otherwise so just merge whenever.

PR #743, opened 2019-08-17
fix(semantic-dom-diff): fix snapshot error messages

LarsDenBakker:
Chai requires leaking `this` across functions. It's a lot of fun!


daKmoR:
merged via https://github.com/open-wc/open-wc/pull/697/commits/0656c1ce4320bd0eeb1b91bee3f4d5c77623ae6c

Issue #716, opened 2019-08-10
[testing-helpers] allow rendering non-TemplateResult


LarsDenBakker:
`render` from `lit-html` accepts more than just a `TemplateResult` nowadays. It'd be nice to just accept any valid value for rendering, such as dom nodes or an array of templates.

bennypowers:
I'll pick this one

Issue #713, opened 2019-08-10
[testing-helpers] cleanup forced on afterEach


LarsDenBakker:
We force fixture auto cleanup on `afterEach`. I have a case where I'm running a lot of snapshot and a11y tests, and I'd like to run them in a separate test on the same fixture. Setting up a new fixture for each of them takes a lot of time, it's better to set it up once and cleanup once. This isn't possibl, as I can't register the cleanup on `after`. 

Can we think of some way to allow configuring this behavior, while keeping a good developer experience?

PR #717, opened 2019-08-10
docs: improve testing documentation

LarsDenBakker:
Our testing docs had some outdated information, duplicate documentation and overall there wasn't a clear flow. I tried to improve this, while also keeping the readmes sensible for standalone reading outside our website.

- Created a new top-level testing page that acts as a table of content, delegating most of the information to the separate packages instead of repeating the same.
- Link to testing article from @daKmoR for a step by step guide.
- Made the `@open-wc/testing` readme specific to that library only, and no longer our general testing starting place. Moved specific information like browserstack testing to their relevant packages.
- Explained the relationship between `@open-wc/testing` and `@open-wc/testing-helper`
- Created a brief sinon explainer because their website doesn't explain how to install and use it as a module.
- Ran all docs through grammarly
- Made code snippets consistent

PR #744, opened 2019-08-18
Feat/refactor helpers

LarsDenBakker:
- Render test fixtures in a div fixed positioned in the top left corner.
- Make fixture cleanup work on a per-fixture basis.
- Move opinionated auto fixture cleanup from `testing-helpers` to `testing`.
- Allow rendering any valid lit-html render value, not just TemplateResult. lit-html uses `unknown`, so I followed that.

Fixes https://github.com/open-wc/open-wc/issues/716
Fixes https://github.com/open-wc/open-wc/issues/713

This is quite some changes, what do you guys think about this?

I didn't update the docs, I'll do that after discussion and https://github.com/open-wc/open-wc/pull/717


daKmoR:
removing the cleanup from helpers would be a breaking change...

we probably should ask users? as helpers is one of our most used packages and I assume fixture on of the most used functions there... if it now it suddenly require a manual cleanup that might be not so nice...

LarsDenBakker:
I still want to do this

Issue #747, opened 2019-08-19
[webpack-index-html-plugin] Cannot read property 'resolve' of null ( on seccond build with watcher)


sanguino:
Hi, 

I'm new with webpack and open-wc, so please correct me if this is not an issue and I have made a mistake in my config.

When working with `webpack --watch`, the second time the build is done, the `webpack-index-html-plugin` fails:

```bash
｢wdm｣: TypeError: Cannot read property 'resolve' of null
    at compiler.hooks.emit.tapPromise (***/node_modules/@open-wc/webpack-index-html-plugin/webpack-index-html-plugin.js:116:37)
```
I think the problem is that in [line 136](https://github.com/open-wc/open-wc/blob/8d26f71ffe68f9cba58e8d8eb6a095819525969f/packages/webpack-index-html-plugin/webpack-index-html-plugin.js#L136) the `deferredLegacyBuildResult` is set to null. It is created on the constructor, so it will not be created again on the next build.

A possible solution could be to move the creation of the deferredObject to a `watchRun` hook:

``` js
    compiler.hooks.watchRun.tap(PLUGIN_NAME, () => {
      if (this._config.multiBuild && this._config.legacy) {
        deferredLegacyBuildResult = createDeferredPromise();
      }
      this.emitted = false;
    });
```
Also I reset `this.emitted` to emit the `tapPromise` hook on each run too.

I could create a pull request if you think this is a good solution.

LarsDenBakker:
We don't use webpack during development, so we didn't cover these use cases. It sounds like you're in the right track here, if you create a PR we can look at it.

sanguino:
I did all mentioned below and added a reset to `entries` and `entryNamesForVariations` and added test for each case using watcher.

sanguino:
Resolved: https://github.com/open-wc/open-wc/pull/770

PR #754, opened 2019-08-21
Feat/webpack index html plugin watcher

sanguino:
Pull request for issue https://github.com/open-wc/open-wc/issues/747

### DIFF
--- a/packages/webpack-index-html-plugin/test/integration.test.js
+++ b/packages/webpack-index-html-plugin/test/integration.test.js
@@ -4,10 +4,17 @@ const path = require('path');
 const fs = require('fs');
 const os = require('os');
 const { testSnapshots } = require('@open-wc/building-utils/testing-helpers/snapshots');
+const { PLUGIN_NAME } = require('../src/utils');
 
 const fixturesDir = path.join(__dirname, 'fixtures');
 
-function compileAsync(config) {
+function touchAppFile(name) {
+  const appPath = path.join(fixturesDir, name, 'app.js');
+  const now = new Date();
+  fs.utimesSync(appPath, now, now);
+}
+
+function compileAsync(config, addWatcher, name) {
   return new Promise((resolve, reject) => {
     const compiler = webpack({
       ...config,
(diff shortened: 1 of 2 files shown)


daKmoR:
hey there 🤗 
awesome you are taking this on 💪 

I am however not able to reproduce it working 😅 

what I did
```bash
# checkout your branch
yarn
cd packages/building-webpack
yarn demo:webpack-dev:modern
```
- edit packages/building-webpack/demo/js/demo-app.js => reloads and updates
- edit packages/building-webpack/demo/js/index.html => nothing happens? (only after restarting `yarn demo:webpack-dev:modern` the updated content is visible)

that is what we are trying to solve here right?

sanguino:
Hi!

The promise is used when multibuild and legacy are true, so to reproduce it i did:

* Changed `mutibuild` from `production` to `true` in https://github.com/open-wc/open-wc/blob/master/packages/building-webpack/modern-and-legacy-config.js#L129

* run the legacy and moder command:
``` bash
yarn demo:webpack-dev
```
This will works in my brach, but in master will trows something like that:

``` bash
✖ ｢wdm｣: TypeError: Cannot read property 'resolve' of null
    at compiler.hooks.emit.tapPromise (/Users/MPU8L0000044/Developer/OPEN/open-wc/packages/webpack-index-html-plugin/webpack-index-html-plugin.js:114:37)
```

Thanks!

sanguino:
I'm not sure if I explain it very well 😅. Maybe running the integration test in master could be easier.

https://github.com/sanguino/open-wc/blob/feat/webpack-index-html-plugin_watcher/packages/webpack-index-html-plugin/test/integration.test.js

Tell me if this is enough, maybe I could make a test that really changes a file, not only touching it.

Thanks!

sanguino:
@daKmoR, I just read your comment again, I didn't understand you at all, sorry. 

I just trying to solve the error I mention in the last comments, the promise that is used in multi build was destroyed after a build, and fails the second time. My PR resolves this.

But you are right, there is another problem with the plugin in watch mode. The firstime a multibuild is launch, the plugin is executed twice, one for legacy and one for non legacy. But, after touching any file, it is executed only one time, with legacy config, so `emitIndexHTML` is not been called again, so the index html will never be updated.

I will try to make a test to evaluate it better, and try to fix it too.

Thanks!

sanguino:
The pull request https://github.com/open-wc/open-wc/pull/760 make the master branch perfect to test the error I try to solve:

``` bash
yarn demo:webpack-dev
```
If you modify any file, when webpack compile again fails because the promise didn't exists:

``` bash
｢wdm｣: TypeError: Cannot read property 'resolve' of null
    at compiler.hooks.emit.tapPromise (/Users/MPU8L0000044/Developer/OPEN/open-wc/packages/webpack-index-html-plugin/webpack-index-html-plugin.js:113:37)
```
The other problem is related to the `createEntrypoints` function is inside a `entryOption` hook, so it's only executed once. I tried to move it to `beforeCompile` hook, but I found an unexpected behaviour:

1 `yarn demo:webpack-dev`, builds twice (legacy and modern)
2 edit `index.html`, builds only legacy, html is not modified.
3 edit any `.js`, builds twice and html is modified with previous changes.

The reason why when building legacy the html is not created, is because there is a `if (this._config.legacy) -> return` before the `emitIndexHTML` function. But I have no idea why editing an html the watcher only rebuilds the legacy config. I tried to switch legacy and modern configs order in the array, and I could make It works, but I think it's not the way to solve it.

IMHO this behaviour is different than the original issue, so if you want, I could open another issue and try to resolve it in a different branch and different PR to resolve why the watcher only rebuild one config on html edit. 

Thanks!

(Sorry my english, I write this comment very quickly, and my english is not as good as I wish)

LarsDenBakker:
Thanks for your contribution! I did some other fixes to the config, and merged this as part of https://github.com/open-wc/open-wc/pull/770

PR #759, opened 2019-08-22
docs(testing): update README adding karma webpack doc

### LINKED ISSUE #750: [testing] register chai plugins type error
Hi!

Since the last minor version `2.3.0`, where there is an update of using chai plugins, it throws an error  whenever you try to execute the tests with `npm run test` command.

This is the error I mention:

`
TypeError: Cannot read property 'use' of undefined
    at Module../node_modules/@open-wc/testing/register-chai-plugins.js (http://localhost:9876/absoluteC:/Users/AppData/Local/Temp/_karma_webpack_/commons.js?75a6c001bc3dabe4571a9762f775846886fba511:27698:13)
    at __webpack_require__ (http://localhost:9876/absoluteC:/Users/AppData/Local/Temp/_karma_webpack_/runtime.js?c38d51cd4af62e9eacd04bb5dfbfbac27880397b:85:30)
    at Module../node_modules/@open-wc/testing/index.js
`

I do not know if there is any config missing I have to include due to this update, but it is failing when it tries to use those plugins as `window.chai` seems not to be registered well.

This is the line I mention in `register-chai-plugins.js` file: `window.chai.use(chaiDomDiff);`

Thanks in advance!!

acuevas91:
Update README to add karma webpack documentation as it occurs an error if you do not include chai in `karma.conf.js`

Closes #750

### DIFF
--- a/packages/testing/README.md
+++ b/packages/testing/README.md
@@ -79,6 +79,21 @@ describe('Plugin: chai-dom', () => {
 });
 ```
 
+> **__NOTE:__** if you are using `karma-webpack` you have to add `chai/chai.js` to your files in the `karma.conf.js`
+
+```javascript
+files: [
+  { pattern: require.resolve('chai/chai.js' }, 
+  // other files
+]
+```
+
+> and replacing in your test files where you `import { chai } from '@open-wc/testing` by referencing chai on window (if it applies).
+
+```javascript
+window.chai.use(sinonChai);
+```
+
 <script>
   export default {
     mounted() {
--- a/packages/testing/README.md
+++ b/packages/testing/README.md
@@ -88,7 +88,7 @@ files: [
 ]
 ```
 
-> and replacing in your test files where you `import { chai } from '@open-wc/testing` by referencing chai on window (if it applies).
+> If your test files use `import { chai } from '@open-wc/testing`, replace those imports with referencing chai on window.
 
 ```javascript
 window.chai.use(sinonChai);


packages/testing/README.md
```diff
@@ -79,6 +79,21 @@ describe('Plugin: chai-dom', () => {
...
+> **__NOTE:__** if you are using `karma-webpack` you have to add `chai/chai.js` to your files in the `karma.conf.js`
+
+```javascript
+files: [
+  { pattern: require.resolve('chai/chai.js' }, 
+  // other files
+]
+```
+
+> and replacing in your test files where you `import { chai } from '@open-wc/testing` by referencing chai on window (if it applies).
```
bennypowers:
```suggestion
> If your test files use `import { chai } from '@open-wc/testing`, replace those imports with referencing chai on window.
```

daKmoR: question is do we still want to update/fix this? e.g. allow to import chai from `@open-wc/testing`?

bennypowers: I think we should, however, it's a bit awkward to export a global. if the user doesn't set up chai, do we throw an error?

LarsDenBakker: We already import chai in our module, it's just that it doesn't work with webpack out of the box

LarsDenBakker:
@daKmoR @bennypowers shall we add the export? do we want to wait with this Pr on that?

bennypowers:
If we don't decide about the export by the end of the day we should at least merge this in because it's a pressing issue for users.

My take on the export is to add it back in because in the first place it was a breaking change released on a patch version.

daKmoR:
let's add the export 👍

acuevas91:
> let's add the export

What do you mean by adding the export?

daKmoR:
> What do you mean by adding the export?

in here https://github.com/open-wc/open-wc/blob/master/packages/testing/index.js#L23 add something like this
```js
// @ts-ignore
const { expect, should, assert } = window.chai;
// @ts-ignore
const chai = window.chai;
export { expect, should, assert, chai };
```

\+ adding to types https://github.com/open-wc/open-wc/blob/master/packages/testing/index.d.ts#L29 via
```js
export { expect, assert, should, chai };
````
\+ do the same for `index-no-side-effects.js` and `index-no-side-effects.d.ts`

then we can get rid of
```
> If your test files use `import { chai } from '@open-wc/testing`, replace those imports with referencing chai on window.
...
```

sorry for this extra thing 🙈 if too much just let us know 🤗

PR #762, opened 2019-08-24
Feat/dev server improve error handling

LarsDenBakker:
This improves some rough edges around error handling across the board.

- when es-module-lexer fails to parse a file it doesn't give a great error message. with this change, we hand off the file to babel parser to generator a developer-friendly error message. 

- when there is an error compiling (node resolve, compatibility mode, etc.) one of the modules that are early in the request chain it might happen before the message channel endpoint is connected. right now we don't display this error message, with this change we cache error message and send them if they occur within 1 sec

- right now we only log compile errors to the browser and not the terminal because if you're reloading frequently it's hard to know whether an error has been solved or not since your terminal is just full of errors. with this change, we log reloads of your index.html so that errors are visually grouped per reload so that we can turn logging to the terminal back on

- if there is an error in a module the first time it is served, it is not watched for reloads because we only check for 2xx response codes. now we watch any status code and use generic util for finding the served file path

- when you have multiple browser tabs open (for example due to restarting the server a few times) and have watch mode turned on, all the browser tabs will reload on file changes spamming the terminal and slowing down the server. we now only allow one message channel to be active at any given time, which improves the developer experience

- because we only allow one message channel, we have to change the way `karma-esm` listens for file changes. we now pass a custom `fileWatcher` from `karma-esm` to the dev server and listen to that directly. this is a change I was intending to do anyway because it's more robust than relying on a network call


daKmoR:
I like that summary 🤗 
should we put the content of this MRs message in a `DesignDecisions.md` file or so?

LarsDenBakker:
Well maybe not the literal text, since it speaks about before and after. But yeah we could add some docs about some design decisions.

daKmoR:
sure... error message should be more helpful imho

edit it with this
```js
    if (error.code === 'MODULE_NOT_FOUND') {
      const relativeImportFilePath = path.relative(rootDir, sourceFilePath);
      const resolvedImportPath = toBrowserPath(relativeImportFilePath);
      const realtivePathToErrorFile = resolvedImportPath.startsWith('.')
        ? resolvedImportPath
        : `./${resolvedImportPath}`;
      throw new Error(
        `Could not resolve "import ... from '${importPath}';" in "${realtivePathToErrorFile}"`,
      );
    }
```

which imho improves the message
```
Error compiling: Could not resolve "import { ... } from 'asdf';" in "./demo/node-resolve/module-a.js"
// vs old
Error compiling: Could not find module "asdf".
```

maybe for another time... having two "failed" imports after each other only results in one error
```
import 'foo/bar.js'; // will only show error for this
import 'asdf'; // but not for this
```

LarsDenBakker:
We used to add the filename of where the import could not be resolved in the middleware, but it looks like that was removed at some point. Definitely need to add that back in. 

I'm not sure `import { ... } from 'asdf'` is the right way, because you might be doing a default import, an import for side effects or it could even be a dynamic import.

Stopping at the first error is pretty common, one error might be the cause of other errors and you end up printing a lot of messages otherwise. I'm not sure if it really improves the DX.

daKmoR:
> Stopping at the first error is pretty common

sounds good 👍 

When I got the error I really didn't understand what that means `Error compiling: Could not find module "asdf".` 
- Why does it search for this module? 
- Where does it search for this module?
- Is it an es module or module in a sense of a package?

the best would be
```
Error in "await import('asdf');" in "./path/to/file.js:300:20"
                                    ^ Could not resolve es module "asdf"
```

but yeah that is probably tougher to do so, for now, I think we could go with
```
Error compiling: Could not resolve "import { ... } from 'asdf';" in "./demo/node-resolve/module-a.js"
```
- Why does it search for this module? => because it is an import be it static named, static default or a dynamic on (we don't know it here) => so let's show the "most common one" with ... and the user will need to find out for himself
- Where does it search for this module? => path to file is there (actual line + character is sadly not that easy to get)
- Is it an es module or module in a sense of a package? => clear due to import that it is actually a bare module of a package

LarsDenBakker:
Yeah, you're right. It'd be great if we can use babel's code frame error messages here. I'll see what we can do.

Issue #787, opened 2019-09-07
[es-dev-server] add HMR


LarsDenBakker:
Since we're serving unbundled code, replacing modules at runtime on the fly can be implemented pretty easily.

There is a lot of complexity around this, so I'd like to evaluate first whether this adds any value before promoting it in any way. But I suspect we can create a pretty good user experience around editing CSS and HTML.

LarsDenBakker:
Related: https://github.com/systemjs/systemjs/pull/2014 and https://github.com/Polymer/lit-element/issues/673

bennypowers:
should check this out with https://www.npmjs.com/package/custom-elements-hmr-polyfill

LarsDenBakker:
Going to close this one, this isn't a goal in the short or mid term.

justinfagnani:
@LarsDenBakker FYI, I would love to help design and build this into es-dev-server in a way that doesn't require compiling modules away. I have several ideas here, and it would help with @rictic's goal of landing our LitElement HMR support.

LarsDenBakker:
It would be great if we can figure out something that doesn't require compiling modules away. I thought about some different solutions, but it all gets rather complicated. Interested to hear your ideas.

Tiberriver256:
Possibly useful:
https://gist.github.com/treshugart/3b8ee09c16efcbb43453f3f78e59d575

FredKSchott:
Relevant: https://github.com/pikapkg/esm-hmr

It should be relatively plug-and-play for something like es-dev-server to integrate, and open to any feedback if you have time to collaborate / work on it with us

rictic:
@Tiberriver256 I've been working on a roughly similar design, but with the ability for individual webcomponent classes and base classes to define a callback that opts them into handling the details of hot reloading: https://github.com/Polymer/lit-element/pull/802

I've spent a little time with snowpack and esm-hmr today, and I like the design much better than any other HMR system I've seen. For one thing, the code that you write works outside of a bundler (no reliance on an implicit `module` object).

LarsDenBakker:
@rictic what are the plans for publishing this somewhere? 

@FredKSchott I saw it, and it looks relatively straightforward. 

Initially my thoughts for HMR were to create a general reloading system that didn't require any extra work on the side of the user. This isn't possible with ESM since you can't really look into the module loader, so I started working on it in system js (https://github.com/systemjs/systemjs/pull/2014) but then I got sidetracked with other things :)

[...]

PR #788, opened 2019-09-07
[WIP] feat(es-dev-server): add experimental hot module replacement

LarsDenBakker:
Fixes https://github.com/open-wc/open-wc/issues/787

### DIFF
--- a/packages/es-dev-server/demo/hot-reload/hmr.js
+++ b/packages/es-dev-server/demo/hot-reload/hmr.js
@@ -1,0 +1,8 @@
+import { applyPolyfill, ReflowStrategy } from 'custom-elements-hmr-polyfill';
+
+applyPolyfill(
+    // resets the body's innerHTML, thus rerenders all elements
+    // but doesn't call all lifecycle methods in a standard way (less calls)
+    ReflowStrategy.RERENDER_INNER_HTML,
+    /* buffers changes for 500ms */ 500
+);
--- a/packages/es-dev-server/demo/hot-reload/index.html
+++ b/packages/es-dev-server/demo/hot-reload/index.html
@@ -1,0 +1,27 @@
+<html>
+
+<head>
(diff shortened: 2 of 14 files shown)


bennypowers:
is this only blocked on review?

daKmoR:
as far as I can tell `custom-elements-hmr-polyfill` still needs some improvements before it's generally usable... especially a good "ReflowStrategy" as key... 

currently, it's basically a quicker reload... but not what people would expect for HMR... state would need to stay intact - developers are "spoiled" with how nice HMR works in react... so I guess anything less will not fly 🙈 

PS: definitely a good starting point but I think it needs more work before
PPS: imho it should be really close to what react does... e.g. do not register any real classes but proxy classes and if a reregister gets triggered then instead replace those proxy functions and trigger a connectedCallback/rerend - I think this will probably also need some smart way to only resend changed files (e.g. server needs a dependency graph to send all files possibly affected by a file change... )

LarsDenBakker:
It does implement a proxy class actually, it just re renders through inner html by default. If a custom element base class supports it you can trigger a re render on all instances, but I dont think thats the reponsibility of this library.

daKmoR:
you should know which element got updated right? e.g. which tag?

so you should be able to do something like this right?
```js
// beware mega psoido code :p
const toUpdateElements = deepQuerySelectorWhichPiercesShadowDom(tagToUpdate);
toUpdateElements.forEach(el => {
  // replace proxy function stuff
  el.disconnectedCallback();
  el.connectedCallback();
});
```

which should work for every web component be it LitElement or Stencil or ... right?

LarsDenBakker:
That doesnt re run work in the constructor, which is a problem for LitElement and can be a problem for author code.

Because of the proxy we can keep a reference to all connected elements so thats good

LarsDenBakker:
Closing this for now to avoid clutting the review lane, still working on it though.

PR #789, opened 2019-09-07
Fix/edge tests

LarsDenBakker:


packages/es-dev-server/test/middleware/transform-index-html.test.js
```diff
@@ -11,7 +11,7 @@ const host = 'http://localhost:8080/';
 const fixtures = ['simple', 'inline-module', 'import-map'];
 const snapshotsDir = path.join(__dirname, '..', 'snapshots', 'transform-index-html');
 
-describe('transform-index-html middleware', () => {
+describe.only('transform-index-html middleware', () => {
```
daKmoR: missed an `only` here?

LarsDenBakker: Yeah, well I was figuring out why updating the dependencies breaks in the CI

Westbrook: Could be of interest https://www.npmjs.com/package/eslint-plugin-no-only-tests ?

daKmoR: uh very nice we should add that 👍

Issue #782, opened 2019-09-03
[semantic-dom-diff] whitespace treated as semantic for some values of `class`


bennypowers:
Today I received the following test failure:

```
+ expected - actual

-<button
-
-  id="button"
->
+<button id="button">
   <fc-icon
     id="icon"
     size="18"
   >
```

This indicates to me that something is off in the package's handling of whitespace.

the template in question is 
```js
return html`
  <button id="button"
      class="${classMap({ active })}"
      aria-label="${ifDefined(label)}"
      ?disabled="${disabled}"
      ?active="${active}">
    <fc-icon id="icon"
        icon="${ifDefined(icon)}"
        size="${ifDefined(iconSize)}"
    ></fc-icon>
  </button>
`;
```

This appears to be to do with the `class="${classMap({ active })}"` attribute, as removing it removes the errors.

Similarly, setting `class` to `class="${ifDefined(active ? classMap({ active }) : undefined)}"` also negates the issue.

LarsDenBakker:
I'm not sure if the issue with whitespace, we always create our own HTML string representation.

You can check what you're sending to the differ by logging innerHTML or outerHTML whatever you're putting into the chai expect and then compare that with the markdown file on disk.

bennypowers:
@LarsDenBakker Thanks for getting back to me on this.

Here's what I receive when I log the input HTML string alongside the expected HTML string, before asserting with `expect(element).shadowDom.to.equal(html)`


```html
LOG: 'EXPECTED HTML', '
<button id="button">
  <div class="layout-align-space-between-center layout-row">
    <div
      class="layout-align-center-center layout-row mrg-item"
      style="width: 100%"
    >
      <span class="capitalize mrg-item">
        <slot>
        </slot>
      </span>
    </div>
  </div>
</button>
'
LOG: 'INPUT HTML', '<!---->
      <button id="button" class="">
        <div class="layout-row layout-align-space-between-center">

          <!-- loading -->


          <!-- Icon -->


          <!-- Label -->
          <div class="layout-row layout-align-center-center mrg-item" style="width: 100%">
            <span class="mrg-item capitalize">
              <slot></slot>
            </span>
          </div>

          <!-- Right Icon -->

        </div>
      </button>
    <!---->'
```

So it appears that comparing an empty `class=""` input with a non-existent class attr in the expected HTML string causes this problem.

Investigating further with some logs in `getAttributeString`, I noticed this:

```
[...]

PR #799, opened 2019-09-15
fix(semantic-dom-diff): strip empty class attrs when diffing

bennypowers:
affects: @open-wc/semantic-dom-diff

fixes: #782

### DIFF
--- a/packages/semantic-dom-diff/get-diffable-html.js
+++ b/packages/semantic-dom-diff/get-diffable-html.js
@@ -20,6 +20,13 @@ const VOID_ELEMENTS = [
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
What's the bug here exactly? Judging from the test it strips the class attribute correctly already?

bennypowers:
check the diff on the test file

https://github.com/open-wc/open-wc/pull/799/files#diff-b5a9aedaa9e6dbf283193f5f19a32577R320-R321

I added an `id` attribute there.

Now, checkout `get-diffable-html.js` from master and rerun the tests, you'll see it fails with the same symptoms as issue #782

PR #812, opened 2019-09-22
fix: ensure correct module load order on edge

### LINKED ISSUE #781: Tests flaky on Edge
Our tests are flaky on Edge because of es-module-shims doesn't guarantee execution order of modules with side effects on Edge. Because we started exporting chai from a global, we are relying on a side effect.

Let's wait for https://github.com/guybedford/es-module-shims/pull/55 to see if we need to solve this on our end somehow.

tpluscode:
fixes #781


LarsDenBakker:
Thanks for the contribution!

PR #816, opened 2019-09-24
fix for issue 813

petecarapetyan:
If you wanted to try out the generated app before or after this was reviewed, here is the command I used to create the app that found the original bug:

`npm init @open-wc  --type scaffold --scaffoldType app --features building --buildingType rollup --scaffoldFilesFor building --tagName any-name --writeToDisk true --installDependencies yarn`


LarsDenBakker:
What a fascinating bug.

PR #827, opened 2019-09-28
WIP: docs: add codelabs

LarsDenBakker:


codelabs/README.md
```diff
@@ -0,0 +1,8 @@
+To build the codelabs you need to install Go and the codelabs claat tool:
+
+- Install go: https://golang.org/doc/install
+- Setup go in your bash profile: `export PATH=$PATH:$HOME/go/bin`
+- Install claat with `go get github.com/googlecodelabs/tools/claat`
+- Run yarn codelabs:build
+
+See https://medium.com/@zarinlo/publish-technical-tutorials-in-google-codelab-format-b07ef76972cd for a full guide, and https://github.com/googlecodelabs/tools for the CLI tool itself.
```
daKmoR: we wanna use Google code labs? I find it rather complicated for what it does? we should be able to reproduce it with web components right?

also nice would be to only store the "source" in git and generate the output when netlify is building the page (probably not that easy with Google code labs)
otherwise contributing will be really tough 😥

LarsDenBakker: It's just a single command line, so it's not really complicated other than that you need to install Go. 

Building our own solution is just more to maintain again. However we could use the codelab web components directly, and we could just read and parse the markdown in the browser it's not that complex.

The building should happen in netlify for sure, where is the configuration for this stored?

LarsDenBakker:
@daKmoR I added codelabs as a package, I thought that would be easier to find/manage for people. do you think that's ok? or should it be a folder on the top level?

daKmoR:
we do not plan on publishing a codelabs package - as in a tool that can generate codelabs any time soon right?

I would suggest to not have a package but have it as an internal tool for now... once we have a certain amount of codelabs which work fine we can think about publishing the tool 🤗 

about findability, I think that is mostly about how we present it on the website. (maybe have a codelabs on howto add a codelab to open-wc? 😬 )

so a root `codelabs` folder sounds good to me 👍

LarsDenBakker:
@daKmoR what I mean is to put the codelab source code in it's own package, with it's own build scripts etc. this way it's easier to have an overview and we don't clutter the root with extra folders.

Issue #706, opened 2019-08-08
[create] use standard webpack mode notation


LarsDenBakker:
We've been using the wrong webpack mode flag all this time. We need to update the scaffolding to use the correct flag. See https://github.com/open-wc/open-wc/pull/705

LarsDenBakker:
This has been fixed now.

PR #839, opened 2019-10-06
fix(create): use standard webpack mode notation

Aebrathia:
Hello,

I think I fixed all wrong mode notations on create (#706). I made a search in the create folder on "mode" and "webpack".

Please tell me if I need to do something else !

Thanks


LarsDenBakker:
Thanks, much appreciated!

Issue #804, opened 2019-09-17
[prettier] Use it to format/order package.json


daKmoR:
There is a Plugin for it 💪 
https://www.npmjs.com/package/prettier-plugin-package

🤗

CodingMagesty:
Hey, I would like to PR my solution, but I get 403, trying to push my branch to the repo. Any suggestions on what to do?

CodingMagesty:
Hey, please review my PR for this issue. https://github.com/open-wc/open-wc/pull/841

LarsDenBakker:
Closing, since this was implemented.

PR #841, opened 2019-10-09
chore: let prettier handle package.json files

CodingMagesty:
Issue #804


CodingMagesty:
Hey, @daKmoR 
I would like to merge changes according to the issue with package.json, but I am getting errors with CI tests. Could you please review it?
Gleb

daKmoR:
ah you added a dependency but didn't update the `yarn.lock`...

I think running `yarn` and adding the change to your commit should fix it 👍 

PS: we have multiple `packages/*/package.json` files so we might need to lint/fix them all 🤗

daKmoR:
thaaankkk you 🤗 

PS: I did a small fix to apply prettier also to package.json files and then in a separate commit run the formatter 👍

Issue #714, opened 2019-08-10
[testing-karma-bs] add package main


LarsDenBakker:
[testing-karma-bs] does not have a package main, we should add one

PR #844, opened 2019-10-10
Added package main index.js file at testing-karma-bs package

zisiszikos:
#714 

Added an `index.js` file as the main file of the package. It is exporting an object with the `bsSettings` function as a key.


LarsDenBakker:
Thanks for the contribution!

PR #845, opened 2019-10-11
feat(building-rollup): add workbox

thepassle:
Adds workbox to the rollup config.

This should be unobstrusive to people who are just using `build-rollup`, it'll output a service worker, wether they use it (and register it in `index.html`) is up to them. For the generators we could scaffold the sw registration by default. It's also fully configurable if the user provides their own `workbox-config.js`

Some considerations/todos:
- Document a way to add sw registration to index-html-plugin?
- by default does _not_ skipwaiting and clients claim, this becomes very painful very quickly when people start lazyloading/using dynamic imports. We can point people to patterns on how to activate a new service worker (eg: [pwa-helpers](https://github.com/thepassle/pwa-helpers))


packages/building-rollup/modern-and-legacy-config.js
```diff
@@ -107,6 +109,23 @@ function createConfig(_options, legacy) {
 
       // hash
       entrypointHashmanifest(),
+
+      production &&
```
LarsDenBakker: As discussed I think it's clearer to add a boolean option to prevent adding the workbox plugin at all. This way people either go with our defaults, or add their own configuration. The 'extra' config option is just confusing, I think. We should do the same for the other plugins.

LarsDenBakker:
Workbox only applies to SPA's, so we should probably split up the rollup config between a SPA and a 'classic' (JS in, JS out) config.

packages/building-rollup/modern-and-legacy-config.js
thepassle: So something like:

```js
export default createBasicConfig({ 
  input: './index.html',
  useWorkbox: true
  // optional (otherwise use default settings):
  // workboxConfig: require('./workbox-config.js),
});

```

LarsDenBakker: Only booleans for the different plugins (they all default to true):

```js
export default createBasicConfig({ 
  input: './index.html',
  plugins: {
    workbox: false,
    babel: false,
    indexHtml: false,
  }
});
```

And then if you want to customize workbox, you do:

```js
const basicConfig = createBasicConfig({ 
  input: './index.html',
  plugins: {
    workbox: false,
    babel: false,
    indexHtml: false,
  }
});

export default merge(basicConfig, {
  plugins: [
    workbox(/*  own config */)
  ]
});
```

thepassle:
So then we'd export:

```js
module.exports = { 
    createDefaultSPAConfig, // what is currently 'create default config', but with workbox
    createCompatibilitySPAConfig, // what is currently 'create compatability config', but with workbox
    createDefaultConfig // new config, without index.html plugin and without workbox?
};
```

?

LarsDenBakker:
Yeah something like that :) maybe we can think of more inspiring names

packages/building-rollup/modern-and-legacy-config.js
thepassle: Where is  `merge()` from?

LarsDenBakker: Deepmerge, we use it all over the place.

LarsDenBakker: By that I mean it should be easy to find an example somewhere

thepassle: Hahah I got it, no worries 😛

packages/building-rollup/modern-and-legacy-config.js
```diff
@@ -39,74 +46,90 @@ function createConfig(_options, legacy) {
       chunkFileNames: '[name]-[hash].js',
     },
     plugins: [
-      indexHTML({
-        ...(options.indexHTMLPlugin || {}),
```
LarsDenBakker: if we leave this option (and the option below) in but don't mention it anymore in the docs, then I think we can release this without a breaking change. what do you think?

thepassle: It would be nice if we can do it with a breaking change, I've updated the pr, can you take another look when you have time?

packages/building-rollup/README.md
```diff
@@ -187,16 +187,26 @@ If you need to customize the output of your `index.html` you can pass extra opti
 
 ```javascript
 import { createDefaultConfig } from '@open-wc/building-rollup';
+import { indexHTMLPlugin } from '@open-wc/rollup-plugin-index-html';
```
LarsDenBakker: We need to update the sentence at line 186 as well, since we don't recommend passing custom options anymore.

LarsDenBakker:
Looks good to me, just need 1 small doc update

abdonrd:
@thepassle @LarsDenBakker How it works with the `createCompatibilityConfig`?
One `sw.js` to both configs?  One for each one?

abdonrd:
I just create a few issues about this: #873 #874

PR #846, opened 2019-10-11
Refactor building-rollup

thepassle:
We want to do some refactorings to building-rollup, since this is a breaking change, lets merge all the changes in this branch first


thepassle:
Doesnt turn out to need to be breaking after all :-)

PR #848, opened 2019-10-12
fix(semantic-dom-diff): support negate

### LINKED ISSUE #731: [semantic-dom-diff] expect.dom.to.not.equal is not possible
whether `not` is chained or not, the test **always** checks the dom **is** equal.  this is true for `dom`, `lightDom`, and `shadowDom`

rodrigo-garcia-leon:
Adds support for negating assertions.

Fixes #731


LarsDenBakker:
Thanks a lot!

PR #849, opened 2019-10-12
Update README.md

davie-robertson:
Fixed a typo - `target the lates 2 versions` and instead replaced it with `target the 2 most recent versions` as this better grammar


davie-robertson:
Oops, _as this is better grammar_

LarsDenBakker:
Thanks, much appreciated.

Issue #850, opened 2019-10-13
[docs] add haunted code examples


LarsDenBakker:
Similar to https://open-wc.org/developing/code-examples.html, we should have a page which displays code examples for common code patterns using haunted https://github.com/matthewp/haunted

PR #855, opened 2019-10-13
WIP: docs: add haunted code examples

thepassle:
Adds code demos for [haunted](https://github.com/matthewp/haunted)

This is WIP, but I mostly refactored the `basic` demos to Haunted, I think most things are still applicable to usage with Haunted/becoming productive with Haunted

The `intermediate` section will go more into hooks usage

~Not sure what should be in the `hard` section yet~

Also looking for suggestions for what else to add? Probably:
- ~useContext~
- ~useMemo~
- ~useRef~
- ~How to handle styling (just style tags in a template I guess? until theres news about https://github.com/matthewp/haunted/issues/121)~
- ~I want to expand on attributes a little bit more~
- ~Improve documentation in the code demos themself~

Maybe @matthewp can chip in what kind of other demos you'd like to see?

fixes https://github.com/open-wc/open-wc/issues/850


matthewp:
As I mentioned in chat, I've completed the generator app: https://github.com/matthewp/my-app

Style tag is the way to go for now. We could have a `useStyle` or something that uses adoptedStylesheets but currently the `css` tag template function is not part of lit-html, it's in lit-element.

thepassle:
Okay great :-)

So far I've added the following:
Screenshot 2019-10-15 at 20 19 30

I think this can be reviewed/merged, and we can add more demos over time

matthewp:
I get an error when I try to visit https://open-wc-haunted-demos.stackblitz.io/basic, just me?

matthewp:
Looks really great!  In the events example instead of `dispatch(el, 'event-fired', 'some data')` you can also do `dispatch(this, 'event-fired', 'some data')` if using a function function. I sometimes prefer to use `this` because I want to use destructing to get properties from the arguments, so no `el` in scope.  Of course this won't work with arrow functions though.

thepassle:
Yeah stackblitz can be a little flaky sometimes unfortunately 😕 usually a (few) refresh does the trick

Neat, ill add that to events demo 🙂

LarsDenBakker:
Looks good to me, you can merge when you think it's ready.

PR #862, opened 2019-10-15
chore(testing-karma): remove unused dependencies

web-padawan:
Looks like these 2 dependencies are leftovers from using `karma-webpack`

- `istanbul-instrumenter-loader` was used through webpack config,

- `karma-sourcemap-loader` was used through `preprocessors`.

Now when`karma-esm` is used, those packages can be removed.


LarsDenBakker:
Thanks!

PR #864, opened 2019-10-15
chore(demoing-storybook): export withA11y decorator

jorenbroekema:
Ehh yeah so I noticed this was not exported, even though it's already in the dependencies of demoing-storybook. I prefer exporting from one source (@open-wc/demoing-storybook) so would be awesome if you guys can export it so I can import from you guys :)!

Note: Registering the addon is already included in create -> generators -> demoing-storybook -> .storybook/addons.js so that's good ^^

Note 2: Was there a reason that addDecorator wasn't exported? I added it

### DIFF
--- a/packages/demoing-storybook/index.js
+++ b/packages/demoing-storybook/index.js
@@ -1,7 +1,8 @@
 export { html } from 'lit-html';
 
-export { storiesOf, addParameters } from '@storybook/polymer';
+export { storiesOf, addParameters, addDecorator } from '@storybook/polymer';
 export { action } from '@storybook/addon-actions';
+export { withA11y } from '@storybook/addon-a11y';
 export { linkTo } from '@storybook/addon-links';
 export { withNotes } from '@storybook/addon-notes';
 export { document } from 'global';


LarsDenBakker:
Looks good to me, but @daKmoR is probably the best one to decide

Westbrook:
I'm pretty sure the reasoning for this is that both of these methods should really only be used in https://github.com/open-wc/open-wc/blob/master/packages/demoing-storybook/demo/.storybook/config.js

@daKmoR might feel differently, but, because of that, I'm not sure this change is needed in that there should be very few (if any) uses of them in actual stories. Further, moving to this approach would mean that we rely on `withA11y` in the `config.js` via `@storybook/polymer` while users rely on it from `@open-wc/demoing-storybook` which seems a little unbalanced. It would be better to use the same approach in both cases for simplicity.

jorenbroekema:
Personally I don't think we should restrict where you use them. If someone wants to add a decorator in some of his components, but not others, they probably have a valid reason. Maybe they have a folder with experimental features, or something where a certain addon makes no sense.

In any case, I would prefer having 1 dependency for my storybook-related stuff. This is for me one of open-wc's selling points, that if I want to demo my WC, I can import all the stuff I need for it from one package. As it is now, I need to import @open-wc/demoing-storybook, but additionally I need to import some addons and decorators.. And also, many other addons are exported in demoing-storybook, so at the very least it would be inconsistent. 

So I agree with your last point @Westbrook  that if we export it, we should "use our own exports" e.g. whether that's a local path to demoing-storybook/index.js (`import { withA11y } from '../path/to/demoing-storybook/index.js';`) or from the package (`import { withA11y } from '@open-wc/demoing-storybook';`)

Westbrook:
I'm not trying to tell people where to use these methods, but am trying to support them following the best practices outlined by the vendor, https://github.com/storybookjs/storybook/tree/next/addons/a11y#getting-started:

```
// should only be added once
// best place is in config.js
addDecorator(withA11y)
```

If you think there is a good enough reason not to follow that, let's make it nicer! Please update https://github.com/open-wc/open-wc/blob/master/packages/demoing-storybook/demo/.storybook/config.js to support the single-origin idea outlined in your last comment. Sound good?

jorenbroekema:
I find it pretty difficult to figure out what exactly the guidelines or best practices are in terms of addons, when to use the full import `import '@storybook/addon-something`, when to use named imports from the package for story specific stuff, etc. etc. 

The link @Westbrook posted shows docs where there seems to be a clear preference to only import once. In other addons I don't see this reflected, and sometimes the examples specifically import on a per story basis. 

Similarly in open-wc I miss consistency. Not really sure what to do with this anymore to be very honest, maybe it doesn't make sense to strive for this "import everything from open-wc package" ideal. I don't really know storybook well enough to come to a clear conclusion or opinion on this right now.

So I prefer to wait with more work until someone steps in with his/her opinion on the matter :)

daKmoR:
makes sense 🤗 thxxxx 🤗

Issue #853, opened 2019-10-13
Add workbox to building-webpack


thepassle:
We added service worker generation to the building-rollup config, but it'd be nice to have it for building-webpack as well to keep things in sync.

Docs: https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin
Inspiration: https://github.com/open-wc/open-wc/pull/845

bennypowers:
closed by #865

PR #865, opened 2019-10-15
feat(building-webpack): add workbox to webpack conf

thepassle:
https://github.com/open-wc/open-wc/issues/853


daKmoR:
seems like a breaking change right?
we should mark it as such and give an example in the commit message on howto upgrade right?

thepassle:
I dont think its breaking, previous behavior should still be the same, except it outputs a sw now. Options to disaple plugins are optional

daKmoR:
ah options to disable are optional - nice 🤗

PR #877, opened 2019-10-17
Generate the SW once with the createCompatibilityConfig

### LINKED ISSUE #874: [building-rollup] Service Worker is generated twice with createCompatibilityConfig
Screenshot 2019-10-17 at 10 20 25

abdonrd:
Fix #874

I use `!legacy` to just create the SW in the second config, when the legacy files are already created.

### DIFF
--- a/packages/building-rollup/modern-and-legacy-config.js
+++ b/packages/building-rollup/modern-and-legacy-config.js
@@ -119,7 +119,8 @@ function createConfig(_options, legacy) {
       entrypointHashmanifest(),
 
       production &&
-        options.plugins.workbox && !legacy &&
+        options.plugins.workbox && 
+        !legacy &&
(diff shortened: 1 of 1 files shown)


packages/building-rollup/modern-and-legacy-config.js
```diff
@@ -119,7 +119,7 @@ function createConfig(_options, legacy) {
       entrypointHashmanifest(),
 
       production &&
-        options.plugins.workbox &&
+        options.plugins.workbox && !legacy &&
```
thepassle:
```suggestion
        options.plugins.workbox && 
        !legacy &&
```

Because prettier is complaining ^^

thepassle:
@abdonrd Thanks for the fixes!

PR #884, opened 2019-10-18
Use the workbox-config.js as default config

### LINKED ISSUE #879: [testing-karma] Errors with karma 4.4.0
Hi 👋 

I am using @open-wc/testing-karma, and since the update of karma to 4.4.0, a lot of things have stopped to work, and can't start a test suite, crashing after the build.

Also, the main project tests crashes when the karma dependency is resolved to 4.4.0

**Steps to reproduce:**

- Clone master branch
- Delete yarn-lock
- yarn install && npm run test

Regards

abdonrd:
One idea to fix #879


LarsDenBakker:
Seems like a good change to me, what do you think @thepassle ?

thepassle:
Actually, after giving it some more thought, this way would only let users override the `generateSW` mode. So that would mean if a user wants to switch to `injectManifest` mode they'd have to use a different way of overriding (using deepmerge as mentioned in the docs), and we'd now have 2 ways of applying an override which may be confusing?

thepassle:
I think this could be remedied when I upgrade rollup-plugin-workbox to export `generateSW` and `injectManifest` seperately though, then this could work very nicely i think

abdonrd:
Would it make sense for the `rollup-plugin-workbox` to do this job?

abdonrd:
Also we can update the default render message as the official Workbox message:

```javascript
render: ({ swDest, count, size }) => {
  const prettySize = prettyBytes(size);

  console.log(`The service worker file was written to ${swDest}`);
  console.log(`The service worker will precache ${count} URLs, totaling ${prettySize}.`);
},
```

thepassle:
+1 for making the default render the same

Im not sure if we should include this behavior in the plugin itself, does the webpack plugin do this? I would like to make the webpack/rollup plugin as close to eachother as possible. If the webpack plugin does it I think we can add it too.

abdonrd:
I have no idea, I have never used the webpack plugin.

thepassle:
Im gonna work on updating the rollup plugin this week, and ill check if the webpack plugin does it. I’ll keep you posted :-)

thepassle:
I updated the `rollup-plugin-workbox` in `building-rollup` in [this pr](https://github.com/open-wc/open-wc/pull/891/files). Its a fairly small change, so if you could sync up, add the `getWorkboxConfig` to the `generateSW` function in the rollup config, and add some documentation about it, we can merge this :)

abdonrd:
Merged `master` in the PR!

abdonrd:
@thepassle done!

packages/building-rollup/README.md
```diff
@@ -432,7 +432,9 @@ This configuration will by default generate a service worker for you, using [rol
 
 ### Overriding the workbox config
 
-If you want to override the default config with your own workbox configuration, you can disable the default workbox configuration by setting `options.plugins.workbox` to false in the `options` object that you pass to `createDefaultConfig`, and then you can override the plugins
+If you want to override the default config with your own workbox configuration, you can simply create a `workbox-config.js` in the root of the project.
```
thepassle:
```suggestion
If you want to override the default config with your own workbox configuration, you can disable the default workbox configuration by setting `options.plugins.workbox` to false in the `options` object that you pass to `createDefaultConfig`, and then you can override the plugins
```

packages/building-rollup/README.md
```diff
@@ -432,7 +432,9 @@ This configuration will by default generate a service worker for you, using [rol
...
+
+And if you want to further customize, you can disable the default workbox configuration by setting `options.plugins.workbox` to false in the `options` object that you pass to `createDefaultConfig`, and then you can override the plugins
```
thepassle:
```suggestion
Alternatively, you can create a `workbox-config.js` in the root of your project, and it will be automatically be picked up by the configuration if available. Do note that this only works for `generateSW` mode.
```

abdonrd:
Thanks @thepassle!

thepassle:
@abdonrd Thanks for your contribution!

PR #887, opened 2019-10-19
feat(demoing-storybook): allow <style> tag to exists before the custom element

doplumi:
A typical use-case for demoing a custom element consists of showing how it can be styled from the user.

Usually, a <style> tag is added above the content we need to style:

```html
<style> 
  my-el { 
    ... 
  } 
</style>

<my-el></my-el>
```


LarsDenBakker:
@daKmoR I think this is reasonable, right?

daKmoR:
thaaankkk you 🤗

PR #888, opened 2019-10-20
feat(semantic-dom-diff): strip attributes that contains only whitespace

RameezAijaz:
Its an improvement of (#556). It strips attribute that contains only whitespace as a value.


bennypowers:
Nice going, @RameezAijaz

RameezAijaz:
Thanks @bennypowers. I appreciate you guys reviewing and merging PRs quickly.

PR #898, opened 2019-10-23
Fix links in docs/guide readme

cnagadya:
The _Available Recommendations_, _Our Philosophy_ and _FAQs_ links in the docs/guide [readme](https://github.com/open-wc/open-wc/tree/master/docs/guide) are currently returning 404 since relative links[ point to the branches root directory](https://github.blog/2013-01-31-relative-links-in-markup-files/)


cnagadya:
Nice catch @daKmoR. Any ideas on how to make the relative links work in markdown then?

LarsDenBakker:
@daKmoR is this correct now?

LarsDenBakker:
@daKmoR ?

daKmoR:
last I checked it was still not working 🙈

we probably need
1. use the path that works for GitHub
2. add a custom script which modifies the links before netlify deploys the site

PR #902, opened 2019-10-25
Feat/codelab basics 2

LarsDenBakker:
Preview at https://deploy-preview-902--open-wc-org.netlify.com/codelabs/basics/lit-html.html#0


packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+We defined the tag name for our element as `todo-app`, now we need to add this to the HTML of our page:
+
+```html
+<todo-app></todo-app>
+```
+
+If you run this in the browser you should see `lit element connected` logged to the terminal.
+
+<aside class="notice">
+  Because LitElement also does some work in the connectedCallback, we must always call super.connectedCallback. This is a common source of confusion, so make sure to remember!
```
thepassle:
```suggestion
  Because LitElement also does some work in the `connectedCallback`, we must always call `super.connectedCallback()`. This is a common source of confusion, so make sure to remember!
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+Now that we have defined our element, we can start adding a template. lit-element uses lit-html to handle the templating. lit-html works by writing HTML inside of [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) which are strings that can span multiple lines:
+
+```js
+const template = `
+ <h1>Hello world</h1>
+`;
+```
+
+In order to create an actual `lit-html` template, we prefix the template literal with a special HTML tag:
```
thepassle: here we use `lit-html` with backticks, but earlier with 'lit-element' we dont, we should either consistently do it, or not do it

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+```js
+import { html } from 'https://unpkg.com/lit-element?module';
+
+const template = html`
+  <h1>Hello world</h1>
+`;
+```
+
+This is using a feature called [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates), where the `html` tag is a function which returns the prepared template ready for rendering. We won't go into details of how it works exactly, but by using this syntax `lit-html` can very efficiently update the dynamic parts of your template when your element re-renders.
```
thepassle:
```suggestion
This is a native browser feature called [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates), where the `html` tag is a function which returns the prepared template ready for rendering. We won't go into details of how it works exactly, but by using this syntax `lit-html` can very efficiently update the dynamic parts of your template when your element re-renders.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+Most popular IDEs support syntax highlighting of HTML inside template literals, for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
```
thepassle:
```suggestion
Most popular IDEs support syntax highlighting of HTML inside template literals, but for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+```html
+<!DOCTYPE html>
+<html>
+  <head></head>
+
+  <body>
+    <todo-app></todo-app>
+
+    <script type="module">
+      ...code sample goes here...
```
thepassle: is this a to-do?

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+```js
+const author = 'open-wc';
+const homepage = 'https://open-wc.org/';
+const footTemplate = html`
+  <footer>Made with love by <a href="${homepage}">${author}</a></footer>
+`;
+```
+
+When embedding variables like this, lit-html creates static and dynamic parts of your template. When re-rendering the same template you can change the value of these variables, and lit-html will know how to only update the parts that changed. This makes it very efficient!
```
thepassle: Nice examples 😊

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+When this event handler is called, we create a new todo item and add it to the array of todos. Because we're mutating the todos array, `LitElement` is not able to pick up that we changed something. We can let `LitElement` know about this by calling the `requestUpdate` which exists on any element that extends from `LitElement`.
+
+When you click add, you should see the new element appear on the screen.
+
+This allows us to observe the power of lit-html in action. If you inspect the DOM while adding a todo item, you will see that only the new todo item is actually flashing:
+
+TODO: add img
+
+![todo](./assets/foo.png)
```
thepassle: Maybe we can show a gif here?

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+    </script>
+  </body>
+</html>
+```
+
+</detail>
+
+## Reactive property changes
+
+Right now we're triggering an update manually when we make a change. This is fine for some use cases, but it can get pretty cumbersome to do this manually all the time and you will not be able to respond updated triggered by a parent element.
```
thepassle:
```suggestion
Right now we're triggering an update manually when we make a change. This is fine for some use cases, but it can get pretty cumbersome to do this manually all the time and you will not be able to respond to an update triggered by a parent element.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+```js
+static get properties() {
+  return {
+    todos: { type: Array }
+  };
+}
+```
+
+Because we're mutating the todos array using `push`, `LitElement` is not able to pick up the change. `LitElement` change detection is based on immutability, it only detects a change if you create a new array with your changes added. This is a common pattern in front-end to simplify data flow and make change detection easier.
```
thepassle: Maybe it can be good to explain a bit more how the static properties getter in lit-element works. For a long time, for me, it still felt a lot like magic. But then I understood that lit-element just used the static properties getter to make getters and setters, and that _calling_ those setters, sets in motion a chain of events to update. It really reduced the 'magic' feeling for me, and made it 'click' much more.

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+    `;
+  }
+}
+
+customElements.define('todo-list', TodoList);
+```
+
+Ok so the structure of the class should be pretty familiar now, there is a `todos` property and a template to render. The template looks almost the same as the previous, except that now there is an if statement at the start. We need this because unlike before, our todo list isn't the one in charge of the data.
+
+The parent element is still in charge, and we expect it to pass along the todos list to this component. This means that we cannot assume the list to always be there when we do a render. If we don't take care of this somehow, our component will crash because you cannot run a `map` function on `undefined.` Adding early returns in your render function is a simple way to do this, it's easy to see which properties are required for rendering.
```
thepassle: very nice

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+Next, we need to somehow let the parent element know that the user clicked on the checkbox or remove the button. We can do this using DOM events. DOM events are great because the structure and hierarchy of our application are reflected in the DOM, so when a component fires an event only it's parent components can receive them. This creates an automatic scoped system.
```
thepassle:
```suggestion
Next, we need to somehow let the parent element know that the user clicked on the checkbox or remove the button. We can do this using DOM events. DOM events are great because the structure and hierarchy of our application are reflected in the DOM, so when a component fires an event only its parent components can receive them. This creates an automatic scoped system.
```

`it's` stands for `it is`
`its` is possessive

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+    <p>Total finished: ${this._finishedCount}</p>
+    <p>Total unfinished: ${this._unfinishedCount}</p>
+  `;
+}
+```
+
+If you refresh, the UI should remain unchanged. Congratulations, you're already composing elements like a pro :)
+
+In the template, we added we're seeing something different than before. We're passing along our list of todos to the `todo-list` using an attribute with a `.` in front of it:
```
thepassle:
```suggestion
In the template we added, we're seeing something different than before. We're passing along our list of todos to the `todo-list` using an attribute with a `.` in front of it:
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+</html>
+```
+
+</detail>
+
+## Styling
+
+We've covered the basics of templating and managing data with lit-element. The last remaining topic we need to look into is styling. This isn't a codelab on CSS, so we will only look at some of the specifics of working with styling in lit-element.
+
+For styling, lit-element uses shadow dom. If you're not familiar with shadow dom, I recommend following the web component basics codelab.
```
thepassle: Link here?

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+        )}
+      </ol>
+    `;
+  }
+}
+
+customElements.define('todo-list', TodoList);
+```
+
+The styles we define here only apply to our element. This is a guarantee built into the browser because we're using shadow DOM, lit-element does not need to do any extra work for that. This means we can write simple CSS selectors, and not need to worry about creating conflicts with styles defined elsewhere on the page.
```
thepassle: We should be consistent in the use of `shadow dom` or `shadow DOM`

packages/codelabs/src/basics/lit-html.md
```diff
@@ -0,0 +1,1139 @@
...
+      }
+
+      customElements.define('todo-app', TodoApp);
+    </script>
+  </body>
+</html>
+```
+
+</detail>
+
```
thepassle: Maybe we can 'wrap up' by summarizing what we did/learned in this codelab, and maybe where to go next? The end feels a little 'abrupt' to me now.

packages/codelabs/src/basics/lit-html.md
LarsDenBakker: Good one, we should only highlight code snippets

packages/codelabs/src/basics/lit-html.md
LarsDenBakker: note to self: also update this in the basic web component codelab

packages/codelabs/src/basics/lit-html.md
LarsDenBakker: note to self: also update this in the basic web component codelab

packages/codelabs/src/basics/lit-html.md
LarsDenBakker: yep need to gather the final code results of all the steps

packages/codelabs/src/basics/lit-html.md
LarsDenBakker: The idea behind the basics codelab is to know enough to be functional, but not necessarily not how everything works behind the scenes. Do you think this should still fit in there, or this is more something for an advanced codelab?

thepassle: Yeah an aside would be good

PR #907, opened 2019-10-26
feat(demoing-storybook): move to storybook 5.3 with docs mode

daKmoR:
- is a breaking change!
- moves to storybook 5.3 alpha web-components
- there is no longer a need for `default-storybook-webpack-config.js`
- `withClassPropertiesKnobs` got replaced by a decorator `withWebComponentsKnobs`
- withWebComponentsKnobs works based on `custom-elements.json`

See [Preview](https://deploy-preview-907--open-wc-org.netlify.com/demoing/)
With updated
- Readme
- Migration Guide
- Live Storybook demo

PS: anchor hook do not work as it seems that github and vuepress have different methods of creating the link able ids for headings 😭  => maybe an update to vuepress helps? but that is for another time...


blikblum:
Add a withWCKnobs alias to withWebComponentsKnobs ?

daKmoR:
@blikblum sorry I missed that 🙈 

> Add a withWCKnobs alias to withWebComponentsKnobs ?

what would be the reason for it? if it's purely for length then it's maybe something for later... I doubt you will need to add the decorator often (usually only once per file) so having it more descriptive is probably better 🤗  but we will see once it gets used in the wild 💪

blikblum:
Just for the size, but not big deal since editors can autocomplete.

PR #909, opened 2019-10-27
docs: add simple example to event unit test guide

### LINKED ISSUE #894: [docs] improve event listener docs
The testing docs at https://open-wc.org/faq/unit-testing-custom-events.html don't make it clear how to handle events that are fired after a user action, for example on click. We should explain that you need to set up the event listener before doing the action. For example:

```
    it('throws an event with the expected value', async () => {
      const listener = oneEvent(component, 'change');
      component.shadowRoot.querySelector('button').click();
      const { detail } = await listener;
      expect(detail).to.equal('expected value');
    });
```

bennypowers:
fixes #894

### DIFF
--- a/docs/faq/unit-testing-custom-events.md
+++ b/docs/faq/unit-testing-custom-events.md
@@ -1,4 +1,30 @@
-# Testing Event Listeners in lifecycle events
+# Unit Testing Events
+
+Testing events can be tricky; the test needs to set up it's own test-specific event listener before the event fires. In most cases, this can be easily accomplished with the `oneEvent` helper.
+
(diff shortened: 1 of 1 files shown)


bennypowers:
Looks like some console.log statement made it through in a previous PR, and that's gunking up the CI

daKmoR:
I'm wondering why we have a separate FAQ for it?

wouldn't the readme of testing-helpers be a better place?

LarsDenBakker:
I guess at one point we decided to add a FAQ but we never expanded on it. I like the concept of it though

Issue #833, opened 2019-10-03
[testing-helpers] test fixture fails silently with multiple elements


LarsDenBakker:
Test fixtures return the first element child. When multiple elements are present, it leads to unexpected results for the user. It would be better to throw an error in case of multiple elements, so that it's easier to debug.

Saulo8732:
May I work on this?

LarsDenBakker:
That would be great

PR #910, opened 2019-10-27
feat(testing-helpers): allow rendering non-TemplateResult

bennypowers:
`render` from `lit-html` accepts more than just a `TemplateResult` nowadays.
testing-helpers can support string, number, boolean, TemplateResult,
Node, or Arrays or iterables of such.

Note: in the case of Array or Iterable, only the first child will be
taken. All others will be silently ignored. See https://github.com/open-wc/open-wc/issues/833

fixes: #716 
affects: @open-wc/testing-helpers

### DIFF
--- a/packages/testing-helpers/src/litFixture.js
+++ b/packages/testing-helpers/src/litFixture.js
@@ -1,25 +1,53 @@
+import { TemplateResult } from 'lit-html';
 import { fixtureWrapper } from './fixtureWrapper.js';
 import { render } from './lit-html.js';
 import { elementUpdated } from './elementUpdated.js';
+import { NODE_TYPES } from './lib.js';
+
+/**
+ * @typedef {
+     import('lit-html').TemplateResult | import('lit-html').TemplateResult[]
+   | Node | Node[]
+   | string | string[]
+   | number | number[]
+   | boolean | boolean[]
+ } LitHTMLRenderable
+ */
+
+const isUsefulNode = ({ nodeType, textContent }) => {
+  switch (nodeType) {
+    case NODE_TYPES.COMMENT_NODE:
+      return false;
+    case NODE_TYPES.TEXT_NODE:
+      return textContent.trim();
+    default:
+      return true;
+  }
+};
 
 /**
  * Setups an element synchronously from the provided lit-html template and puts it in the DOM.
  *
  * @template {Element} T - Is an element or a node
- * @param {import('lit-html').TemplateResult} template
+ * @param {LitHTMLRenderable} template
  * @returns {T}
  */
 export function litFixtureSync(template) {
   const wrapper = fixtureWrapper();
   render(template, wrapper);
-  return /** @type {T} */ (wrapper.children[0]);
+  if (template instanceof TemplateResult) {
+    return /** @type {T} */ (wrapper.firstElementChild);
+  }
+  const [node] = Array.from(wrapper.childNodes).filter(isUsefulNode);
+
+  return /** @type {T} */ (node);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/testing-helpers/src/litFixture.js
```diff
@@ -1,25 +1,48 @@
...
- * @param {import('lit-html').TemplateResult} template
+ * @param {LitHTMLRenderable} template
  * @returns {T}
  */
 export function litFixtureSync(template) {
   const wrapper = fixtureWrapper();
   render(template, wrapper);
-  return /** @type {T} */ (wrapper.children[0]);
+  const [node] = Array.from(
+    template instanceof TemplateResult ? wrapper.children : wrapper.childNodes,
```
LarsDenBakker: Why not `firstElementChild` ?

bennypowers: could be `TextNode`

LarsDenBakker: Fair enough, I hadn't really considered that to be a use case.

bennypowers: huh...
Screen Shot 2019-10-27 at 10 48 15

bennypowers: Ok I addressed this comment with a small optimization. If it's a `TemplateResult`, we can skip the iteration.

packages/testing-helpers/src/litFixture.js
```diff
@@ -1,25 +1,49 @@
...
  * @returns {T}
  */
 export function litFixtureSync(template) {
   const wrapper = fixtureWrapper();
   render(template, wrapper);
-  return /** @type {T} */ (wrapper.children[0]);
+  if (template instanceof TemplateResult) {
+    return /** @type {T} */ (wrapper.firstElementChild);
+  }
+  const [node] = Array.from(wrapper.childNodes).filter(not(isCommentNode));
```
LarsDenBakker: Should we also filter text nodes with no or only whitespace chars? This can happen when indenting templates

bennypowers: I pushed some code for this but I don't understand the test case well enough to write a test yet.

LarsDenBakker: This creates some empty text nodes:

```fixture(html`

<div></div>

`)```

where I expect to get the div from the fixture

Comment (bennypowers) [reply packages/testing-helpers/src/litFixture.js]: Ok I added a test for this, but control won't get to `wrapper.childNodes` when a `TemplateResult` is passed:

```js
if (template instanceof TemplateResult) {
  return /** @type {T} */ (wrapper.firstElementChild);
}
```
Is this a concern when passing Node or String?

WRT string also, since lit-html's `render` now supports string as well, should we consider removing `stringFixture` altogether?

LarsDenBakker: Why would we make this logic different between `TemplateResult` and others? It's just the means of defining the template, right? 

Indeed we can remove `stringFixture` but that's breaking so for now it can just forward to the regular fixture. stringFixture might be nice if you don't want to use lit-html, but that's already imported in those files I think.

bennypowers: The logic is different because the type is different.

Consider 

```js
render(document.createTextNode('not an element'))
```

In that case, 

```js
wrapper.firstElementChild; // null
```

However, `wrapper.childNodes.length` is 3 - a comment node, our text node, and another comment node.

LarsDenBakker: Right, I see. Thanks for the explanation.

LarsDenBakker:
LGTM, @daKmoR ?

PR #916, opened 2019-10-28
feat(es-dev-server): enable compression

### LINKED ISSUE #723: [es-dev-server] Enable file compression
Can you enable file compression to know the actual transferred data in browser and better performance

bennypowers:
affects: es-dev-server

closes #723


bennypowers:
note this uses the koa-compress default threshold of 1kb, but can be configured by passing a koa-compress options object instead of `true` as the `compress` config item.

```js
createConfig({ compress: { threshold: 1 } });
```

LarsDenBakker:
Do you think turning this on by default will speed things up locally?

bennypowers:
Not necessarily, but as mentioned in the issue, since it's best practice, and cheap to enable, (and for other reasons cited there) we might as well.

daKmoR:
@LarsDenBakker please check and give 👍  or 👎  for using compress by default

LarsDenBakker:
Lets do compression by default

PR #919, opened 2019-10-29
feat(create): dynamic READMEs

### LINKED ISSUE #917: [create] readme improvements
In the readme generated by `npm init @open-wc` we should only include the commands which we actually scaffolded, I think we can take a similar approach like the package.json where things are conditionally added.

Also, we should explain all the different commands we scaffolded. This way people have a good reference on which commands to use.

bennypowers:
closes #917

affects: @open-wc/create

WIP pending review.

Reviewers please note language in README partials especially around test snapshots
Please also note design of the new template variables, should these be higher up the inheritance diagram?


packages/create/src/generators/wc-lit-element/templates/partials/README.testing.md
```diff
@@ -0,0 +1,35 @@
...
+To run the suite of karma tests, run
+```bash
+<%= scriptRunCommand %> test
+```
+or to run them in compatibility mode for legacy browsers
+```bash
+<%= scriptRunCommand %> test:compatibility
+```
+
+To run the tests in watch mode (for <abbr title="test driven development">TDD</abbr>, for example), run
```
LarsDenBakker: Can you mention that you can visit http://localhost:9876/debug.html while in watch mode to debug in the browser?

LarsDenBakker:
This is really great work. I'm missing the building readme?

packages/create/src/generators/wc-lit-element/templates/partials/README.testing.md
bennypowers: TY, added.

bennypowers:
Should the partials be in their own feature's directories?

e.g.
```bash
cd packages/create/src/generators
git mv wc-lit-element/templates/partials/README.demoing.md demoing-storybook/templates/partials/README.demoing-storybook.md
```

bennypowers:
> This is really great work. I'm missing the building readme?

sorry, I don't understand what you mean, please explain.

LarsDenBakker:
What I mean is that I don't see anywhere that we add text to the readme for the building configuration and commands.

bennypowers:
Yes, I noticed that too in the original readme. Worth adding.

Thoughts on :point_up: re location of partials?

LarsDenBakker:
That's an interesting idea, we could also move other configuration/files there. It might make managing the scaffolding easier, because they often get out of sync. Probably we should discuss and do this in a separate PR?

bennypowers:
Original README didn't have building instructions, so we should open a new PR after this one to add a new partial for building.

LarsDenBakker:
Sure whichever way you prefer

bennypowers:
Unless there's anything pressing, this is ready to go (with #937 and #938 opened for later consideration)

PR #924, opened 2019-10-31
Fix/codelab basics 1 text

SwyserDev:
Here are a few suggestions to the text of the codelabs.

### DIFF
--- a/packages/codelabs/src/basics/web-components.md
+++ b/packages/codelabs/src/basics/web-components.md
@@ -222,7 +222,7 @@ If we run this code in the browser, the element should turn blue when clicked on
 
 ## Templating
 
-The second API in Web Components that we will look into is HTML templates. When writing web components,we usually need to do more than just setting some styles or text. We often need to render larger pieces of HTML as part of our component and update parts of it when the user interacts with the page.
+The second Web Components API that we will look into is HTML templates. When writing web components,we usually need to do more than just setting some styles or text. We often need to render larger pieces of HTML as part of our component and update parts of it when the user interacts with the page.
 
 To do this efficiently, the browser provides us with a template element. This template allows us to define the structure of a piece of HTML upfront, and efficiently clones this each time the element is rendered on the page. This is a lot faster than recreating the same HTML structure each time. Writing and cloning templates is (intentionally) pretty low level, you will usually not want to do this manually. You can read more about the basic API [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).
 
--- a/packages/codelabs/src/basics/web-components.md
+++ b/packages/codelabs/src/basics/web-components.md
@@ -178,7 +178,7 @@ class CoolHeading extends HTMLElement {
 ```
 
 <aside class="notice">
-We are adding the event listener in the constructor, because the `connectedCallback` method can be called multiple times during the lifecycle of our element. For example when using drag and drop, an element is disconnected from one location and connected again in another. That would cause the event listener to be registered multiple times, a common source of bugs.
+We're adding the event listener in the constructor, because the `connectedCallback` method can be called multiple times during the lifecycle of our element, for example when using drag and drop, or when an element is disconnected from one location and connected again in another. That would cause the event listener to be registered multiple times, a common source of bugs.
 </aside>
 
 If we run this code in the browser, the element should turn blue when clicked on.
--- a/packages/codelabs/src/basics/web-components.md
+++ b/packages/codelabs/src/basics/web-components.md
@@ -244,7 +244,7 @@ LitElement is written and distributed as ES Modules, this means we can import it
 
 Make sure you add `type="module"` to the script tag!
 
-Next, we need to define our class. Instead of extending `HTMLElement`, we are now extending `LitElement`. `LitElement` extends `HTMLElement` already, so we are still creating an actual custom element:
+Next, we need to define our class. Instead of extending `HTMLElement`, we are now extending `LitElement`. `LitElement` extends `HTMLElement` already, so we're still creating an actual custom element:
 
 ```js
 class WebComponentApis extends LitElement {
--- a/packages/codelabs/src/basics/web-components.md
+++ b/packages/codelabs/src/basics/web-components.md
@@ -358,7 +358,7 @@ LitElement offers a lot more features than just rendering templates, check out o
 
 ## Shadow DOM
 
-The last important API in Web Components we will look into is Shadow DOM. Traditionally, the context of HTML and CSS have always been global. This scales pretty badly, because we then constantly need to make sure that the id's of all the elements are unique and often CSS selectors can get pretty complex. This is why many front-end frameworks offer some form of encapsulation. Web Components provides us with this capability using `Shadow DOM`, this capability is now also built into the browser.
+The last important Web Components API we will look into is Shadow DOM. Traditionally, the context of HTML and CSS have always been global. This scales pretty badly, because we constantly need to make sure that the id's of all the elements are unique and often CSS selectors can get pretty complex. This is why many front-end frameworks offer some form of encapsulation. Web Components provides us with this capability using `Shadow DOM`, this capability is now built into the browser.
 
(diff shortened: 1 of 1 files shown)


packages/codelabs/src/basics/web-components.md
```diff
@@ -45,31 +45,31 @@ To get started, let's create a basic HTML page:
 </html>
 ```
 
-If you run this in the browser and see hello world, you're good to go!
+If you run this in the browser and see hello world, then you're good to go!
```
thepassle: I think this sentence runs better without 'then'

```suggestion
If you run this in the browser and see hello world, you're good to go!
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -45,31 +45,31 @@ To get started, let's create a basic HTML page:
...
 
 ## Custom Elements
 
-First, we will take a look at the most important web component API: custom elements.
+First, we will take a look at the most important web component API: `Custom Elements`.
```
thepassle:
```suggestion
First, we will take a look at the most important web component API: Custom Elements.
```

The name of the specification is 'Custom Elements', but it's not code, so don't use backtics

packages/codelabs/src/basics/web-components.md
```diff
@@ -178,10 +178,10 @@ class CoolHeading extends HTMLElement {
 ```
 
 <aside class="notice">
-We're adding the event listener in the constructor here because the `connectedCallback` can be called multiple times during the lifecycle of our element, for example when using drag and drop an element is disconnected from one location and connected again in another. That would cause the event listener to be registered multiple times, a common source of bugs.
+We are adding the event listener in the constructor, because the `connectedCallback` method can be called multiple times during the lifecycle of our element. For example when using drag and drop, an element is disconnected from one location and connected again in another. That would cause the event listener to be registered multiple times, a common source of bugs.
```
thepassle:
```suggestion
We're adding the event listener in the constructor, because the `connectedCallback` method can be called multiple times during the lifecycle of our element, for example when using drag and drop, or when an element is disconnected from one location and connected again in another. That would cause the event listener to be registered multiple times, a common source of bugs.
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -178,10 +178,10 @@ class CoolHeading extends HTMLElement {
...
 </aside>
 
-If we run this code in the browser, the element should turn blue on click.
+If we run this code in the browser, the element should turn blue when clicked on.
```
thepassle:
```suggestion
If we run this code in the browser, the element should turn blue when clicked.
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -222,13 +222,13 @@ If we run this code in the browser, the element should turn blue on click.
 
 ## Templating
 
-The second Web Components API we will look into is HTML templates. When writing web components, we usually need to do more than just setting some styles or text. We often need to render larger pieces of HTML as part of our component and update parts of it when the user interacts with the page.
+The second API in Web Components that we will look into is HTML templates. When writing web components,we usually need to do more than just setting some styles or text. We often need to render larger pieces of HTML as part of our component and update parts of it when the user interacts with the page.
```
thepassle:
```suggestion
The second Web Components API that we will look into is HTML templates. When writing web components,we usually need to do more than just setting some styles or text. We often need to render larger pieces of HTML as part of our component and update parts of it when the user interacts with the page.
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -244,7 +244,7 @@ LitElement is written and distributed as es modules, this means we can import it
 
 Make sure you add `type="module"` to the script tag!
 
-Next, we need to define our class. Instead of extending `HTMLElement`, we are now extending `LitElement`. `LitElement` extends `HTMLElement` already, so we're still creating an actual custom element:
+Next, we need to define our class. Instead of extending `HTMLElement`, we are now extending `LitElement`. `LitElement` extends `HTMLElement` already, so we are still creating an actual custom element:
```
thepassle:
```suggestion
Next, we need to define our class. Instead of extending `HTMLElement`, we are now extending `LitElement`. `LitElement` extends `HTMLElement` already, so we're still creating an actual custom element:
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -358,15 +358,15 @@ LitElement offers a lot more features than just rendering templates, check out o
 
 ## Shadow DOM
 
-The last important web component API we will look into is shadow DOM. Traditionally, HTML and CSS have always been global. This scales pretty badly, we need to make sure that ids are unique on the page and CSS selector can get pretty complex. This is why many front-end frameworks offer some form of encapsulation. With shadow DOM, this capability is now built into the browser.
+The last important API in Web Components we will look into is Shadow DOM. Traditionally, the context of HTML and CSS have always been global. This scales pretty badly, because we then constantly need to make sure that the id's of all the elements are unique and often CSS selectors can get pretty complex. This is why many front-end frameworks offer some form of encapsulation. Web Components provides us with this capability using `Shadow DOM`, this capability is now also built into the browser.
```
thepassle:
```suggestion
The last important Web Components API we will look into is Shadow DOM. Traditionally, the context of HTML and CSS have always been global. This scales pretty badly, because we constantly need to make sure that the id's of all the elements are unique and often CSS selectors can get pretty complex. This is why many front-end frameworks offer some form of encapsulation. Web Components provides us with this capability using `Shadow DOM`, this capability is now built into the browser.
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -358,15 +358,15 @@ LitElement offers a lot more features than just rendering templates, check out o
...
+The best way to visualize this is to inspect the element we created in the previous step. In the DOM inspector you will see that the children rendered are not direct children of our element, but rather they are wrapped inside of a shadow root:
 
 ![Shadow DOM example](./assets/shadow-dom-1.png)
 
-This shadow root is a special type of dom node, which encapsulates the elements inside it. Styles defined inside this shadow root do not leak out, and styles defined outside the shadow root do not reach in. Also, it's not possible to use `querySelector` to select elements inside or outside the shadow root. This way we can build reusable components and be confident that they will always work the same way in any environment.
+This shadow root is a special type of DOM node which encapsulates the elements inside of it. Styles defined inside this shadow root do not leak out, and styles defined outside the shadow root do not reach in, hence encapsulation. Also, it's not possible to use `querySelector` to select elements inside or outside the shadow root. This way we can build reusable components than gives us the confidence that they will always work the same way, no matter the environment.
```
thepassle:
```suggestion
This shadow root is a special type of DOM node which encapsulates the elements inside of it. Styles defined inside this shadow root do not leak out, and styles defined outside the shadow root do not reach in, hence encapsulation. Also, it's not possible to use `querySelector` to select elements inside or outside the shadow root. This way we can build reusable components and gives us the confidence that they will always work the same way, no matter the environment.
```

packages/codelabs/src/basics/web-components.md
```diff
@@ -358,15 +358,15 @@ LitElement offers a lot more features than just rendering templates, check out o
...
 
-Let's visualize this by adding some styles to our component. In `LitElement`, we can add styles to our component using a static `styles` field:
+Let's test this out by adding some styles to our component. In `LitElement`, we can add styles to our component using a static `styles` field:
```
thepassle:
```suggestion
Let's try this out by adding some styles to our component. In `LitElement`, we can add styles to our component using a static `styles` field:
```

packages/codelabs/src/basics/web-components.md
SwyserDev: I was taught to use contractions sparingly when writing to a large audience, because people that struggle with english get confused.

LarsDenBakker:
Thanks a lot guys

SwyserDev:
twas-my-pleasure-twas-my-pleasure-1889-10-guy-53649786

PR #925, opened 2019-10-31
fix: codelabs build correct path to build file

SwyserDev:
The package.json pointed to the wrong file for "codelabs:build" script.


LarsDenBakker:
Thanks!

PR #939, opened 2019-11-02
feat(es-dev-server): add response transformers

LarsDenBakker:
This adds a new option to the dev server config to transform a response before serving. This has a number of use cases:
- An easier way to write transformations than custom middlewares. Easier for people to understand, and easier to hook into other tools.
- Serve virtual files, such as an auto generated storybook index.html
- Transform javascript **before** it goes through compatibility transforms.
- Light transformations, such as rewriting an app's base path

### DIFF
--- a/packages/es-dev-server/README.md
+++ b/packages/es-dev-server/README.md
@@ -265,7 +265,7 @@ You can rewrite certain file requests using the `responseTransformers` property.
 
 A response transformer is a function which receives the original response and returns an optionally modified response. This transformation happens before any other built-in transformations such as node resolve, babel or compatibility. You can register multiple transformers, they are called in order.
 
-The funtions can be sync or async, the full signature:
+The functions can be sync or async, see the full signature below:
 
 ```typescript
 ({ url: string, status: number, contentType: string, body: string }) => Promise<{ body?: string, contentType?: string } | null>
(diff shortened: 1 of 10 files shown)


bennypowers:
what's the advantage here over passing middlewares? compare the approach of [lws](https://github.com/lwsjs/local-web-server#built-in-middleware-stack)

LarsDenBakker:
@bennypowers Two reasons:
- Easier syntax, people don't need to know koa specifics. They just write a function that receives code and returns code.
- These run before compatibility, so you can return modern code and it would be still run through babel etc.

packages/es-dev-server/README.md
```diff
@@ -253,10 +254,95 @@ module.exports = {
...
+### Response transformers
+
+You can rewrite certain file requests using the `responseTransformers` property. With this you can adjust served files or serve virtual non-existing files programmatically.
+
+<details>
+  <summary>Read more</summary>
+
+A response transformer is a function which receives the original response and returns an optionally modified response. This transformation happens before any other built-in transformations such as node resolve, babel or compatibility. You can register multiple transformers, they are called in order.
+
+The funtions can be sync or async, the full signature:
```
LarsDenBakker:
```suggestion
The functions can be sync or async, see the full signature below:
```

PR #955, opened 2019-11-06
Fix/codelabs text

SwyserDev:
I made some minor changes to the grammar and spelling of the codelabs for lit-html and lit-element.

### DIFF
--- a/packages/codelabs/src/basics/lit-html.md
+++ b/packages/codelabs/src/basics/lit-html.md
@@ -26,17 +26,15 @@ In this codelab, you will learn the basics of building web components with lit-h
 
 **How it works**
 
-This codelab will go step by step, explaining each code change. At the bottom of each section, there is a "View final result" button where you can see the final code result you should end up with. The codelab is sequential, results from the previous step carry on to the next step.
-
-At the bottom of each step, there is a final code snippet to make sure you're still on the right track or to help you when you get stuck.
+This codelab will go step by step, explaining each code change. At the bottom of each section there is a "View final result" button, this will show you the correct code that you should end up with incase you get stuck. The steps are sequential, thus results from the previous steps carry over to the next step.
 
 ## Setup
 
-In this codelab, we will build a simple todo app, a great showcase for learning the basics of lit-html and lit-element.
+In this codelab, we will build a simple todo app, which is a great showcase for learning the basics of lit-html and lit-element.
 
-You can follow this codelab using anything that can display a simple HTML page. We recommend using an [online code editor like jsbin](https://jsbin.com/?html,output), but you could also create a basic html page using your IDE.
+You can follow this codelab using anything that is able to display a simple HTML page. We recommend using an [online code editor like jsbin](https://jsbin.com/?html,output), but you can also create your own html page using your favorite IDE.
 
-To get started, create a simple web page:
+To get started, let's create a basic HTML page:
 
 ```html
 <!DOCTYPE html>
@@ -47,11 +45,11 @@ To get started, create a simple web page:
 </html>
 ```
 
-If you run this in the browser and see the message on the screen, you're good to go.
+If you run this in the browser and see '_My todo app_' on the screen, you're good to go.
 
 ## lit-element setup
 
-lit-element takes care of most of the boilerplate when writing components, providing a great developer experience while staying close to the browser platform. It does not require any build step to run in the browser, and it's only 7kb. This makes it an ideal lightweight choice.
+lit-element takes care of most of the boilerplate when creating components. This provides a great developer experience while staying close to the browser platform. It does not require any build steps to run in the browser. It's a mere 7kb in size and makes it the ideal choice if you are aiming at lightweight components.
 
 lit-element is written and distributed as es modules, this means we can import it using the browser's native module loader. Let's create a module script, and import LitElement from a CDN:
 
@@ -67,13 +65,13 @@ lit-element is written and distributed as es modules, this means we can import i
 </html>
 ```
 
-Make sure you add `type="module"` to the script tag.
+Make sure that you add `type="module"` to the script tag.
 
 <aside class="notice">
-In this example we are using `unpkg`, a CDN where we can easily import any module available on NPM. When working on a real project, you will probably want to use an actual package manager like NPM or yarn.
+In this example we are using `unpkg`, a CDN where we can easily import any module that is available on NPM. When working on a real project, it is a good idea to use an actual package manager like NPM or yarn.
 </aside>
 
-Next, we need to define our web component. When writing a vanilla web component we would extend from `HTMLElement`. With lit-element we need to import and extend from `LitElement`, which itself is an extension of `HTMLElement`.
+Next, we need to define our web component. When writing a vanilla web component we extend from the native `HTMLElement` class. With lit-element we need to import and extend from `LitElement` class which, in turn, is and extension of the `HTMLElement` class.
 
 ```js
 class TodoApp extends LitElement {
@@ -95,7 +93,7 @@ We defined the tag name for our element as `todo-app`, now we need to add this t
 If you run this in the browser you should see `lit element connected` logged to the terminal.
 
 <aside class="notice">
-  Because LitElement also does some work in the `connectedCallback`, we must always call `super.connectedCallback()`. This is a common source of confusion, so make sure to remember!
+  Because LitElement also does some work in the `connectedCallback`, we must always call `super.connectedCallback()`. This is a common source of bugs and confusion, so make sure to remember this!
 </aside>
 
 <details>
@@ -127,9 +125,7 @@ If you run this in the browser you should see `lit element connected` logged to 
 
 ## Hello world
 
-Now that we have defined our element, we can start adding a template. lit-element uses lit-html to handle the templating.
-
-lit-html works by writing HTML inside of template literals. These are a type of strings which can span multiple lines, ideal for writing HTML:
+Now that we defined our element we can start adding a template. lit-element uses lit-html to handle the templating. lit-html works by writing HTML inside of template literals. Template literals are a type of strings which can span multiple lines, ideal for writing HTML:
 
 ```js
 const template = `
@@ -147,11 +143,11 @@ const template = html`
 `;
 ```
 
-This is a native browser feature called [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). The `html` tag is just a function that gets called with information about template literal it's attached to. We won't go into details of how it works exactly, but by using this syntax `lit-html` can very efficiently create templates and update only the parts that change when re-rendering.
+This is a native browser feature called [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). The `html` tag is just a function that gets called with information about the template literal that it's attached to. We won't go into details of how it works exactly. By using this syntax, `lit-html` can very efficiently create templates and update only the parts that changed when re-rendering.
 
-Most popular IDEs support syntax highlighting of HTML inside template literals, but for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
+Most popular IDEs support the highlighting of HTML inside template literals, but for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
 
-lit-element works with a `render` function, which is called each time the element is updated. From this function, we need to return the template we want to render to the page.
+lit-element works with a `render` function that is called each time the element is updated. Inside this function, we need to return the template that we want to render on the page.
 
 Let's start by displaying the title of our app:
 
@@ -201,7 +197,7 @@ If you refresh the browser, you should see the title displayed on the page.
 
 ## Templating basics
 
-Templates are just javascript variables, we can also create them outside the context of our component. For example to share pieces of a template between different components.
+Templates are just javascript variables, we can also create them outside of our component's context. A typical example is when you want to share pieces of a template between different components.
 
 Let's add a footer to our application. First, create the template for the footer:
 
@@ -213,7 +209,7 @@ const footerTemplate = html`
 
 You can add your own name and website in there as author.
 
-Template literals can contain placeholders. These are indicated by a dollar sign with curly braces: `${expression}`. For example, a common way to use it is to build error messages:
+Template literals can contain placeholders. These are indicated by a dollar sign with curly braces: `${expression}`. For example when building error messages:
 
 ```js
 // regular string
@@ -222,7 +218,7 @@ console.error('An error occurred: ' + message);
 console.error(`An error occurred: ${message}`);
 ```
 
-lit-html takes advantage of this feature to compose templates and to create dynamic parts inside your templates. For example, we can add the footer to our app's template by simply embedding it:
+lit-html takes advantage of this feature in order to compose templates and to create dynamic parts inside your templates. For example, we can add the footer to our app's template by simply embedding it:
 
 ```js
 class TodoApp extends LitElement {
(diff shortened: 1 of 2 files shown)


packages/codelabs/src/basics/lit-html.md
```diff
@@ -22,21 +22,19 @@ In this codelab, you will learn the basics of building web components with lit-h
 - Conditional rendering
 - Handling events
 - Managing data
-- Creating child components
+- Creating encapsulated components
```
thepassle:
```suggestion
- Creating child components
```
While its true that they're encapsulated, I think in the context of the codelab it makes more sense to say 'child' component, to demonstrate how parent/child relations work

packages/codelabs/src/basics/lit-html.md
```diff
@@ -47,13 +45,13 @@ To get started, create a simple web page:
 </html>
 ```
 
-If you run this in the browser and see the message on the screen, you're good to go.
+If you run this in the browser and see '_My todo app_' on the screen, you're good to go.
 
 ## lit-element setup
 
-lit-element takes care of most of the boilerplate when writing components, providing a great developer experience while staying close to the browser platform. It does not require any build step to run in the browser, and it's only 7kb. This makes it an ideal lightweight choice.
+[lit-element](https://github.com/Polymer/lit-element) takes care of most of the boilerplating when creating components. This provides a great developer experience while staying close to the browser platform and does not require any build steps to run in the browser. It's a mere 7kb in size and makes it the ideal choice if you care about performance.
```
thepassle:
```suggestion
[lit-element](https://github.com/Polymer/lit-element) takes care of most of the boilerplating when creating components. This provides a great developer experience while staying close to the browser platform and does not require any build steps to run in the browser. It's a mere 7kb in size which makes it an ideal lightweight choice.
```

imo 'if you care about performance' comes off a little stand-offish

packages/codelabs/src/basics/lit-html.md
```diff
@@ -67,13 +65,13 @@ lit-element is written and distributed as es modules, this means we can import i
 </html>
 ```
 
-Make sure you add `type="module"` to the script tag.
+Make sure that you add `type="module"` to the script tag.
 
 <aside class="notice">
-In this example we are using `unpkg`, a CDN where we can easily import any module available on NPM. When working on a real project, you will probably want to use an actual package manager like NPM or yarn.
+In this example we are using `unpkg`, a CDN where we can easily import any modules that are available on NPM. When working on a real project, it is a good idea to rather use an actual package manager such as NPM or yarn.
```
thepassle:
```suggestion
In this example we are using `unpkg`, a CDN where we can easily import any modules that are available on NPM. When working on a real project, it is a good idea to use an actual package manager such as NPM or yarn.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -95,7 +93,7 @@ We defined the tag name for our element as `todo-app`, now we need to add this t
 If you run this in the browser you should see `lit element connected` logged to the terminal.
 
 <aside class="notice">
-  Because LitElement also does some work in the `connectedCallback`, we must always call `super.connectedCallback()`. This is a common source of confusion, so make sure to remember!
+  Because the LitElement class also does some work in the `connectedCallback` method, so we should always call `super.connectedCallback()`. This is a common source of bugs and confusion, so make sure to remember this!
```
thepassle:
```suggestion
  Because the LitElement class also does some work in the `connectedCallback` method, we should always call `super.connectedCallback()`. This is a common source of bugs and confusion, so make sure to remember this!
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -147,11 +143,11 @@ const template = html`
...
+This is a native browser feature called [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). The `html` tag is just a function that gets called with information about the template literal that it's attached to. We won't go into details of how it works exactly. By using this syntax, `lit-html` can very efficiently create templates and update only the parts that changed when re-rendering.
 
-Most popular IDEs support syntax highlighting of HTML inside template literals, but for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
+Most popular IDEs support the highlighting of HTML inside template literals, but for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
```
thepassle:
```suggestion
Most popular IDEs support syntax highlighting of HTML inside template literals, but for some you might need to install a plugin. [See our IDE section](https://open-wc.org/developing/ide.html#visual-studio-code) to learn more about that.
```

more clear

packages/codelabs/src/basics/lit-html.md
```diff
@@ -147,11 +143,11 @@ const template = html`
...
 
-lit-element works with a `render` function, which is called each time the element is updated. From this function, we need to return the template we want to render to the page.
+lit-element has a `render` function that is called each time the element is updated. Inside this function, we need to return the template that we want to be render on the page.
```
thepassle:
```suggestion
lit-element has a `render` function that is called each time the element is updated. Inside this function, we need to return the template that we want to render to the page.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -411,9 +407,9 @@ class TodoApp extends LitElement {
 }
 ```
 
-Because template literals allow us to place any expression inside the curly braces, we can use ternary operators for quick conditional logic.
+Because template literals allow us to place any expression inside of the curly braces, we can use ternary operators for quick and easy conditional logic.
 
-You should now see three todo items on the screen, where the first one is already finished.
+You should now see three todo items on the screen and the first one is already finished.
```
thepassle:
```suggestion
You should now see three todo items on the screen, where the first one is already finished.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -506,15 +502,15 @@ class TodoApp extends LitElement {
 }
 ```
 
-On the add button, we added an event listener for the `click` event. This is done by prefixing the event name with a `@`:
+To the add button we attached an event listener that listens for the `click` event. This is done by prefixing the event name with a `@`:
```
thepassle:
```suggestion
On the add button we attached an event listener that listens for the `click` event. This is done by prefixing the event name with a `@`:
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -527,17 +523,17 @@ _addTodo() {
...
 
-This allows us to observe the power of lit-html in action. If you inspect the DOM while adding a todo item, you will see that only the new todo item is actually flashing:
+This allows us to observe the awesome power of lit-html in action. If you inspect the DOM while adding a new todo item, you will notice that only the new todo item is flashing:
 
 ![todo](./assets/add-todo.gif)
 
-When something in the DOM inspector flashes, it means that the browser is doing actual work to update the DOM tree. This is very expensive work, things like styles and layout need to be recalculated up and down the tree, so you want to minimize this as much as possible. lit-html knows exactly what changed where and it will update only that part efficiently.
+When something in the DOM inspector flashes, it means that the browser is doing actual work to update the DOM tree. This is very expensive, things like the styles and layout need to be recalculated up and down the element tree, so it is wise to minimize this as much as possible. lit-html knows exactly what changed where and it will update only that part, making it much more efficient.
```
thepassle:
```suggestion
When something in the DOM inspector flashes, it means that the browser is doing actual work to update the DOM tree. This is very expensive, things like the styles and layout need to be recalculated up and down the element tree, so it is wise to minimize this as much as possible. lit-html knows exactly what changed where and it will update only that part, making it super efficient.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -608,7 +604,7 @@ In the inspector you also see comment nodes between different parts of your temp
 
 Right now we're triggering updates manually whenever we make a change. This is fine for some use cases, but it can get pretty cumbersome and we are not able to respond to changes triggered by parent components.
 
-It's better to let `LitElement` observe data changes for us, and trigger the updates when necessary. We can do this by defining `todos` as a property of our element.
+It's better to let `LitElement` observe data changes for us and then trigger the updates were necessary. We can do this by defining `todos` as a property of our element.
```
thepassle:
```suggestion
It's better to let `LitElement` observe data changes for us and, trigger updates when necessary. We can do this by defining `todos` as a property of our element.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -643,9 +639,9 @@ class TodosApp extends LitElement {
 
 This way when you change the property on your element, it goes through a custom getter/setter function which triggers an update only when the new value passes a simple equality check.
 
-For strings, numbers and booleans this will work without any problems. However, if you are using arrays or objects and mutate them it will not trigger any update. This is because the actual array or object itself did not change. We need to use immutable data patterns, where a new object is created for each change. This is a common pattern in front-end to simplify data flow and make change detection easier.
+For strings, numbers and booleans this will work without any problems. However, if you are using arrays or objects and mutate them it will not trigger any update. This is because the actual array or object itself did not change. We need to use an immutable data patterns, where a new object is created for each change. This is a common pattern in front-end to simplify data flow and make change detection easier.
```
thepassle:
```suggestion
For strings, numbers and booleans this will work without any problems. However, if you are using arrays or objects and mutate them it will not trigger any update. This is because the actual array or object itself did not change. We need to use immutable data patterns, where a new object is created for each change. This is a common pattern in front-end to simplify data flow and make change detection easier.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -763,7 +759,7 @@ _removeTodo(todo) {
 }
 ```
 
-The delete button should now be fully functional. Because filter returns a new array, `LitElement` will automatically detect the changes and re-render our component.
+The delete button should now be fully functional. Due to the fact that filter returns a new array, `LitElement` will automatically detect the changes and then re-render our component.
```
thepassle:
```suggestion
The delete button should now be fully functional. Because filter returns a new array, `LitElement` will automatically detect the changes and re-render our component.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -908,11 +904,11 @@ and only in this case, hidden is false:
 <div></div>
 ```
 
-A big exception to this are the `aria-` attributes, which do use literal "true" and "false" values. How lovely inconsistent :)
+A big exception to this is `aria-` attributes, which do use literal "true" and "false" values. Notice the inconsistency, how lovely :)
 
 </aside>
 
-We're listening to the input's `change` event to update our data when the checkbox value was changed. Besides the todo object, we are also passing along the event object itself. We need this to get the value of the checkbox.
+We're listening to the input's `change` event, and then update our data when the checkbox value is changed. Besides the todo object, we are also passing along the event object itself. We need this to be able to get the value of the checkbox.
```
thepassle:
```suggestion
We're listening to the input's `change` event to update our data when the checkbox value is changed. Besides the todo object, we are also passing along the event object itself. We need this to be able to get the value of the checkbox.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -1026,11 +1022,11 @@ _changeTodoFinished(e, changedTodo) {
 
 Now that we can manage our todo items, it's pretty easy to display some more information about our list. For example, we can add the total amount of finished and unfinished items.
 
-These are not new sources of data, but rather a simple calculation based on existing data. We could do these calculations inline in our template where we want to display them, but this can get messy pretty quickly.
+These are not new sources of data, but rather a simple calculation based on existing data. We could do these calculations inline in our template where we want to display them, but this can get messy, pretty quickly.
 
-A good practice is to use the top of your render function as a place to prepare variables with some meaningful names and to use those in your template for best readability.
+Good practice is to use the top of your render function as a place to prepare variables with some meaningful names and then use them in your template. This provides the best readability.
```
thepassle:
```suggestion
Good practice is to use the top of your render function as a place to prepare variables with some meaningful names to use them in your template. This provides the best readability.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -1026,11 +1022,11 @@ _changeTodoFinished(e, changedTodo) {
...
 
-Let's add our calculations to the render function, and display the calculated amounts in the template:
+Let's add our calculations to the render function, and then display the calculated amounts in the template:
```
thepassle:
```suggestion
Let's add our calculations to the render function, and display the calculated amounts in the template:
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -1065,7 +1061,7 @@ render() {
 }
 ```
 
-Remember that the render function can be called quite often. If the computations are pretty expensive it's better to only do them once and cache the results.
+Remember that the render function sometimes gets called quite frequently. If the computations are expensive, it's better to only do them once and then cache the results.
```
thepassle:
```suggestion
Remember that the render function can be called quite often. If the computations are expensive, it's better to only do them once and cache the results.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -1408,25 +1404,25 @@ render() {
 }
 ```
 
-If you refresh, the UI should remain unchanged. Congratulations, you're already composing elements like a pro :)
+If you refresh, the UI should remain unchanged. If that is the case, congratulations! You're now composing elements like a pro :)
 
-In the template we added, we're seeing something different than before. We're passing along our list of todos to the `todo-list` using an attribute with a `.` in front of it:
+In the template that we added we're seeing something different than before. We're passing along our list of todos to the `todo-list` using an attribute with a `.` in front of it:
```
thepassle:
```suggestion
In the template we added, we're seeing something different than before. We're passing along our list of todos to the `todo-list` using an attribute with a `.` in front of it:
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -1408,25 +1404,25 @@ render() {
...
 
 ```js
 html`
   <todo-list .todos=${this.todos}></todo-list>
 `;
 ```
 
-This is a special lit-html syntax, where instead of setting an attribute it will set a property on the element. The differences between attributes and properties are something that confuses a lot of people, but it's a very simple concept.
+This is lit-html specific syntax that allows you to set a property on the element instead of an attribute. The differences between attributes and properties often confuses a lot of people, but it's actually a very simple concept to grasp.
```
thepassle:
```suggestion
This is lit-html specific syntax that allows you to set a property on the element instead of an attribute. The differences between attributes and properties often confuses a lot of people, but it's a very simple concept.
```

packages/codelabs/src/basics/lit-html.md
```diff
@@ -1558,10 +1554,10 @@ class TodoList extends LitElement {
 customElements.define('todo-list', TodoList);
 ```
 
-The styles we define here only apply to our element. This is a guarantee built into the browser because we're using shadow DOM, lit-element does not need to do any extra work for that. This means we can write simple CSS selectors, and not need to worry about creating conflicts with styles defined elsewhere on the page.
+The styles we define here only apply to our element. This is guaranteed, because we are using shadow DOM the is built in the browser and lit-element does not need to do any extra work. This means we can write simple CSS selectors, and we don't need to worry about causing conflicts with styles defined elsewhere on the page.
```
thepassle:
```suggestion
The styles we define here only apply to our element. This is guaranteed, because we are using shadow DOM which is built in the browser and because of that, lit-element doesn't need to do any extra work. This means we can write simple CSS selectors, and we don't need to worry about causing conflicts with styles defined elsewhere on the page.
```

thepassle:
@SwyserDev Thanks for the contribution!

LarsDenBakker:
This is really great, thanks a lot!

PR #956, opened 2019-11-06
fix (es-dev-server): use `require` to resolve `@babel/preset-env`

goremikin:
`@babel/preset-env` should be resolved relative to es-dev-server location


daKmoR:
thankkk you 🤗

PR #970, opened 2019-11-09
WIP: chore: integration tests 

alex-nt:
A first example of an integration test. I need a bit feedback to see if I should continue this way or if I should seek another way. It is based on https://medium.com/@zorrodg/integration-tests-on-node-js-cli-part-1-why-and-how-fa5b1ba552fe. Can somebody look over this? @bennypowers?


bennypowers:
Let's try using `mock-fs` so we don't need to actually write to disk. I have a drop of work prepared in that direction, I'll share it when I get back on that machine.

alex-nt:
I'm also having some problems with running directly (from `src`) the `create.js` file. Apparently node does not support imports in `.js` files. :( Any idea? Should I just rename everything to `.mjs`?

bennypowers:
run node with `-r esm` to support import statements

[esm](https://github.com/standard-things/esm)

alex-nt:
Hmmm, it fails to create a new process on circleci but works on local 😢.

bennypowers:
I'm getting the same error locally that I'm seeing on CI. What command do you use to run this test locally?

bennypowers:
The team discussed the issue of integration tests, and we've decided to start with a different approach (#976). We want to do that because it will let us get the tests out sooner while focussing output more than on input, i.e. at this stage we can avoid testing keyboard UI without impacting on the level of confidence we derive from the tests.

That being said, since you've invested time and have experience in that area, We'd happily review a PR which adds that capability on top of the other tests. If you like, you can open a new PR with your existing branch rebased on to the new tests.

We'd also value your input on the other questions of what #976 doesn't address, specifically, a handy way to generate snapshots and a sensible convention on where to save them.

Thanks again for the effort here and we look forward to reviewing future PRs.

PR #973, opened 2019-11-10
setAttribute instead of getAttribute

hackdaworld:
The number of arguments passed to the function assumes it should be setAttribute instead of getAttribute.


LarsDenBakker:
You're right, thanks! Can you sign the CLA?

hackdaworld:
I signed the CLA. However, I used an invalid email address for this commit. I think I have to do a new pull request ...

PR #974, opened 2019-11-10
setAttribute instead of getAttribute

hackdaworld:
The number of arguments passed to the function assumes it should be setAttribute instead of getAttribute.


thepassle:
@hackdaworld Thanks! 🥳

Issue #941, opened 2019-11-02
[create] integration tests


bennypowers:
#919 introduced a regression which caused the `create` generator to crash if no options were selected. To prevent this kind of regression in the future, we should introduce some integration tests.

alex-nt:
I'll try to work on this. I'll look into the code in the morning and see how far I can get.

PR #976, opened 2019-11-11
test(create): add basic integration test for app with all features

bennypowers:
fixes: #941
closes: #970

## What this PR Does:

tests that the following command
```
node -r @babel/register packages/create/src/create.js \
  --destinationPath ./output \
  --type scaffold \
  --scaffoldType app \
  --features linting testing demoing building \
  --buildingType rollup \
  --scaffoldFilesFor testing demoing building \
  --tagName scaffold-app \
  --writeToDisk true \
  --installDependencies false
```

in fact creates the desired output files.

Each output file's content is checked against the snapshot.

## What this PR Doesn't Do:
- test any other commands
- test the wizard UI
- test the keyboard input
- provide an automated way to update the snapshots
- provide clear guidance or convention for naming snapshots based on input/test case


LarsDenBakker:
I like this. We are committing the whole generated output as a snapshot, but I think it's a nice way to keep a grasp on what the generator does and that our changes don't mess it up.

thepassle:
I think this looks really nice :)

I wonder if it would be possible to run lint/test on the output files? It has happened in the past tht the generator runs fine and scaffolds some files, but that there was a mistake in the files (wrong import, introduced  a linting problem, etc)

if we can somehow run linting and tests we can be really confident that we didnt break anything

bennypowers:
@thepassle I'm going to look into running the linter on the output in tests

only reason i can think of not to would be that `npm i` will be expensive in terms of CI time. we could hack around that by treating the ACTUAL test output as a package in the monorepo, for purposes of linting

bennypowers:
@thepassle 

https://github.com/open-wc/open-wc/blob/91321f43ed939aef70a6282f4f24265b8906078c/packages/create/test/integration.test.js#L121-L126

thepassle:
So nice 😄 Can we run tests on the output too?

bennypowers:
> output ?
`ACTUAL_PATH` is the output, so this runs eslint programmatically on the generated files

AFAICT it's the same as `cd scaffold-app && npx eslint`

thepassle:
? Not sure what you mean with your last comment

bennypowers:
edited. I mean this test *does* lint the generator output

bennypowers:
🤦‍♂ oh you mean *TESTS*

uhh yeah let me check

bennypowers:
so, these tests run in a node context.. i suppose i could spin karma up in a child process

bennypowers:
discussing it on slack, we agreed it's best to leave this at simple tests for now. later on we'll implement proper e2e tests that run, install, and test output in a separate ci process

PR #990, opened 2019-11-14
feat(es-dev-server): add auto compatibility mode

LarsDenBakker:
This adds a new `auto` compatibility mode. In this mode, we look at the browser's user agent and decide what kind of compatibility to do for that specific browser. 

On modern browsers (latest 2 of chrome, safari firefox (we can add edge in jan), we don't do anything. This keeps the server as fast as possible, even on compatibility mode. 

On other browsers, we use `@babel/preset-env` and target this one browser specifically. We also use some checks using `caniuse-api` in order to see if this specific browser supports es module features (module script, dynamic import and import.meta.url) and compile to systemjs if necessary.

Because of this new mode, compatibility `all` and `modern` are now redundant. These modes will now log a deprecation warning, and default to `auto` mode. This avoids creating a break change.

There might be cases where `auto` isn't doing it's job properly, as a fallback there are `min` and `max` modes which force compilation. `min` is latest 2 chrome, safari, firefox & edge + es modules. `max` is es5 + systemjs.

Compatibility mode can be turned off with the existing `none` mode.

TODO:
- [ ] Update readme
- [x] Update tests
- [x] Add polyfills flag
- [x] Update karma-esm
- [ ] Update all uses of `es-dev-server` to remove compatibility flags
- [ ] Update all uses of `karma-esm` to remove compatibility flags
- [x] Fine tune which testing libs are excluded from babel/module
- [x] Ensure file caching works properly in testing
- [x] Fix assets 404ing
- [ ] Re-add failed test import logging
- [x] Fix closing testing throwing errors


LarsDenBakker:
We're releasing this in https://github.com/open-wc/open-wc/pull/996

PR #992, opened 2019-11-17
Chore/improve ci caching

daKmoR:


daKmoR:
tested the approach - now merged to master via a different PR

PR #993, opened 2019-11-17
chore: cache dependencies on circleci

daKmoR:
💼 Cache your dependencies
🔑 Use your `yarn.lock` as a cache key
📦 Run `yarn install` only if no cache was found
🏅 Win time

PS: For us it was 54s to do yarn install vs 8s to restore cache => 84% faster 🎉
PPS: when there are no dep changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,14 +17,20 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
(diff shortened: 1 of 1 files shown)


daKmoR:
jup just strings... it's sort of a "convention" so you just bump the "v" version if you wanna completely invalidate the cache

PR #1000, opened 2019-11-19
feat(building-rollup): dedupe lit-html by default

### LINKED ISSUE #748: [building] rollup-plugin-node-resolve dedupe option
We should document the new dedupe option from `rollup-plugin-node-resolve`, and set up some defaults.

bennypowers:
exposes `dedupeModules` option in config generator options.

closes #748

affects: @open-wc/building-rollup

Relevant section of deploy preview:
https://deploy-preview-1000--open-wc-org.netlify.com/building/building-rollup.html#deduplicating-node-modules

### DIFF
--- a/packages/building-rollup/README.md
+++ b/packages/building-rollup/README.md
@@ -13,20 +13,20 @@ The input for rollup is the same `index.html` you use for development. Any modul
 
 ## Setup
 
-### New project
+### New Project
 
 ```bash
 npm init @open-wc
 ```
 
-### Existing project
+### Existing Project
 
 ```bash
 npm init @open-wc
 # Upgrade > Building > Rollup
 ```
 
-### Manual setup
+### Manual Setup
 
 1. Install the required dependencies:
 
@@ -77,7 +77,7 @@ We use [rollup-plugin-index-html](https://open-wc.org/building/rollup-plugin-ind
 - `start:build` runs your app after it has been built using the build command
 - `build` builds your app and outputs it in your `dist` directory
 
-## Supporting legacy browsers
+## Supporting Legacy Browsers
 
 `createDefaultConfig` works for browsers which [support es modules](https://caniuse.com/#feat=es6-module). If you need to support older browsers such as IE11 you need to use our `createCompatibilityConfig` in your `rollup.config.js`:
 
@@ -91,7 +91,7 @@ In addition to outputting a regular build of your app, it outputs a legacy build
 
 At runtime we determine which version of your app should be loaded, so that legacy browsers don't force you to ship more and slower code to most users on modern browsers.
 
-## Config features
+## Config Features
 
 `createDefaultConfig`:
 
@@ -132,7 +132,7 @@ At runtime we determine which version of your app should be loaded, so that lega
     - minify + treeshake js
     - minify html and css in template literals
 
-## Adjusting browser support for the modern build
+## Adjusting Browser Support for the Modern Build
 
 The legacy build targets IE11, which is the earliest browser supported by the webcomponents polyfill. For the modern build we target the 2 most recent versions of the major browsers (chrome, firefox, safari and edge).
 
@@ -140,7 +140,7 @@ You can adjust this by adding a [browserslist](https://github.com/browserslist/b
 
 > Warning: you should not add IE11 or other very early browsers as a target in your browserslist, as it would result in a broken modern build because it makes some assumptions around browser support. Use the `--legacy` flag for legacy builds.
 
-## Customizing the babel config
+## Customizing the Babel Config
 
 You can define your own babel plugins by adding a `.babelrc` or `babel.config.js` to your project. See [babeljs config](https://babeljs.io/docs/en/configuration) for more information.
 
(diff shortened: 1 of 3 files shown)


bennypowers:
Reviewers, will this have knock-on effects in other packages?

packages/building-rollup/README.md
```diff
@@ -263,7 +263,42 @@ export default merge(basicConfig, {
...
+
+Node module resolution is fundamentally different to browser behaviour. Node resolution allows - and in fact encourages - multiple copies of single modules to exist in your app's module graph. This can cause issues for libraries which expect to be loaded only once by the browser. `lit-html` is one such library, which uses [`instanceof`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) extensively throughout the source. If it's loaded twice, [bugs can occur](https://github.com/Polymer/lit-html/issues/320#issuecomment-390768466).
+
+To avoid this problem, we've included `lit-html` in the [list of packages for `rollup-plugin-node-resolve` to deduplicate](https://github.com/rollup/rollup-plugin-node-resolve/pull/201). You can customize this list yourself by adding your module names to the `dedupeModules` option of the config generator.
+
+```js
```
LarsDenBakker: The convention for this right now is to add a boolean to the object passed to the config creator to turn off a plugin. This is a pretty nice way as well, but we should be consistent here.

packages/building-rollup/modern-config.js
```diff
@@ -14,6 +14,8 @@ const getWorkboxConfig = require('@open-wc/building-utils/get-workbox-config');
 
 const production = !process.env.ROLLUP_WATCH;
 
+const DEFAULT_DEDUPE_MODULES = ['lit-html'];
```
LarsDenBakker: I think we should add `lit-element` here as well.

LarsDenBakker:
Is there a 'dedupe all' option? I'd personally go for that.

packages/building-rollup/modern-config.js
bennypowers: `rollup-plugin-node-resolve`'s `dedupe` option can take as its value either an array of package names, or a function of `importee -> boolean`

So we could pass `() => true` to dedupe everything, meaning our option's type could be `Boolean | Function | 'all' | string[]`, and in case of 'all', we'd pass that constant true func.

bennypowers:
`rollup-plugin-node-resolve`'s `dedupe` option can take as its value either an array of package names, or a function of `importee -> boolean`

So we could pass `() => true` to dedupe everything, meaning our option's type could be `Boolean | Function | 'all' | string[]`, and in case of 'all', we'd pass that constant true func.

packages/building-rollup/modern-config.js
bennypowers: Yes, I see now that lit-element uses `instanceof` for `css` tagged templates
https://github.com/Polymer/lit-element/blob/f552fc47b12bf3e272aaeafe03d9d73bc4fff1f5/src/lib/css-tag.ts#L64

packages/building-rollup/README.md
bennypowers: So i'll add a check for `dedupe === false`. However, I think this paragraph should stay in the docs, in case a user has some weird reason for wanting to override the defaults but still dedupe. Is that unreasonable?

LarsDenBakker:
Sounds good to me, I think you would only want duplicate packages in specific use cases

packages/building-rollup/README.md
LarsDenBakker: They would pass false, and then add their own version of the plugin I think

bennypowers:
Per our discussion in slack, we might want to change the base of this MR to some `dedupe` branch, which would include support for deduping modules in `building-webpack` and `es-dev-server` as well.

At the moment it would seem all three require their own bespoke implementations, until support for import maps lands in the bundlers.

packages/building-rollup/modern-and-legacy-config.js
```diff
@@ -16,6 +16,8 @@ const getWorkboxConfig = require('@open-wc/building-utils/get-workbox-config');
 const production = !process.env.ROLLUP_WATCH;
 const prefix = '[owc-building-rollup]';
 
+const DEFAULT_DEDUPE_MODULES = ['lit-html', 'lit-element'];
```
LarsDenBakker: Unfortuantely, we need to list all deep imports here as well :( see: https://github.com/rollup/plugins/issues/82

We could scan the file system for all modules perhaps?

packages/building-rollup/modern-config.js
```diff
@@ -25,6 +27,7 @@ module.exports = function createBasicConfig(_options) {
     outputDir: 'dist',
     extensions: DEFAULT_EXTENSIONS,
     indexHTMLPlugin: {},
+    dedupeModules: [],
```
LarsDenBakker: we should stick to only flags to disable plugins entirely, and document how people can add their own configuration.

in this case it's easy to add your own copy of node-resolve I think

bennypowers:
Considering #748 is closed, I'm closing this PR.

PR #1003, opened 2019-11-19
fix(create): change storybook config for app generator

### LINKED ISSUE #999: [demoing-storybook] storybook of newly scaffolded application fails
Storybook fails with
```sh
ERROR in ./.storybook/config.js
Module not found: Error: Can't resolve '@storybook/polymer' in '/Users/kuko/Workspace/Everifin/app/everifin-app/.storybook'
 @ ./.storybook/config.js 6:0-61 18:0-12 19:0-9
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./node_modules/@storybook/addon-docs/dist/frameworks/common/config.js ./node_modules/@storybook/addon-docs/dist/frameworks/common/../../../dist/frameworks/web-components/config.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true&quiet=true
```

my `unchanged` storybook's config file looks like this:
```js
import { configure, addDecorator } from '@storybook/polymer';
import { withA11y } from '@storybook/addon-a11y';
import '@storybook/addon-console';

const req = require.context('../components', true, /\.stories\.js$/);
function loadStories() {
  req.keys().forEach(filename => req(filename));
}

addDecorator(withA11y);
configure(loadStories, module);
```

I am not sure why, because I can see different config file of appropriate generator in source code
https://github.com/open-wc/open-wc/blob/master/packages/create/src/generators/demoing-storybook/templates/static/.storybook/config.js

Moreover, when I manually change that file, storybook is complaining about missing `custom-elements.json`, that is not automatically generated.

kuka-radovan:
fixes #999

These changes was forgotten in #907


daKmoR:
thaaankkk you 🤗

PR #1008, opened 2019-11-20
fix(demoing-storybook): escape quotes in Storybook scripts

amielucha:
Resolves an issue when the stories source pattern was not being recognized due to incorrect interpretation of single quotes on Windows.


LarsDenBakker:
Thanks!

PR #1015, opened 2019-11-20
chore: update lockfile

LarsDenBakker:
This updates the lockfile, which we should do periodically.

A lot of file changed as the snapshots hashes updated (would be great to not have hashes in these tests at some point)

I had to update the the rollup-plugin-index-html test to handle new rollup syntax.


LarsDenBakker:
@daKmoR did you merge this as part of the other merge?

daKmoR:
no I did not... probably needs to be regenerated as some versions of demoing-storybook changed 🙈

Issue #967, opened 2019-11-07
[lit-helpers] spread directives


LarsDenBakker:
We are thinking about adding some lit-html helpers for commonly requested functionality, starting with a spread directive.

Currently I'm thinking of this syntax:
```js
html`
  <my-element ...=${spreadBindings({ 'foo': 'bar', '.foo': 'bar', '@foo': 'bar', '?foo': 'bar' })}>
`
```

Where the keys aligns with the lit-html syntax of setting attributes, properties and event listeners.

Additionally because setting properties or attributes is the most common use case, I think we should add a helper for that:
```js
html`
  <my-element ...=${spreadProps({ 'foo': 'bar' })}>
  <my-element ...=${spreadAttributes({ 'foo': 'bar' })}>
`
```

We're looking to gather feedback on what people think of this, and if this is how they expect this to work.

See also the discussion on adding this to lit-html: https://github.com/Polymer/lit-html/issues/923

bennypowers:
We discussed this on Slack and came to the following conclusions:

- we might have preferred something like
```js
html`<x-l 
    ...="${{ prop2: 1, prop2: 2 }}"
    ..$="${{ attr1: '1', attr2: '2' }}"
    ..?="${{ bool1: true, bool2: false }}"
    ..@="${{ click: this.onClick }}"
></x-l>`
```
But this would require upstream changes to `lit-html`, since an attribute directive doesn't have access to the attribute name. Therefore, we prefer to export specifically-named directives for props, attrs, listeners, etc. This also means that the attr name is of no consequence:
```js
html`
  <x-l atty-mc-attrface="${spreadProps({ foo, bar, baz })"></x-l>
`
```

PR #1023, opened 2019-11-22
feat(lit-helpers): add spread directives

LarsDenBakker:
This introduces a new package, `@open-wc/lit-helper`, with two spread directives `spread` and `spreadProps`.

I considered adding `spreadAttributes`, but I felt that it wasn't needed because you can already use `spread` to do the same since there is no syntax difference for attributes.

Fixes https://github.com/open-wc/open-wc/issues/967

### DIFF
--- a/packages/lit-helpers/README.md
+++ b/packages/lit-helpers/README.md
@@ -1,0 +1,108 @@
+# Lit Helpers
+
+[//]: # 'AUTO INSERT HEADER PREPUBLISH'
+
+A library with helpers functions for working with [lit-html](https://lit-html.polymer-project.org/) and [lit-element](https://lit-element.polymer-project.org/)
+
+## Installation
+
+```bash
+npm i --save @open-wc/lit-helpers
+```
+
+## Spread directives
+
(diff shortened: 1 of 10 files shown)


packages/lit-helpers/README.md
```diff
@@ -0,0 +1,108 @@
...
+renderSpread({ foo: undefined' });
+```
+
+<script>
+  export default {
+    mounted() {
+      const editLink = document.querySelector('.edit-link a');
+      if (editLink) {
+        const url = editLink.href;
+        editLink.href = url.substr(0, url.indexOf('/master/')) + '/master/packages/testing-helpers/README.md';
```
daKmoR:
```suggestion
        editLink.href = url.substr(0, url.indexOf('/master/')) + '/master/packages/lit-helpers/README.md';
```

LarsDenBakker:
There is a section in the docs that explains why no `spreadAttributes`.

For sure spreading adds some overhead, it should only be used in scenario's where you just don't know all the keys. We can add a warning in the docs for that.

LarsDenBakker:
commitlint makes this PR appear red, but the commit will be squashed so it's fine.

bennypowers:
should probably rebase manually to appease commitlint

Issue #912, opened 2019-10-28
[eslint] eslint-plugin-lit


LarsDenBakker:
https://www.npmjs.com/package/eslint-plugin-lit is excellent, but we don't include it in our config. Our config should work also when you're not using lit, but since it's our default recommendation I think we should offer this somehow.

How shall we do this? A separate package? I don't think an eslint config can have multiple exports, right?

The `create` package would also need to update and include the correct linting configuration

bennypowers:
What if we added it to the `eslintrc` for the generators, but not to the `@open-wc/eslint-config` package?

LarsDenBakker:
I feel that people would still miss out. Also, the lit plugin requires some configuration regarding best practices. Id say lets make a separate package.

thepassle:
> What if in addition to adding it to the docs we added it to the `eslintrc` for the generators, but not to the `@open-wc/eslint-config` package?

That used to be the case, but looking at the generator now, I dont see it there anymore. I guess it got lost in the refactor.

I'd be up for having a lit configuration, but we have to be very clear in how we communicate we have a `vanilla` config and a `lit` config.

LarsDenBakker:
Now that I think about it, how bad is it to just include this plugin in the default config? Most of it's users will use lit, and for those who don't the plugin just won't do anything. There are lots of linting options that won't ever apply to many people.

Our linting doesn't include that much special anyway, it's easy for people to assemble their own.

The lit plugin isn't lit-element specific either, so for example people using haunted will benefit from it as well.

LarsDenBakker:
@daKmoR I tried adding this to our linting config, and now it's giving errors for tests where we're doing things like ``` html`<${tagName}></${tagName}` ```. The linter doesn't like that, of course :)

I know this is something you use a lot for tests, how do you think we should handle this? Should we still add these linting rules to our default config, and then projects should overwrite it if needed?

bennypowers:
i think we have two options 
1. override for our tests
2. add an exception for tests to the plugin

LarsDenBakker:
Can you disable specific linting rules per package? I think option 1 is ok for now, it's a more specialist use case anyway.

PR #1024, opened 2019-11-22
feat(eslint-config): add eslint-plugin-lit

LarsDenBakker:
This adds eslint-plugin-lit, except for tests and demos where we sometimes want to do funky things.

This is technically a breaking change, that's convenient for some and inconvenient for others. What do you guys think?

Fixes https://github.com/open-wc/open-wc/issues/912


thepassle:
LGTM, how should we go about the breaking change?

LarsDenBakker:
It should be in the commit message, @daKmoR double check if I formatted the message correctly? :)

thepassle:
according to conventional commits there needs to be a colon after `BREAKING CHANGE`: https://www.conventionalcommits.org/en/v1.0.0-beta.4/

im not sure if it matters though 🙈

LarsDenBakker:
Let's change the message to that when we squash the commits. I need someone to approve the PR though :)

PR #1030, opened 2019-11-23
Feat/create tests

LarsDenBakker:
- adds a11y and snapshot tests to project generator
- improved some of the code formatting of the app generator


thepassle:
LGTM, except the snapshots need to be updated

(Ran the generator with the new tests, installed, and ran the tests; all pass)

Issue #1045, opened 2019-11-25
[create] generated snapshot tests all test owc-app


LarsDenBakker:
We (I) recently added snapshot and a11y tests to the app generator, but due to the art of copy + paste all of them test the `<owc-app>` element rather the specific element being tested.

We should update this to test the correct element, and update the snapshot files.

PR #1046, opened 2019-11-25
Fix/correct snapshots

thepassle:
fixes https://github.com/open-wc/open-wc/issues/1045


bennypowers:
Am I correct in assuming this should have been caught by (as of yet unwritten) e2e tests?

PR #1051, opened 2019-11-26
docs: document preserveSymlinks option

Viniter:
- Add mentions of `preserveSymlinks` option to `karma-esm` and `testing-karma` docs
- Improve `es-dev-server` doc
- Improve `preserveSymlink` related error message, to be more useful when it comes up while trying to run tests


LarsDenBakker:
Thanks, this is definitely an improvement!

PR #1053, opened 2019-11-27
docs: pwa-ify docs site

thepassle:
Pwa-ifies the docs site. Will show an update popup when new content is available.

Screenshot 2019-11-27 at 10 39 00

Screenshot 2019-11-27 at 10 39 51


daKmoR:
I personally find these popups super annoying 😭 

could we find another way of achieving it? reload on-page navigation or do it like it's on lernjavascriptonline - where it is an indication "dot" in the burger menu (which then shows a "Update content" menu point)

I know it's probably tough but these popups are imho one of the worst trend on the web...

thepassle:
We can also just call skipWaiting and clientsClaim, but i'll have to see how to override the workbox config vuepress uses

docs/.vuepress/config.js
```diff
@@ -119,9 +119,19 @@ module.exports = {
...
   },
   dest: '_site',
-  plugins: ['@vuepress/google-analytics'],
+  plugins: [
+    '@vuepress/google-analytics',
+    [
+      '@vuepress/pwa',
+      {
+        serviceWorker: true,
+        updatePopup: true,
```
thepassle: remove updatePopup, and add `generateSWConfig`, which are the generateSW options from workbox

thepassle:
@daKmoR 
Updated to disable the popup from showing, and use `skipWaiting` and `clientsClaim` instead.

What this means is that if:
- a user has already visited open-wc, and has an activated service worker
- we make some changes, deploy the docs site again
- on the users next visit to our docs site, that user will still get the 'old' docs site, but the service worker will call `clientsClaim` and `skipWaiting` and start caching the new files
- so that on the _next_ visit, the user gets the latest docs

What do you think?

Westbrook:
I think for a site like ours where use terms are probably pretty low, but revisiting is probably pretty high (at least based my usage) that the pattern behind `skipWaiting` and `clientsClaim` should be great. We can massage later if we want to be more aggressive with updates.

LarsDenBakker:
Agree, let's move forward with this PR

daKmoR:
jup let's improve for now we can make it even nicer if needed later

thepassle:
Oki, i just need to add some tags to the head and icons etc

daKmoR:
@thepassle you let us know once it's ready to review/ship right?

thepassle:
Yep, just havent had my laptop open much over the holidays 😄 Will finish this sometime this week

LarsDenBakker:
Such bliss

thepassle:
Actually this should be good to go

thepassle:
Not stale, this can be merged

thepassle:
This was already merged in a different PR

PR #1069, opened 2019-11-30
Added Sublime Text 3

maartenst:


Westbrook:
@maartenst the code here looks good, but CommitLint is very particular with its commit syntax and for good long term repo health reasons. Can you update your commit to be inline with https://github.com/conventional-changelog/commitlint#what-is-commitlint?

maartenst:
better now?

Westbrook:
Thanks for your contribution!

PR #1073, opened 2019-12-01
Feat/live binding

LarsDenBakker:
This implements a `live` directive, based on discussions at https://github.com/Polymer/lit-html/issues/877

### DIFF
--- a/packages/lit-helpers/test/live.test.js
+++ b/packages/lit-helpers/test/live.test.js
@@ -3,19 +3,21 @@ import { spy } from 'sinon';
 import { render } from 'lit-html';
 import { live } from '../src/live.js';
 
-class MyElement extends HTMLElement {
-  set myProp(value) {
-    this._myProp = value;
-  }
-
-  get myProp() {
-    return this._myProp;
-  }
-}
-
-customElements.define('my-element', MyElement);
-
 describe('live', () => {
+  before(() => {
+    class LitHelpers extends HTMLElement {
+      set myProp(value) {
+        this._myProp = value;
+      }
+
+      get myProp() {
+        return this._myProp;
+      }
+    }
+
+    customElements.define('lit-helpers', LitHelpers);
+  });
+
   describe('property bindings', () => {
     let wrapper;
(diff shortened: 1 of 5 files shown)


packages/lit-helpers/test/live.test.js
```diff
@@ -0,0 +1,132 @@
+import { expect, html, fixture } from '@open-wc/testing';
+import { spy } from 'sinon';
+import { render } from 'lit-html';
+import { live } from '../src/live.js';
+
+class LitHelpers extends HTMLElement {
```
daKmoR: that should go in a before or directly in the function as otherwise it will always be executed even if another test has .only

packages/lit-helpers/test/live.test.js
```diff
@@ -0,0 +1,132 @@
...
+customElements.define('lit-helpers', LitHelpers);
+
+describe('live', () => {
+  describe('property bindings', () => {
+    let wrapper;
+    beforeEach(async () => {
+      wrapper = await fixture(document.createElement('div'));
+    });
+
+    function renderLive(value) {
```
daKmoR: probably should go in a before?

packages/lit-helpers/test/live.test.js
LarsDenBakker: Does that really matter though? I can put it in a before since it's a side effect, but I'm not sure what difference it would make.

packages/lit-helpers/test/live.test.js
LarsDenBakker: I don't think so, it's just a function definition.

packages/lit-helpers/test/live.test.js
LarsDenBakker: Added a before for this

ernsheong:
My scroll is in a <div>, not a custom element, how can I actually use this on a div?

LarsDenBakker:
You can use the directive on any element in a lit html template

ernsheong:
Thanks for your fast reply. This works!

image

PR #1077, opened 2019-12-03
feat(es-dev-server): allow max compat on modern browsers

LarsDenBakker:
Loads regenerator runtime on modern browsers as well in max compat mode, this way you can use max compat as a fallback to work anywhere.

### DIFF
--- a/packages/building-utils/index-html/create-index-html.js
+++ b/packages/building-utils/index-html/create-index-html.js
@@ -49,7 +49,7 @@ const { createContentHash, cleanImportPath, polyfillFilename } = require('./util
 /**
  * @typedef {object} PolyfillsConfig
  * @property {boolean} [coreJs] whether to polyfill core-js polyfills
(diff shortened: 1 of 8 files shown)


LarsDenBakker:
@daKmoR @Westbrook @bennypowers @thepassle got the CI green on this now, if you have a chance to look at this

LarsDenBakker:
Max compatibility compiles to es5, which compiles `for of` loops and async functions to use regenerator. Currently we load the regenerator library using `nomodule` because we make the assumptions that we don't serve es5 to modern browsers. This isn't the case in max compatibility mode, where we force es5 on all browsers. In that case we need to always load the regenerator runtime as well, which is what this change does. 

Max compatibility is useful to have a safe fallback of something that's compatible with all browsers, if auto detection fails or when you're testing on an obscure browser.

Regarding the comments, I think there is a why comment in the change I did here. Are you referring to the code for the regenerator in general? There can be improvements there.

Issue #1083, opened 2019-12-05
[es-dev-server] dynamic import with variables not compiled to es5


LarsDenBakker:
imports like this: 
```
import(`/foo/${bar}.js`)
```

are not compiled to es5 correctly

PR #1092, opened 2019-12-06
fix(es-dev-server): compile generated systemjs template literals

LarsDenBakker:
Apparently the module transform adds template literals in some situations, we need to compile those away for IE11.

Fixes https://github.com/open-wc/open-wc/issues/1083


thepassle:
LGTM

PR #1094, opened 2019-12-08
[building] support optional chaining and nullish coalescing

LarsDenBakker:
It might be a while before https://github.com/babel/babel/pull/10811 is released, I suggest we move forward with supporting optional chaining and nullish coalescing.


thepassle:
LGTM 👍

PR #1097, opened 2019-12-08
feat(es-dev-server): add module dedupe option

LarsDenBakker:
This adds a `dedupe` option to go along with https://github.com/open-wc/open-wc/pull/1000. The implementation is actually identical to the rollup implementation, because we're both using the same resolve package underneath. There are some limitations I'd like to see different, but I'd rather keep both options in sync and try to fix them on rollup's side too.

TODO:
- [x] Add tests
- [x] Update docs
- [x] Figure out how to handle duplicates not in root
- [x] Figure out relative imports

### DIFF
--- a/packages/es-dev-server/src/command-line-args.js
+++ b/packages/es-dev-server/src/command-line-args.js
@@ -96,6 +96,13 @@ export const commandLineOptions = [
     description: 'Resolve bare import imports using node resolve.',
   },
   {
+    name: 'dedupe',
+    alias: 'd',
+    type: Boolean,
+    description:
+      'Dedupe modules by resolving them always from the root, ensuring only one version of a package is ever resolved.',
+  },
+  {
     name: 'preserve-symlinks',
     type: Boolean,
     description:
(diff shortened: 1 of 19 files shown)


LarsDenBakker:
This is ready for review now

thepassle:
Will take a look today

packages/es-dev-server/src/command-line-args.js
```diff
@@ -216,6 +223,11 @@ export function readCommandLineArgs(argv = process.argv, config = {}) {
 
   let { open } = options;
 
+  if ('dedupe' in options) {
+    options.dedupeModules = options.dedupe;
+    delete options.dedupe;
```
thepassle: Just out of interest, why is it needed  to delete `options.dedupe` here?

LarsDenBakker: Since I'm aliasing dedupe to dedupeModules, I want to ensure it doesn't accidentally get picked up somewhere else when the config object is passed along.

LarsDenBakker: A "cleaner" way would be to just pass a new object to go from cli args to config, but this is less code.

LarsDenBakker: A "cleaner" way would be to just pass a new object to go from cli args to config, but this is less code.

thepassle: Cool, makes sense

thepassle:
LGTM, seems to run as intended locally as well 👍

PR #1108, opened 2019-12-10
feat(classes-in-html): add classes in HTML feature

manolakis:
This PR is based on [carehtml](https://www.npmjs.com/package/carehtml), but it adds a cache because of [this](https://github.com/Polymer/lit-html/pull/274#issuecomment-420078190).

With this PR we are able to use the next syntax inside our LitElement components

```js
import { html as leHtml, LitElement } from 'lit-element';
import { wrap } from '@open-wc/lit-helpers';

const html = wrap(leHtml);

class MyButton extends LitElement {
  render() {
    return html`
      <button><slot></slot></button>
    `;
  }
}

class MyElement extends LitElement {
  render() {
    return html`
      <${MyButton} @click="${e => console.log(e)}">Click me!</${MyButton}>
    `;
  }
}
```

### DIFF
--- a/packages/lit-helpers/README.md
+++ b/packages/lit-helpers/README.md
@@ -29,7 +29,7 @@ render(
     <div
       ...=${spread({
         'my-attribute': 'foo',
-        '?my-boolean-attribute': true
+        '?my-boolean-attribute': true,
         '.myProperty': { foo: 'bar' },
         '@my-event': () => console.log('my-event fired'),
       })}
@@ -92,7 +92,7 @@ renderSpread({ foo: 'buz', lorem: 'ipsum' });
 renderSpread({ foo: 'buz' });
 
 // result: <div>
-renderSpread({ foo: undefined' });
+renderSpread({ foo: undefined });
 ```
 
 ## Live binding
--- a/packages/classes-in-html/LICENSE
+++ b/packages/classes-in-html/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 open-wc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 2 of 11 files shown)


daKmoR:
is there a specific reason to require manual wrapping like so?
```js
const html = wrap(leHtml);
```

wouldn't it be nice to just do 
```js
import { html } from '@open-wc/lit-helpers';
```

if you would need both html then you could still do it like so
```js
import { html as leHtml } from 'lit-html';
import { html } from '@open-wc/lit-helpers';
```

packages/lit-helpers/package.json
```diff
@@ -33,6 +33,8 @@
   },
   "devDependencies": {
     "@open-wc/testing": "^2.4.2",
+    "htm": "^2.2.1",
+    "preact": "^8.5.1",
```
daKmoR: why is there a devDependency on htm and preact?

what are we testing for? for what reason?

manolakis: You are right. I just retrieve the unit test created in carehtml but I'm going to remove those not related with lit-element/lit-html removing those dependencies.

manolakis:
About the wrap. the `html` function exposed by `lit-element` is different than the one exposed by `lit-html` and we must wrap the correct one depending where we want to use this syntax.

I guess the most common is going to be the one provided by 'lit-element' so we could just provide that one wrapped.

LarsDenBakker:
This is very valuable! I'd like to really dig in and test this for usability and performance before merging this in, since I want to avoid breaking change later down the road.

manolakis:
Yes, I did some performance test and obviously there is a penalty using this but IMHO we obtain more benefits than the performance we lost.

I didn't want to pollute the code with them but I could create another branch in my fork and share it to you. It would be great to have more eyes on them and probably we could improve them.

manolakis:
So... it's failing on IE11 😞 
I'm going to check why.

LarsDenBakker:
A separate branch with the performance tests would be nice

Westbrook:
I’d really like to see more documentation around this addition before I’d say it’s “ready to go”. Particularly, we should be VERY clear about the performance costs of choosing to go with something like this, so much so that it would be great to have side by sides in the documentation or CI that we can point people to when asking. After that, I think the most important thing would be some writing on why someone would use this approach over standard element definitions. From the outside, it’s convenient to neg on web components for the concepts this references, however in more than five years of custom element development I’ve never actually run into this issue in the real world. Maybe I’m not using enough community code? Maybe my lack of extended experience in other frontend techniques has me brainwashed? Having a clear “use this when” story would be helpful in clearing that up for myself as well as people less experienced with web component development. 

Separately, while this can bring benefit to the DX of a project, it would seemingly negate much of the “real HTML is good” argument behind web components at large. I’ve not spent any time with the element names it generates, but I assume from the outside that it would hash them in someway? Maybe some clarity as to the expected output when using this would be? 

Thanks in advance for helping me to better understand this addition!

packages/lit-helpers/README.md
```diff
@@ -120,3 +120,33 @@ html`
     }
   }
 </script>
+
+## Use of Custom Element classes instead tags into the DOM
```
thepassle: I would really love more documentation on this, what is it, why is it, what does it do, how does it do it

manolakis:
I just improved the docs and I published [here](https://github.com/manolakis/open-wc/commit/bb774756541e018aac7415fb394a544c21ee29dc) the performance test.

Please, feel free to comment. It would be great if we could improve them.

Thanks!

LarsDenBakker:
So it looks like we'll be going with improving https://github.com/bashmish/carehtml, thanks a lot for your work and sparking the discussions

I'd still like to look into a babel plugin which can compile out the care-html parts

LarsDenBakker:
We will still need a lit-specific wrapper, so should we create `@open-wc/lit-care` or `lit-care`?

LarsDenBakker:
I am closing this PR to keep the PR lane clean from things that can't be merged right away. We're experimenting with some different approaches in different branches, it's starting to look really good now :)

PR #1118, opened 2019-12-12
docs: fix typo in semantic-dom-diff readme

blikblum:
In "Ignoring children" section the element name in markup is different from the passed to option


LarsDenBakker:
Thanks!

PR #1119, opened 2019-12-12
docs: fix moduleDirs type definition in karma-esm

blikblum:
In karma-esm readme moduleDirs type is defined as string but is an array


LarsDenBakker:
Good catch! Could you run `yarn format`?

blikblum:
I edited directly in Github

blikblum:
This was handled as part of #1130

Issue #1127, opened 2019-12-13
[karma-esm] implement importMap support


blikblum:
I tried to pass importMap option to karma-esm as pointed in docs but seems that this feature is not implemented at all.

I looked in the code and only place that deals with import maps are  in https://github.com/open-wc/open-wc/blob/8633499fa17987a57d6c25b63dc554e693cdacd8/packages/es-dev-server/src/utils/import-maps.js#L9 which uses importMaps extracted from an index html content. So it seems that importMap option is ignored at all.

This is something i can take a look but would like some guidance at what place this code handling should go.

At first look the importMap path content should be read somewhere in karma-esm, passed as option to dev-server and injected in es-dev-server `getTransformedIndexHTML`

blikblum:
I looked at the code and the correct place to handle this is in karma-esm leaving es-dev-server untouched

LarsDenBakker:
This should be added in a responseTransformer in karma-esm, to inject into debug.html an context.html

blikblum:
Done

PR #1130, opened 2019-12-13
fix(karma-esm): Implement import map support

blikblum:
Currently importMap option passed to karma-esm is ignored (#1127)

This PR loads the importMap data and inject into karma debug and context files

It also fixes how chai path in test importmap is defined

The last part try to pass the flags to enable import map support in chrome based on [SO](https://stackoverflow.com/a/51070140). ~~This is not working~~. Using `--enable-experimental-web-platform-features` flag works

If i run Karma with Chrome and enable manually the importmap support it works

### DIFF
--- a/packages/karma-esm/demo/import-map/karma.config.js
+++ b/packages/karma-esm/demo/import-map/karma.config.js
@@ -19,7 +19,11 @@ module.exports = config => {
     customLaunchers: {
       ChromeHeadlessNoSandbox: {
         base: 'ChromeHeadless',
-        flags: ['--no-sandbox', '--disable-setuid-sandbox'],
+        flags: [
+          '--no-sandbox',
+          '--disable-setuid-sandbox',
+          '--enable-experimental-web-platform-features',
+        ],
       },
     },
 
--- a/packages/karma-esm/src/setup-es-dev-server.js
+++ b/packages/karma-esm/src/setup-es-dev-server.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const util = require('util');
 const portfinder = require('portfinder');
 const chokidar = require('chokidar');
 const fetch = require('node-fetch');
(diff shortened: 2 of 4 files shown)


packages/karma-esm/demo/import-map/karma.config.js
```diff
@@ -19,7 +19,11 @@ module.exports = config => {
     customLaunchers: {
       ChromeHeadlessNoSandbox: {
         base: 'ChromeHeadless',
-        flags: ['--no-sandbox', '--disable-setuid-sandbox'],
+        flags: [
+          '--no-sandbox',
+          '--disable-setuid-sandbox',
+          '--enable-experimental-web-platform-features',
```
LarsDenBakker: I dont think we should enable this by default for everyone, because it enabled more than just this.

Doesnt karma reuse your existing chrome indtallation? Otherwise, we should use document this step for users in their own config.

packages/karma-esm/src/setup-es-dev-server.js
```diff
@@ -7,14 +8,36 @@ const regexpKarmaLoaded = /window\.__karma__\.loaded\(\);/gm;
...
 
+/**
+ * Load importMap content from a filepath
+ * @param {string} path
+ * @returns {Promise<string>}
+ */
+async function loadImportMap(path) {
+  return new Promise((resolve, reject) => {
+    if (!path) {
+      resolve('');
```
LarsDenBakker: I prefer null or undefined over an empty string.

packages/karma-esm/src/setup-es-dev-server.js
```diff
@@ -7,14 +8,36 @@ const regexpKarmaLoaded = /window\.__karma__\.loaded\(\);/gm;
 const regexpScriptSrcGlobal = /<script type="module"[^>]*src="([^"]*)"/gm;
 const regexpScriptSrc = /<script type="module"[^>]*src="([^"]*)"/m;
 
+/**
+ * Load importMap content from a filepath
+ * @param {string} path
+ * @returns {Promise<string>}
+ */
+async function loadImportMap(path) {
+  return new Promise((resolve, reject) => {
```
LarsDenBakker: Using util.promisify for this is more idiomatic

packages/karma-esm/src/setup-es-dev-server.js
```diff
@@ -81,6 +114,7 @@ async function setupDevServer(karmaConfig, esmConfig, watch, babelConfig, karmaE
     typeof esmConfig.port === 'number' ? esmConfig.port : await portfinder.getPortPromise();
   const karmaHost = `${karmaConfig.protocol}//${karmaConfig.hostname}:${karmaConfig.port}`;
   const devServerHost = `${karmaConfig.protocol}//${karmaConfig.hostname}:${devServerPort}`;
+  const importMap = await loadImportMap(esmConfig.importMap);
```
LarsDenBakker: I think its better to check if there is an importMap property and conditionally load it, that makes it easier to see that its an optional property.

packages/karma-esm/demo/import-map/karma.config.js
blikblum: > I dont think we should enable this by default for everyone, because it enables more than just import maps.

Yes this should be opt in, documented together importMap option

> Doesnt karma reuse your existing chrome indtallation?

No. I set the flag in Chrome but when running in karma seems all flags are reset

packages/karma-esm/src/setup-es-dev-server.js
blikblum: Done

packages/karma-esm/src/setup-es-dev-server.js
blikblum: Done

packages/karma-esm/src/setup-es-dev-server.js
blikblum: Done

packages/karma-esm/demo/import-map/karma.config.js
blikblum: Added documentation

LarsDenBakker:
Great! Thanks

blikblum:
Many thanks

PR #1131, opened 2019-12-14
feat(testing-helpers): add fixture option to define wrapper el

### LINKED ISSUE #1125: [testing-helpers] Allow to inject wrapper element in fixture
Currently there's no way to customize the creation of wrapper element used as parent to the template.

A use case is testing an element that uses the dependency injection technique described by Polymer dev [here](https://www.youtube.com/watch?v=6o5zaKHedTE) and [here](https://youtu.be/x9YDQUJx2uw?t=2444). It relies on a parent element that listen to an event dispatched in the connectedCallback

An actual example of testing is [here](https://github.com/blikblum/lit-element-mobx-realworld-example-app/blob/master/test/app-header.js#L23). `app-header` element depends on a context provided by a parent element. Currently there's no way to properly test it.

This is something i can work if approved

blikblum:
Fixes #1125 

This is a quick PR to show how the feature could be implemented (mith minimal changes and backward compatible)

The option name could change from `wrapper` top `parent`


blikblum:
Renamed option from `wrapper` to `parent` since the later is shorter and easier to remember

packages/testing-helpers/src/fixtureWrapper.js
```diff
@@ -7,8 +7,7 @@ export const cachedWrappers = [];
  *
  * @returns {Element}
  */
-export function fixtureWrapper() {
-  const wrapper = document.createElement('div');
+export function fixtureWrapper(wrapper = document.createElement('div')) {
```
bennypowers: nit: for consistency, rename to `parent`

packages/testing-helpers/src/fixture-no-side-effect.js
```diff
@@ -10,14 +10,15 @@ import { isValidRenderArg } from './lib.js';
  *
  * @template {Element} T
  * @param {import('./litFixture').LitHTMLRenderable} template Either a string or lit-html TemplateResult
+ * @param {object} [options]
```
bennypowers: Please define a `FixtureOptions` typedef

packages/testing-helpers/src/stringFixture.js
```diff
@@ -7,10 +7,11 @@ import { elementUpdated } from './elementUpdated.js';
  *
  * @template {Element} T - Is an element or a node
  * @param {string} template
+ * @param {object} [options]
  * @returns {T}
  */
-export function stringFixtureSync(template) {
-  const wrapper = fixtureWrapper();
+export function stringFixtureSync(template, options = {}) {
+  const wrapper = fixtureWrapper(options.parent);
```
bennypowers: nit: Please rename here to parent

blikblum: Done

packages/testing-helpers/src/fixture-no-side-effect.js
blikblum: Done

blikblum:
@bennypowers i did the changes as your review. Is there anything else i should do?

daKmoR:
potentially open todos
- rename `parent` to `parentNode` to make it clear that it's a node
- add "nicer" jsDocs for the options to improve autocomplete docs

PS: @bennypowers is in charge of this change - he will make the change himself when he's available, you're welcome to push as well. thank you for your patience 🤗

bennypowers:
We're on the home stretch

blikblum:
Manu thanks

PR #1135, opened 2019-12-16
docs(demoing-storybook): correct import statement in MDX snippet

bennypowers:


packages/storybook-addon-web-components-knobs/demo/stories/demo-wc-card.stories.mdx
```diff
@@ -1,4 +1,4 @@
-import { Story, Preview, Meta, Props } from '@storybook/addon-docs/blocks';
+import { Story, Preview, Meta, Props } from '@open-wc/demoing-storybook';
```
daKmoR: I am afraid here the original is correct as we are not using `@open-wc/demoing-storybook` in that packages...

daKmoR:
thxxxx 🤗

PR #1136, opened 2019-12-16
Fix/character encoding

amielucha:
The lack of character encoding declaration resulted in garbled text in the Storybook navigation, and potentially in other areas of the application.

The charset meta should be included as early in possible in the document so I also moved the title tag below meta to keep all meta tags together.


LarsDenBakker:
Great! Thanks for digging into it

daKmoR:
aaahhh that is super nice 🤗 

I was annoyed by that but did not know what it was 🙈 

so simple - so elegant - thxxxxx 🤗

Issue #1091, opened 2019-12-06
chai-dom not registered in custom karma webpack setup


bschlenk:
I'm trying migrate some of my existing tests over to the `@open-wc/testing` library, mainly to use the snapshots and chai-dom.

My setup uses webpack with karma to run the tests. This seems to cause chai-dom to not be registered. It looks like [register-chai-plugins.js](https://github.com/open-wc/open-wc/blob/master/packages/testing/register-chai-plugins.js#L2) imports `chai-dom` and does nothing with it. [Chai dom is a UMD](https://github.com/nathanboktae/chai-dom/blob/master/chai-dom.js#L1-L10), so when run this way it exports `ChaiDom` instead of implicitly calling `chai.use` with it.

Is there a way to force chai-dom to be registered in this case?

LarsDenBakker:
I *think* we can make it work for both webpack and browser es modules by doing this instead:

```
// non es module plugins
import 'chai-dom';
import 'sinon-chai';

// es module plugins
import { chaiDomDiff } from '@open-wc/semantic-dom-diff';
import { chaiA11yAxe } from 'chai-a11y-axe';

let chai = window.chai;
if (!chai && require) {
  chai = require('chai');
}

// @ts-ignore
chai.use(chaiDomDiff);
// @ts-ignore
chai.use(chaiA11yAxe);
```

Could you test if this works on your setup? I don't have a webpack karma setup to test this out.

bschlenk:
That didn't work for me, but this does work:

```js
// non es module plugins
import chaiDom from 'chai-dom';
import 'sinon-chai';

// es module plugins
import { chaiDomDiff } from '@open-wc/semantic-dom-diff';
import { chaiA11yAxe } from 'chai-a11y-axe';

// @ts-ignore
chai.use(chaiDomDiff);
// @ts-ignore
chai.use(chaiA11yAxe);

chai.use(chaiDom);
```

What happens if you run `import chaiDom from 'chai-dom';` with your setup? If that returns null or an empty object, then you should be able to check that and only call `chai.use(chaiDom)` if it is an actual chai plugin.

sinon-chai would have the same problem, since it is a UMD as well.

LarsDenBakker:
Unfortunately the browser complains that there is no export from that modules, so we can't do that.

bschlenk:
It looks like the author of chai-dom is considering moving towards an es module approach: https://github.com/nathanboktae/chai-dom/issues/38#issuecomment-549221553

[...]

Issue #989, opened 2019-11-14
Importing @open-wc/testing-helpers fails with window.chai not found


ErikGrimes:
When using @openwc/testing in a Karma/Webpack/Storybook environment, importing testing-helpers fails as window.chai is undefined.  Tests where we import from chai work fine. Any thoughts?

LarsDenBakker:
This is because we import chai, which is just a plain script which registers chai to the window. Unfortuantely, webpack turns it into a module so chai isn't there on the window anymore.

I remember we had discussed this before, but I'm not sure if we came to a solution. Do you remember @daKmoR ?

Perhaps this would work:

```js
import 'chai';
let chai = window.chai;
if (!chai) {
  chai = require('chai');
}
```

PR #1142, opened 2019-12-16
fix(testing): make testing work with webpack

LarsDenBakker:
fixes https://github.com/open-wc/open-wc/issues/1091
fixes https://github.com/open-wc/open-wc/issues/989


LarsDenBakker:
Looks like it doesnt work with system js now, need to check into that

PR #1144, opened 2019-12-17
fix(create): replace webpack-merge by deepmerge in testing dependencies

blikblum:
deepmerge  is smaller and is already a dependency of karma-esm


LarsDenBakker:
Thanks, much appreciated!

PR #1159, opened 2019-12-20
fix(testing-karma): removes snapshot filename processor

benjamind:
Currently the `testing-karma` configuration uses a `snapshot-filename-preprocessor` to rewrite the `.md` files generated by `karma-snapshot` into `.js` files so that `karma` does not emit warnings about the filetype on the console.

Karma will automatically try to deduce filetype from the extension of the file, but you can specify the filetype in the `files` configuration using the `types` property. This eliminates the need to rewrite the filenames since karma will no longer complain.

The browser also does not care about the extension of the file and will assume the content is JS since it gets included into the page in a script tag with a `type="text/javascript"` attribute.


LarsDenBakker:
Nice find, thanks!

Issue #1151, opened 2019-12-19
`storiesPattern` doesn't need be prefixed by `rootDir`


jorgecasar:
I notice that `storyUrls` are listed using the glob `storiesPattern` using `rootDir` as current working directory, then it's unnecessary prefixing it.

Without prefixing `rootDir` we have more flexibility with `stories` config flag (very useful in monorepos)

https://github.com/open-wc/open-wc/blob/bb3875dd7e356a2ca6b723d89b71d78ed3827ce7/packages/demoing-storybook/src/start/cli.js#L23-L24

LarsDenBakker:
I merged https://github.com/open-wc/open-wc/pull/1152 but I'm not sure if it fixes your issue. A user would expect that the stories pattern resolves relative to your current working directory, not the root dir. Especially in monorepositories. With this change, you'd still be resolving from the root dir I think.

@daKmoR should we change this to resolve relative to the working directory?

jorgecasar:
Yes, I tested locally and works in rootDir and monorepository. But It could be great if someone else can test it.

Thanks!

LarsDenBakker:
Closing this as we settled on a solution in https://github.com/open-wc/open-wc/pull/1160

PR #1160, opened 2019-12-21
fix(demoing-storybook): fixes story file listing

benjamind:
Changes in #1152 modified how file listing is done. This does however not capture all cases correctly.

Since the glob path passed to `listFiles` no longer has `rootDir` prepended to it, and since `glob` itself does not fully resolve the paths it returns relative to the `cwd` option, the paths are no longer correctly resolving in all cases.

This PR resolves each path against the root dir, before doing the directory check.

### DIFF
--- a/packages/demoing-storybook/src/shared/listFiles.js
+++ b/packages/demoing-storybook/src/shared/listFiles.js
@@ -1,10 +1,23 @@
 const glob = require('glob');
 const fs = require('fs');
+const path = require('path');
 
+/**
+ * Lists all files using the specified glob, starting from the given root directory.
+ *
+ * Will return all matching file paths fully resolved.
+ */
 module.exports = function listFiles(fromGlob, rootDir) {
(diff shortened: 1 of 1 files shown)


benjamind:
May also be related to #1151

LarsDenBakker:
Thanks for the PR. One thing I'm unsure about is whether we should be resolving relative too the root dir, or relative to the current working directory. What's the user expectation when using a command line tool?

benjamind:
As I understand it right now, `rootDir` defaults to `cwd` unless the user explicitly sets it in the es-dev-server config. So the semantics of the command line arg in the general case are to resolve relative to cwd as a command line user might expect, unless rootDir is set then it's resolved relative to the rootDir which itself is resolved relative to cwd.

So I think these semantics make sense for now.

Though I think the combination of different config files, and how they are extracted to the various tools demoing storybook is using is starting to get quite complex.

PR #1161, opened 2019-12-21
fix(node-resolve): support packages with 'jsnext:main' hint

cristinecula:
There are still packages that have not been updated to use the "module" attribute and still use the "jsnext:main" attribute. es-dev-server v1.32.0 has broken the support for these packages.

For example: https://github.com/moment/moment/blob/master/package.json


LarsDenBakker:
Thanks for your PR. I am in the process of moving all our node resolve logic to use the rollup node resolve plugin within es-dev-server. This way we're always in sync with the logic of rollup. They support jsnext:main but are the only build tool to do so, I know they're planning to remove this soon as well.

cristinecula:
Will you expose the option to control the main fields?

LarsDenBakker:
Yes, it accepts all the rollup node resolve options. Closing this one in favor of https://github.com/open-wc/open-wc/pull/1163

PR #1169, opened 2019-12-26
feat(scoped-elements): add scoped elements feature

manolakis:
This feature allows to define custom elements, scoping them when necessary, allowing to use different versions of the same component without naming collisions.

```js
import { css, LitElement } from 'lit-element';
import { createScopedHtml } from '@open-wc/scoped-elements';
import MyPanel from './MyPanel.js';

const html = createScopedHtml({
  'my-button': MyButton,
  'my-panel': MyPanel,
});

export default class MyElement extends LitElement {
  static get styles() {
    return css`
      .my-panel {
        padding: 10px;
        background-color: grey;
      }
    `;
  }

  static get properties() {
    return {
      text: String,
    };
  }

  render() {
    return html`
      <my-panel class="my-panel">
        <my-button>${this.text}</my-button>
      </my-panel}>
    `;
  }
}
```


packages/scoped-elements/src/scoped-elements.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const registerElements = (elements, myRegistry) => Object.keys(elements)
+    .reduce((acc, key) => {
+      const klass = elements[key];
+      acc[key] = getRegisteredTag(klass) || defineElement(myRegistry, klass);
+
+      return acc;
+    }, {});
+
+export const getTags = (elements, myRegistry) =>
```
LarsDenBakker: I'm not sure if we should add registry composition, what's the use case?

packages/scoped-elements/src/tag.js
```diff
@@ -0,0 +1,60 @@
...
+  const name = klass && klass.name && klass.name.replace(NAME_REGEX  , '');
+
+  if (name) {
+    let tag = toDashCase(name);
+
+    if (tag.indexOf('-') === -1) {
+      tag = `c-${tag}`;
+    }
+
+    if (isTagRegistered(registry, tag)) {
```
LarsDenBakker: Does this mean we create a new tag for each class? I'm not sure if it's a problem, but perhaps we should cache the tagname we created for each class?

packages/scoped-elements/src/scoped-elements.js
manolakis: ups! It's from a test I did about how could it be used with scoped registries, but I forgot to remove it.

manolakis: I'm going to remove it

packages/scoped-elements/src/tag.js
manolakis: This checks that the created tag is not used by any other component, incrementing it if that's the case.

created tags are also cached inside packages/scoped-elements/src/scoped-elements.js#3 (registeredElements is a Map between classes and created tags).

packages/scoped-elements/src/scoped-elements.js
```diff
@@ -0,0 +1,28 @@
+import { createUniqueTag } from "./tag.js";
+
+const registeredElements = new Map();
```
LarsDenBakker: This library will very likely be loaded multiple times per runtime, which means there is no single Map and we will end up with multiple elements defined per class. Do we think that's a problem? Just some food for thought. We could create a `Map` on the window if we really want to share it.

manolakis: Even if it's loaded multiple times it's going to be evaluated just once (http://www.ecma-international.org/ecma-262/6.0/#sec-moduleevaluation). However, I didn't think that different versions could coexist at the same time by different application parts, so at the end it would be evaluated several times (once per version used).

Good point! Thanks @LarsDenBakker

LarsDenBakker: People will use this in a micro front end setup, where separate parts are bundled and dependencies duplicated. So even the same version may end up on the page multiple times.

Issue #991, opened 2019-11-17
[demoing-storybook] add nice ordering by default 


daKmoR:
probably something in the line off

https://gist.github.com/JohnAlbin/83ef5282d07c6aa447ff39719429843e

PR #1171, opened 2019-12-27
feat(demoing-storybook): inject ordered exports into stories

LarsDenBakker:
fixes https://github.com/open-wc/open-wc/issues/991

This adds the new `__orderedExports` option from storybook. A few things to note:
- We still need to update storybook-prebuilt to test this out properly
- Older versions of storybook break when there is an `__orderedExports` because it tries to execute each export as a function. Is this a bug on the storybook side? We will need to require everyone to update to the latest of everything.
- If possible, `__orderedExports` should actually be named `__orderedNamedExports` because we shouldn't include the default export in there.
- The current approach is to export an object of the exportes stories. This creates problems when you export aliases:

```js
const myStory = () => html`Foo`;

export { myStory as notMyStory };
```

This will result in:

```js
const myStory = () => html`Foo`;

export { myStory as notMyStory };

export __orderedExports = { notMyStory };
```

which will throw because `notMyStory` is not a variable. We can't use `myStory` either because that's not the name of the export. We need both the variable name and the exported name.

My proposal is to use an array of exported names, and then access the actual stories on the module object. This will be the simplest implementation.

```js
const myStory = () => html`Foo`;

export { myStory as notMyStory };

export __namedExportsOrder = ['notMyStory'];
```


LarsDenBakker:
@daKmoR what do you think about changing to an array of exported names, and then accessing them on the module object?

daKmoR:
yes I agree if we want to support re-exports then an array of strings is the way to go.

I am pretty sure it's fine with the storybook guys as we discussed both solutions (string and variables) and we then implemented the variables as it was easier... but yeah re-exports have not been considered 🙈

so let's first do a PR for storybook - best pretty soon as it's in 5.3 rc phase

packages/demoing-storybook/src/build/build.js
```diff
@@ -25,12 +54,11 @@ async function buildPreview(outputDir, previewPath, assets) {
...
   };
 
-  const configs = createCompatibilityConfig({
-    input: 'noop',
-    outputDir,
-    extensions: [...DEFAULT_EXTENSIONS, 'mdx'],
-    plugins: { indexHTML: false },
-  });
+  configs[0].treeshake = false;
+  configs[1].treeshake = false;
```
LarsDenBakker: we should remove these

daKmoR:
this now adds an array of strings (names of the exports in order)
```js
export const __namedExportsOrder = ['notMyStory'];
```

PR #1180, opened 2019-12-31
fix(demoing-storybook): serve sourcemap

LarsDenBakker:
Serve storybook manager sourcemap, the preview sourcemap is fixed in the storybook prebuilt


daKmoR:
can you fix the conflict? 🙏  🤗

PR #1185, opened 2020-01-02
feat(scoped-elements): add scoped elements feature

manolakis:
This feature allows to define custom elements, scoping them when necessary, allowing to use different versions of the same component without naming collisions.

```js
import { css, LitElement } from 'lit-element';
import { createScopedHtml } from '@open-wc/scoped-elements';
import MyPanel from './MyPanel.js';

const { html } = createScopedHtml({
  'my-button': MyButton,
  'my-panel': MyPanel,
});

export default class MyElement extends LitElement {
  static get styles() {
    return css`
      .my-panel {
        padding: 10px;
        background-color: grey;
      }
    `;
  }

  static get properties() {
    return {
      text: String,
    };
  }

  render() {
    return html`
      <my-panel class="my-panel">
        <my-button>${this.text}</my-button>
      </my-panel}>
    `;
  }
}
```


packages/scoped-elements/README.md
```diff
@@ -0,0 +1,210 @@
...
+the same components several times, but as those libraries evolve name collision
+problems between different versions of the same library appear because teams are
+not able to evolve their code at the same velocity. That causes bottlenecks in
+software delivery that should be managed by the teams and complex build systems
+trying to alleviate the problem.
+
+[Scoped Custom Element Registries](https://github.com/w3c/webcomponents/issues/716)
+is a proposal that tries to fix that problem, but until is ready and we could
+use a polyfill in those browsers that do not implement it, auto scoping the
+custom elements is a workaround.
```
thepassle: Could you rewrite this paragraph? Its not very clear

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+This package allows you to forget about how the custom elements are defined
+inside the registry. It's going to register them for you, scoping the name if
+necessary avoiding the class name collision, and allowing you to use different
```
thepassle:
```suggestion
necessary avoiding the class name collision, and allows you to use different
```

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+## How it works
+
+`createScopedHtml` is going to define the custom element classes and creates a
+unique scoped tag for each of them if needed. In case those classes were
+previously defined then it's going to return the previous defined tag.
+
+Then, the `html` function provided by the `createScopedHtml` method is going to
+it's going to transform the template literal into another one replacing the tags
```
thepassle: is going to it's going to? Can you rewrite this?

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,210 @@
...
+    return html`
+      <my-button>Click me!</my-button>
+    `;
+  }
+}
+```
+
+## When not to use it
+
+In the case you are developing a custom element that doesn't depends on other
```
thepassle:
```suggestion
If you are developing a custom element that doesn't depend on other
```

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,210 @@
...
+custom elements then you shouldn't use it. The reason of that is because there
```
thepassle:
```suggestion
custom elements then you shouldn't use it. The reason for that is because there
```

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,210 @@
...
+is a little performance penalty on using it, and it's better to avoid it if you
```
thepassle:
```suggestion
is a small performance penalty for using it, and it's better to avoid it if you
```

packages/scoped-elements/src/scoped-elements.js
```diff
@@ -0,0 +1,28 @@
...
+import { tagsCache } from './context.js';
+
+const registerElement = (klass, tag) => tagsCache.set(klass, tag);
+const getRegisteredTag = klass => tagsCache.get(klass);
+
+const defineElement = klass => {
+  const registry = customElements;
+  const tag = createUniqueTag(registry, klass);
+
+  // we extends it just in case the class has been defined manually
```
thepassle:
```suggestion
  // we extend it just in case the class has been defined manually
```

packages/scoped-elements/src/tag.js
```diff
@@ -0,0 +1,60 @@
+function* counterFunc() {
```
thepassle: Does this really need to be a generator function? It seems a little overkill to increment a counter?

manolakis: Yes, you are right. I'm going to change it

manolakis:
@thepassle thanks for your comments

packages/scoped-elements/README.md
manolakis: Check out if it's clear now

packages/scoped-elements/README.md
manolakis: I just rewrited it

packages/scoped-elements/src/context.js
```diff
@@ -0,0 +1,16 @@
+// @ts-nocheck
+if (!window.ScopedElements) {
```
daKmoR: why is this on window?

do we want to support duplicate versions of `scoped-elements`? I don't think that would be a good idea 🙈 

so if `scoped-elements` needs to be unique then this can be a local variables rigth?

LarsDenBakker: There will be almost guaranteed multiple versions of `scoped-elements` on a page. One for each bundle.

manolakis: @LarsDenBakker pointed correctly that the file could be evaluated several times in case of:
1. An application based on micro frontends with separate builds. In that case each build is going to have it's own copy of the file inside it's bundle.
2. Different versions of `@open-wc/scoped-elements` package while not all teams upgrade to the last version (this case is less probable, but possible).

daKmoR: fair enough - so it is by "design" like that...

then it should definitely be a more unique name - maybe `window.__scopedElements`?

and I guess it would be smart to add the major version number as well e.g. so we can later do backwards compatibility checks ala

```js
let scopedElements = window.__scopedElements
if (window.__scopedElements.majorVersion === 0) {
  scopedElements = upgradeIfNeeded(window.__scopedElements);
}

// then always use scopedElements
```

manolakis: If the file is evaluated several times then the cache is going to be splitted into different caches and elements are going to be registered several times with different tag names. Probably is not a big issue but not desirable, don't you think?

manolakis: On the other hand, with local variables is going to work also and code is cleaner...

LarsDenBakker: Yeah I'm unsure whether it's a problem or not, not sure if a large custom element registry is a problem? We could also start without a global variable and change it later if needed?

manolakis: Simpler is better. We could refactor it later.

manolakis: fixed it

tmsns:
Cool idea to support scoped elements in this way. 👍 😎

Some things that popped up when looking into the PR:
- I don't think styles are matching when unique tags are generated. You could have a custom `css` function, though.
- Do we want to support `unsafeHTML` directive within the scoped `html` function? For now, it will just fallback to the regular tag creation.
- Is it clear (enough) for the user that that some regular functionality will be behaving unusually? eg. using `querySelector`. Most of this functionality is of course related to the unique tag name, but still.

And maybe a more general , open, question: is the pattern of using scoped elements to enable multiple versions of a component something that should be used in / promoted for live web sites? 🙈  I think it makes sense in controlled environments like test cases where one need to test multiple extensions of a component.
But in live web sites, it's often a slippery slope:
- inconsistent design and behavior within the page / app
- pages / apps become larger / slower as they need to load x amount of code
Also, the usage of multiple versions components can be tricky (eg. when using packages to release the components, only some package manager can handle multi versions (ie. npm, and not bower))

What do you guys think? 😊

manolakis:
Hello @tmsns,

> I don't think styles are matching when unique tags are generated. You could have a custom css function, though.

Yes, you are right. It's possible to create a custom css function also. However we thought have out of scope the css, but we could add it later in another iteration. By the moment classnames could be used instead tagnames in order to style the child components in case you need it.

> Do we want to support unsafeHTML directive within the scoped html function? For now, it will just fallback to the regular tag creation.

I'm not sure if I'm understanding your question. If your question is how to use a scoped tag inside the unsafeHTML you can use the `getTags` function.

```js
import { getTags } from '@open-wc/scoped-elements';
import { unsafeHTML } from 'lit-html/directives/unsafe-html.js';
import { MyButton }

const [myButtonTag] = getTags([MyButton]);

const markup = `<${myButtonTag}>Some button to click.</${myButtonTag}>`;
const template = html`
  Look out, potentially unsafe HTML ahead:
  ${unsafeHTML(markup)}
`;
```

I'm not sure if I answer your question.

> Is it clear (enough) for the user that that some regular functionality will be behaving unusually? eg. using querySelector. Most of this functionality is of course related to the unique tag name, but still.

If you want to use `querySelector` you must use the previous `getTags` function inside your component, but the doc must be clear about this and I forget to add it. Thanks!

>is the pattern of using scoped elements to enable multiple versions of a component something that should be used in / promoted for live web sites? 🙈 I think it makes sense in controlled environments like test cases where one need to test multiple extensions of a component.
But in live web sites, it's often a slippery slope:
>
>  -  inconsistent design and behavior within the page / app
    pages / apps become larger / slower as they need to load x amount of code
>  -  Also, the usage of multiple versions components can be tricky (eg. when using packages to release the components, only some package manager can handle multi versions (ie. npm, and not bower))

This feature is going to be used mostly in corporate applications in which several teams (that probably are also distributed across several locations) are involved in the development. In this scenario, when there are shared component libraries that must be used by the teams (for example to maintain a corporate brand), when those libraries are updated not all teams have the capability to update their code at the same velocity, blocking the shipping of new features until all teams update their code to the same components library version. This feature allows the shipping of new features while eventually different versions of the same component are used.

I hope I have clarified your doubts, and thank you for your comments :)

tmsns:
> Yes, you are right. It's possible to create a custom css function also. However we thought to have out of scope the css part by the moment, but we could add it later in another iteration. By the moment classnames could be used instead tagnames in order to style the child components in case you need it.

True, and by reading your comment I rechecked the `readme.md` file. I did not catch the `.` in the current css example.  🙈 

> I'm not sure if I'm understanding your question. If your question is how to use a scoped tag inside the unsafeHTML you can use the `getTags` function.
> 
> ```js
> import { getTags } from '@open-wc/scoped-elements';
> import { unsafeHTML } from 'lit-html/directives/unsafe-html.js';
> import { MyButton }
> 
> const [myButtonTag] = getTags([MyButton]);
> 
> const markup = `<${myButtonTag}>Some button to click.</${myButtonTag}>`;
> const template = html`
>   Look out, potentially unsafe HTML ahead:
>   ${unsafeHTML(markup)}
> `;
> ```

That will indeed work.   
I was thinking more in the line of creating a drop-in for `unsafeHTML` as well. So basically:
```javascript
import {unsafeHTML} from '@open-wc/scoped-elements/directives/unsafe-html.js';
```

In the end, it's the same remark as the `css` function.

> If you want to use `querySelector` you must use the previous `getTags` function inside your component, but the doc must be clear about this and I forget to add it. Thanks!

In the end, as you are probably well aware, any usage of the regular tag should be done in a different way. This might be a deal-breaker for some users. But as long as they get a fair warning upfront, I think we should be good. 😃 

> This feature is going to be used mostly in corporate applications in which several teams (that probably are also distributed across several locations) are involved in the development. In this scenario, when there are shared component libraries that must be used by the teams (for example to maintain a corporate brand), when those libraries are updated not all teams have the capability to update their code at the same velocity, blocking the shipping of new features until all teams update their code to the same components library version. This feature allows the shipping of new features while eventually different versions of the same component are used.

I see what the use case is. But, they need to be very careful with this type of usage. I see a number of pitfalls:
- sub components  
When a component will use this technique, it will indeed allow multiple versions of the same component to be on the page. However, any sub component it uses should also use the same strategy. eg:
    ```javascript
    <my-component-v1>
    #shadowRoot
      <my-subcomponent></my-subcomponent> // v3
    </my-component-v1>
    <my-component-v2>
    #shadowRoot
      <my-subcomponent></my-subcomponent> //v4
    </my-component-v2>
    ```
    In this example, you can run into trouble if v2 has been created with another (breaking) version (v4) of `my-subcomponent`, it will be broken as v3 was loaded. This means to make the scoping work consistently, any used component within should use the same pattern.  
Very often, this is a problem, as companies tend to reuse other libraries as well (ie. lion), that don't use this feature.

- developer experience  
This is related to the previous points. For regular developers, they will need to take great care to always use the scoped element export, and to always map the tag name before using it. Next to the fact that this is "some" burdon, sometimes it will be impossible to use when using external libraries. 

- inconsistent design and larger size of sites  
This is what I mentioned before. In the company I work for we had the exact same problem. There was a UI library. But every team started to use a different version. They didn't have the time nor the budget to keep up to date, which resulted in a degraded experience for their customers.  
Web components force their users to use the same component, and to stay in line. With a good CI/CD, it is possible to have a great velocity.  💪 
Btw, if the size get so large of the bundle, there is actually a web feature that allows multiple multi versioning of shared components per page, and that is to leave the SPA and go for an actual page reload. 🙈 This is also what was used in the company. But the inconsistent design and larger size became a problem.

Now, don't get me wrong. I love the idea! ❤️ 
I have been following a similar initiative by @justinfagnani: https://github.com/w3c/webcomponents/issues/716, where point 1 and point 2 will be tackled by. I think scoping can be very useful. I'm using a similar approach when unit testing mixins. I'm just not sure that it is a silver bullet in every case, because of point 3. So I think it's recommendation should be done very carefully.

> I hope I have clarified your doubts, and thank you for your comments :)

Yup, thanks for taking the time! 😃

manolakis:
@tmsns you are absolutely right.

IMHO I think the problem with custom elements is that it is difficult to follow a micro frontend strategy with them. Most JS frameworks give you more flexibility, although you must be careful with the size of the application and the inconsistent design. My point here is that these things should be handled by the product strategy, and tools like a good CI / CD help you a lot as you mentioned. However, the difference is that you have the flexibility of having possible inconsistencies in the design or size of the application favoring the delivery of value to the user.

I also believe that the auto definition is wrong. A CE shouldn't tell you which tag you have to use. You should be able to import a CE and decide how to use it in your component. Otherwise, you may even have problems if two different libraries decide to use the same prefix for their components.

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+### What if I need to use the component tag?
+
+Sometimes we need to use the component tag name, for example, if we want to do a
+`querySelector`. In that case there is a `getTags` function available that receives
+a collection of component classes and returns a collection with the tag names
+used in their definitions.
+
+```js
+import { getTags } from '@open-wc/scoped-elements';
```
LarsDenBakker: If we want to support compiling away scoped elements, we need to think how this will interact with it

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,232 @@
...
+import { ComponentA } from './ComponentA.js';
+import { ComponentB } from './ComponentB.js';
+
+const [componentA, componentB] = getTags([ComponentA, ComponentB]);
```
daKmoR: I am not sure if we wanna expose getTags 🤔 

what happens if we use both?
```js
const html = createScopedHtml({
  'my-button': MyButton, // <-- bind the element with a tag in your html
  'my-panel': MyPanel,
});

const [tagMyButton, tagMyPanel] = getTags([MyButton, MyPanel]);
```

are we sure that it is always the exact same thing? and even if... the API seems confusing to me 🙈 

if we wanna expose it - how about
```js
const { html, tags } = createScopedHtml({
  'my-button': MyButton, // <-- bind the element with a tag in your html
  'my-panel': MyPanel,
});

document.querySelector(tags['my-button']);
```

manolakis: Yes, it´s the same thing. I think we must expose the tags in some way but I like your proposal using `createdScopedHtml`.

I´m going to do the code changes.

packages/scoped-elements/README.md
frank-dspeed: compiling at all is not possible with this as it depends on customElements registry to look if it has a registered element for this so all this will only work in the browser if your bundler shims complet browser environment including customElement registry then sure you can :)

frank-dspeed: to compile this away you would need something like its getUniqTagName function on the bundler side that would need to inspect the whole JS ast and do the replacements this can be done via webpack or rollup for example if you do it in the bundler this is not needed

this works only at runtime and is not bundle able or lets say compile away able.

frank-dspeed: i have my own framework called tag-html i covered near all expectable scenarios about ssr and lit-html and any other webcomponent implamentations.

LarsDenBakker: You can compile away the runtime regexp.

packages/scoped-elements/README.md
frank-dspeed: tags don't need to get exposed in a complicated way you can simply return from createScopedHtml the WeakMap or in your case map and get the ElementName via the Class. also you don't need to expose the original class names in your config.

LarsDenBakker: Can we hold off on this for the first release? We need to see how this works with the build plugin.

frank-dspeed:
here are the relevant parts of my implamentation in my implamentation you can always use registerUniqElement(OriginalDefinitionClass) => uniq key

```js
*** problems with both versions *** 
is that if SSR we would need to shim the customElements api to make sure
that a element don't exists twice and the code need to be fully evaluated
so the uniq feature only works in the browser and only after all elements have registered it is only useable for widgets

https://github.com/open-wc/open-wc/pull/1185 watch this pr if merged apply our code

*/
// Register here in a WeakMap is a good idea for memory optimization as we don't need to get Elements that arn't there
export const registeredElements = new WeakMap();
/* note that registerUniqElement(definition) returns the key*/

const toSneakCase = name => ((name.indexOf('-') === -1) ? `c-${tagName}` : name).toLowerCase();
const incrementTagName = (tagName, counter=1) => {
  const newName = `${tagName}-${counter}`;
  return customElements.get(customElements,newName) ? newName : incrementTagName(tagName, counter + 1);
}
export const registerElement = (tagName, klass, extend  ) => {
    customElements.define(tagName, klass,extend);
    registeredElements.set(klass, tagName);
    return tagName;
}
/*<${registerUniqElement(MyButton)}><${registerUniqElement(MyIcon)} icon="search"></${registerUniqElement(MyIcon)}>Search</${registerUniqElement(MyButton)}>*/
// get a uniq tagName for our customElement Definition and register it in customElements.
// returns a tagName as string for the registrated element
function registerUniqElement(klassName,klass,extend) {
    if (!klass && typeof klassName !== 'string') { klass = klassName }
    else { /* TypeError */ }
  
    /* while this allows us to provide fastpath it could lead to unneeded weakmaps
     in some import scenarios that are often used but the impact is small */
    const existing = registeredElements.get(klass)
    if (existing) { return existing };

    const tagNameName = typeof klassName === 'string' ? klassName : Object.prototype.hasOwnProperty.call(klass, 'name') ? toSneakCase(klass.name) : 'c';
    const uniqueTagName = customElements.get(customElements, tagNameName)
        ? (customElements.get(tagNameName) === klass)
            ? tagNameName
            : incrementTagName(tagNameName)
        : tagNameName;
    
    return registerElement(uniqueTagName, klass,extend)
}

//scope = { 'my-button': MyButton, 'my-panel': MyPanel, }
/* <my-button params="frank">Super hero</my-button> */
export const createScopedHtml = scope => { 
    const replacElementTags = function (strArr, ...argsArr) {       
        Object.keys(scope).map(sk => registeredElements.set(scope[sk], registerUniqElement(sk, scope[sk])))
        strArr.map(str => Object.keys(scope).map(sk => str
            .replace(/<\/sk>/g, `</${registeredElements.get(scope[sk])}>`)
            .replace(/<sk /g, ` <${registeredElements.get(scope[sk])}`)
            .replace(/<sk> /g, `<${registeredElements.get(scope[sk])}`)))
        //Todo Better Regex
        // This array is save it holds refs to customElments registry it self which is immutable!

    
        return [strArr,...argsArr]
    }
}
```

packages/scoped-elements/README.md
daKmoR: agree let's hold of with that for now but make it an object so we can later add "something" if needed?

```js
const { html } = createScopedHtml(...)
```

manolakis: I already did the modification returning the tags in the object. Could we hold it with that or you prefer a rollback to the previous one?

LarsDenBakker: I think we should go with `const html = createScopedHtml(...)` to start off, and evaluate if the other things are needed later and when we know if and how we want to deal with compiling parts of this away. The listed use cases in the docs can be solved with css classes, ids or attributes.

daKmoR: indeed...

manolakis: Yes, that's true. I'm going to rollback to that version without the `getTags` function.

LarsDenBakker: Sorry I dont mean to devaluate your work, I want nothing more than this feature to succeed !

manolakis: No worries, I understand you. And it´s true that users can use css classes, ids or attributes to cover that functionality. Let´s see if it´s really needed in the future. I really appreciate your comments.

The less code the better.

manolakis: I just updated the PR. Let's see if it's ok as a start point now.

daKmoR:
ok, documentation should now be ready 💪  I put it under a new main element "Experiments > Scoped Elements"
please review 🤗 

Docs:
https://deploy-preview-1185--open-wc-org.netlify.com/scoped-elements/#installation

3 Demos:
- Before nesting: https://deploy-preview-1185--open-wc-org.netlify.com/scoped-
elements/demo/before-nesting/
- Fails after nesting: https://deploy-preview-1185--open-wc-org.netlify.com/scoped-elements/demo/no-scope/
- Success with scopes: https://deploy-preview-1185--open-wc-org.netlify.com/scoped-elements/demo/with-scope/

the links in the readme will only work AFTER it's merged to master, so that's why I posted the links here.

I think this is really powerful - but it needs to be sold well and we need to make sure it's a good starting point for a discussion about scoping in self-registration.

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,276 @@
...
+    return html`
+      <my-panel class="panel">
+        <my-button>${this.text}</my-button>
+      </my-panel}>
+    `;
+  }
+}
+```
+
+## Limitations
```
tmsns: Although it might be expected, I would add as a limitation: 
- that all dependencies should be NPM dependencies
- that all references should use node resolution

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+   only side effects free class exports may be used
+
+   ```js
+   export class MyEl { ... }
+   ```
+
+   Note: not share or components withing the main app shell are exempted from this rule.
+
+2. Every (!!!) component with sub component needs to use `scoped-elements`
```
tmsns: As this is an important limitation, it might be useful to add a reference to the use case?  
Specifically, Team Black should implement or should have implemented `scoped-elements` for both version 1 as version 2 of Feature A when using custom elements internally. (as well as any other feature that team exposes, ie Feature B)

Also, not sure if it is a separate limitation, but Feature A and Feature B should not use any global dependencies, only real ES modules.  
ie. if Feature A uses `DOMPurify`, and updates to a breaking version of `DOMPurify` in Feature A v2, it wil be breaking, as `DOMPurify` is installed in the global namespace. (module wrappers might not always do the trick)

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+3. [with-scope](https://open-wc.org/scoped-elements/demo/with-scoping/) [[code](https://github.com/open-wc/open-wc/tree/master/packages/scoped-elements/demo/with-scope)] Successfully "fixing" the demo by using `createScopedHtml` on Page A/B.
+
+## How it works
+
+`createScopedHtml` is going to define the custom element classes and creates a
+unique scoped tag for each of them if needed. In case those classes were
+previously defined then it's going to return the previous defined tag.
+
+Then, the `html` function provided by the `createScopedHtml` method is going to
```
thepassle: This really needs a better explanation. Can someone take a look at this? @manolakis @daKmoR @LarsDenBakker

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,276 @@
...
+   export class MyEl { ... }
+   ```
+
+   only side effects free class exports may be used
+
+   ```js
+   export class MyEl { ... }
+   ```
+
+   Note: not share or components withing the main app shell are exempted from this rule.
```
thepassle: What does this mean? "Not share or components" ? Can we rewrite this?

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+[//]: # 'AUTO INSERT HEADER PREPUBLISH'
+
+:::warning
+This is an experimental feature use it at your own risk and be sure to understand it's [limitations](#limitations).
+No bigger applications is using `scoped-elements` - so there is no prove yet if it works in production.
+This page focuses a lot on in depth explanations as it should help foster a discussion on scoping.
+:::
+
+Complex Web Component applications often are developed by several teams across organizations.
```
thepassle: as an aside, why does every line have a 'manual' line break? Why not just let the paragraph run?

packages/scoped-elements/README.md
frank-dspeed: It should mean nothing as it can be total ignored he was trying to point out that if there is already a component defined with a tagname then this will try to register it also which has no negativ impact at all.

packages/scoped-elements/README.md
frank-dspeed: Global Scope is not handled at all with this this is only about registration of customElements under a uniq tag maybe uniqElement Tag is even a better name thats how i called it in my framework

packages/scoped-elements/README.md
thepassle: Then it should be removed

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,276 @@
...
+│                 │            │                     │         0% - 8% │           -        │
+│                 │            │                     │ 0.13ms - 9.60ms │                    │
+└─────────────────┴────────────┴─────────────────────┴─────────────────┴────────────────────┘
+```
+
+## Special thanks
+
+This package was initially inspired by [carehtml](https://github.com/bashmish/carehtml) and I would like to
+thank [@bashmish](https://github.com/bashmish) for his work on it.
+Also I would thank [@daKmoR](https://github.com/daKmoR) and [@LarsDenBakker](https://github.com/LarsDenBakker)
```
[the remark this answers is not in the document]
frank-dspeed: hahahah mixing kanye west in is a nice joke from you :D

packages/scoped-elements/README.md
tmsns: You're right.  😃 
But unfortunately, it's not only about the registration of customElements. As explained in the documentation, sub node_modules will be used to allow multiple versions. This means that if the breaking change (of Feature A) is a breaking change in a sub dependency (eg.`DOMPurify`), now, all of a sudden, both versions of `DOMPurify` will also be (tried to be) loaded in the same way. This will often lead to unexpected issues (global variables being overriden, global events being hit multiple times, ...), where before the loading of the global variable was only done once.  
Now, as long as all dependencies are real ES modules with no outside dependencies, it *should* not have unexpected consequences. So, this is more about a consequence of using multiple versions.

packages/scoped-elements/README.md
manolakis: It's because my IDE setup. I could change it if it's a problem.

packages/scoped-elements/README.md
manolakis: How about

Note: not shared components or those withing the main app shell are exempted from this rule.

packages/scoped-elements/README.md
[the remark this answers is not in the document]
manolakis: Ok, fair enough.

packages/scoped-elements/README.md
manolakis: I´m resolving it because it's outdated

packages/scoped-elements/README.md
manolakis: I applied @thepassle suggestion with a better explanation.

packages/scoped-elements/README.md
daKmoR: hmm I do not think it matters where you get your dependencies from?
as soon as it is double-loaded (be it nested npm or same file via different imports or a micro architecture with some double loading)

packages/scoped-elements/README.md
thepassle: What exactly do you mean with shared components? Isnt that the point of this library?

packages/scoped-elements/README.md
tmsns: I think it depends. 🙈 You are right by saying it should be double-loaded (or multi-loaded for that matter). If you use Bower to package the components, you won't be able to multi-load. But it's true that within a project it doesn't matter. 😃   
When I mentioned "dependencies", I was talking about packaging strategy, especially valid for the multi teams example.

packages/scoped-elements/README.md
manolakis: mmm probably it's a bit confusing. Maybe "Those components that are not going to be shared in a library are exempted from this rule."

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,261 @@
...
+3. **Team Blue** (on **Page A**) does not use any of those new use cases and they have a tight deadline to meet so they can not update right now.
+4. **Team Green** (on **Page B**) has to deliver an important functionality to your end users but they need to upgrade to **Feature B 2.x** since it can only be solved with this new version.
+5. Since **Feature A 1.x & 2.x** are both used in the same app, this will lead to nested dependencies, which then will lead to [catastrophic failure, and errors](https://open-wc.org/scoped-elements./demo/no-scoping/) [[code](https://github.com/open-wc/open-wc/tree/master/packages/scoped-elements/demo/no-scope)].
+
+Two possible solutions come to mind:
+
+2.  Temporarily (!) allow to ship similar source code (most breaking releases are not a total rewrite) and scope them via `@open-wc/scoped-elements`; see the "fixed" example [with-scope](https://open-wc.org/scoped-elements/demo/with-scoping/) [[code](https://github.com/open-wc/open-wc/tree/master/packages/scoped-elements/demo/with-scope)] running with nested dependencies.
```
manolakis: ops! It should start by the number 1 ;)

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,261 @@
...
+   export class MyEl { ... }
+   ```
+
+   Only side effects free class exports may be used
+
+   ```js
+   export class MyEl { ... }
+   ```
+
+   Note: not share or components within the main app shell are exempted from this rule.
```
thepassle:
```suggestion
   Note: This rule does not apply to components that you write, and are in control of yourself and do not come from a third party library.
```

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+2. Every (!!!) component that uses sub components need to use `scoped-elements`. Any import to a self registering component can result in a browser exception - completely breaking the whole application.
+
+3. Currently, only `lit-html` is supported (though other rendering engines could be incorporated in the future)
+
+4. You can not use tag selectors in css
```
thepassle:
```suggestion
4. You can not use tag selectors in css, but you could use a class name instead
```

packages/scoped-elements/README.md
```diff
@@ -0,0 +1,261 @@
...
+   ```css
+   🚫 my-panel {
+     width: 300px;
+   }
+   ✅ .panel {
+     width: 300px;
+   }
+   ```
+
+5. You can not use tag names using javascript querySelectors
```
thepassle:
```suggestion
5. You can not use tag names using javascript querySelectors, but you can query for the class name instead
```

packages/scoped-elements/README.md
frank-dspeed: Ok maybe i am missing something but as soon as this is merged i could add some extras from my code base i  for example check if the element that i get is the same class that i already wanted to register to dedub registration and this 

registration helper should not care where dependencys come from global dependencys are most time self versioning via namespaces look for example into lit-html it places a lit version var on the global scope

in this registration component you can completly ignore what your users are doing this has only one goal 

RREGISTER A NEW Component under a Uniq NAME or Return the Already Registered thats it all other dependes to the developers choice if he chooses to use global dependencys he is free to do so this will never replace scoped registry and scoped registry will also never handle global modules 

as scoped registry is not Isolate Components 

a pattern for isolated Components with a individual global scope can be also supplyed via IFRAME it adds a extra window object to the script that it can use and then we could deepCopy the Elements i have even a algo that does copy the event listners and all that which clone node and import node would ignore.

frank-dspeed: Ah Maybe the section should point out that this helper is best used with customElements that are coming from a 3th party ???? to make sure they don't  use a already registered tagName?

We Could also add some other interristing part a Loader !

This could even load a customElement and shim customElement calls via setting a scoped customElements Proxy object it don't needs to be a real JS Proxy but i choose this name as it does good explain what it should do it should simply reference this helper as customElement.define()

most self registering modules will have the line customElements.define()
if we wrapp that into import() call where the function that does the call has a local customElements Proxy it will get used!.

Other method is to use for example my esm-loader and simply remove calls to customElements.define.

manolakis: Ok, I think I'm misunderstanding something and also I apologize because my English level doesn't help to explain myself.

This note is referring to the first point.
>1. Components imported via npm **MUST NOT** be self registering components. If a shared component (installed from npm) does not offer an export to the class alone, without the registration side effect, then this component can not be used. E.g. every component that calls `customElement.define`

What it means is that if you are building a component that is susceptible to be used by third parties (which basically means that is going to be imported by other packages), you must **export the component class without registering the component in the registry**, so the third party would be able to use this feature registering the component safely.

If you are creating your components in the same package as the application shell then, probably, you are not going to share or distribute them, so you could register them in the same way you always do.

>Ok maybe i am missing something but as soon as this is merged i could add some extras from my code base i for example check if the element that i get is the same class that i already wanted to register to dedub registration and this

@frank-dspeed sorry but I'm not sure if I'm understanding you well. This feature is checking if the class or the tag are already registered. Each custom element class is going to be registered only once, and its tag is going to be reused by all their references.

packages/scoped-elements/README.md
manolakis: As I said in my previous comment this note refers to how you must share your component.

I think a good practice is export the component class and define it in separate files, like in Lion library, so when you are going to use the component you can decide to use the self-registered one or register it using scoped-elements feature.

manolakis: I think is better to remove this paragraph by the moment to avoid confussion, don't you think?

packages/scoped-elements/README.md
frank-dspeed: @manolakis that is not correct for example when some one does not bundle and is using components like
<script src="" type="module">
<script src="" type="module">
<script src="" type="module">
your map will not get reused as its scoped to the module context not the global context all 2 modules will register a indipendent version. of the Same Class 

so your code needs customElements.get()  === Klass 

to verify if it is already the same class that gets returned.

frank-dspeed: also it would be a good idea to drop the scope registration in favor of a uuid tagname and a global weak map the weakmap should get any uniqe name could be even a cryptic one to make sure it will never exist by any other module then this can get reused over all instances to make sure that registration happens only once for a customElement version that is all it is made for 

get uniqe tagname vor a customElement version based on the global context the customElement version can be validated via 2 ways first way ist customElement.get() === Klass and the secund more effective is class.toString() === customElements.get().toString() 
and then clean up that string so that white space and that don't matters as long as the code is the same content.

frank-dspeed: Its importend maybe to point out that ESM has a global module registry that means when you import a from 'a' in one script and import it in a other you get the same module.

packages/scoped-elements/README.md
frank-dspeed: as i pointed out if we add a check if the returned class from customElements get is === KlassToRegister we can simply ignore all that .

packages/scoped-elements/README.md
LarsDenBakker: Importing the same module twice will return the same module, that's one of the main points of modules.

packages/scoped-elements/README.md
frank-dspeed: he can also get the uniqeTagName via the class it self simply run its getTags function it is already auto registering getTags(class) => uniqeTagname

frank-dspeed: the scope part is only extra sugar that supplys a default prefix for the new uniqe tagName

LarsDenBakker: We don't export a `uniqeTagName` or `getTags` function.

packages/scoped-elements/README.md
manolakis: I changed it to

4. You can not use tag selectors in css, but you could use an id, a class name or even a property instead.

packages/scoped-elements/README.md
manolakis: I changed it into:

5. You can not use tag names using javascript querySelectors, but you could use an id, a class name or even a property instead.

packages/scoped-elements/README.md
thepassle: Alright, I completely lost the plot here 😄 Can we try and keep the discussions somewhat focussed?

packages/scoped-elements/README.md
daKmoR: added a Limitation: "Imported elements need to be fully side effect free (not only element registration)"

packages/scoped-elements/README.md
daKmoR: this was about that in your app shell (or your main app) you do not need to care about using scoped-elements... e.g. self registering elements are fine in that case... as these component are always unique to the app and can never become double-loaded...

but yeah maybe that is too much in depth... maybe it needs a full app example to get this concept clear 🙈

daKmoR:
thanks, everyone for making this happen 🎉  🤗

manolakis:
👏👏👏

Issue #1190, opened 2020-01-02
[testing-karma-bs] Setup user + key documentation incomplete


AERK-work-account:
Documentation on Browserstack setup user + key is incomplete. The example listed is not written for Windows users. See recommended edit below: 
```
#Write in Terminal or Command Line

#for one-time use only (on mac)
export BROWSER_STACK_USERNAME=xxx
export BROWSER_STACK_ACCESS_KEY=xxx

#or for one-time use only (on windows)
set BROWSER_STACK_USERNAME=xxx
set BROWSER_STACK_ACCESS_KEY=xxx

#or add them to your .bashrc
echo "export BROWSER_STACK_USERNAME=xxx" >> ~/.bashrc
echo "export BROWSER_STACK_ACCESS_KEY=xxx" >> ~/.bashrc

#to verify, run:
echo "User: $BROWSER_STACK_USERNAME"
echo "Key: $BROWSER_STACK_ACCESS_KEY"```

daKmoR:
looks good 👍

care to create a PR 🙏

LarsDenBakker:
The PR was merged, thanks!

PR #1191, opened 2020-01-03
update browserstack set up username + key docs

ARuzek:
This extends documentation to include windows command and fixes issue #1190


daKmoR:
thank you for your contribution 🤗

Issue #895, opened 2019-10-22
[docs] events deep dive


LarsDenBakker:
Events are a common source of confusion, people just make events bubbling and composed so that it works everywhere. That's a shame, I think we can really help with some deep dive docs.

Westbrook:
I've been working on this: https://glitch.com/edit/#!/super-area?path=all-three.js:1:0 and dev.to post in this area recently...I think it's a great idea to expand the conversation in this area.

thepassle:
@Westbrook do you think it would be possible to condense your blog post into a deep dive, and also a link to your blog for more in depth info? https://open-wc.org/faq/

Westbrook:
Interesting. What areas do you think would be best to include? If we can agree on a rough structure on what we'd like to present locally, happy to get something together!

Westbrook:
About halfway through expanding the initial release of this page. Look out for a PR on it sometime next week!

PR #1192, opened 2020-01-03
docs: add initial events deep dive

### LINKED ISSUE #1037: Add recommendation for event usage and firing
For usage of events have a TLDR.
- wire-up to local DOM via @some-event in template (let Lit do the actual wire-up)
- wire-up events to this in constructor (no need to clean up - as it will be garbage collected on destroying)
- wire-up to elements outside of scope (ex. window) using connectedCallback , and remove them in disconnectedCallback. 

For firing events have a TLDR.
- prefer "just" events without `bubble` or `composed` e.g. `new Event('something-happened')`
- when needing data consider extending `Event` or using `CustomEvent`
- use `bubble: true` if a node up the tree needs the event
- retrain from using `composed: true` as it leads to event pollution 
- use composed only if absolutely needed

Add a link for details too https://dev.to/open-wc/composed-true-considered-harmful-5g59

Westbrook:
Not wanting great to be the enemy of good, this is a first iteration for the event usage recommendations. Namely, I stole @daKmoR's great TLDR and dressed it up a bit. This should allow us to close #1037 and have something to start sharing as needed.

fixes #1037 
refs #895 

For #895, I'd like to get into the nuance of some of the recommendations and plan to expand on this with further exposition in a later iteration. If we're in agreement, let's get this out and then I can start a separate PR for us to discuss the additional content or code labs, etc.

### DIFF
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -7,7 +7,7 @@ In this section you can find answers to frequently asked questions regarding jav
 
 **Deep dives:**
 
-- [Component Libraries](/guide/component-libraries)
+- [Managing events in your custom elements](./events.html)
 - [Rerender not triggered](./rerender.html)
 - [Unit testing custom events](./unit-testing-custom-events.html)
 - [Unit testing initialization error](./unit-testing-init-error.html)
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -126,6 +126,7 @@ module.exports = {
           title: 'Deep dives',
           collapsable: true,
           children: [
+            'events',
             'rerender',
             'unit-testing-custom-events',
             'unit-testing-init-error',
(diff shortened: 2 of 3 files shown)


docs/faq/README.md
```diff
@@ -7,7 +7,6 @@ In this section you can find answers to frequently asked questions regarding jav
 
 **Deep dives:**
 
-- [Component Libraries](./component-libraries.html)
 - [Rerender not triggered](./rerender.html)
```
daKmoR: you wanna link to the events faq here right?

Westbrook: Interesting...how does something get into the `[[toc]]` annotation on these pages?

daKmoR: it's here
https://github.com/open-wc/open-wc/blob/master/docs/.vuepress/config.js#L128

Westbrook: Ok, this this iteration is good to go, now!

daKmoR:
awesome thxxx 🤗

PR #1194, opened 2020-01-04
feat(testing-helpers): add waitUntil helper

LarsDenBakker:
Async rendering is quite a hassle, this helper helps manage this declaratively in tests and demos. I have a version of this that I use a lot in my own code.

### DIFF
--- a/packages/testing-helpers/src/helpers.js
+++ b/packages/testing-helpers/src/helpers.js
@@ -130,3 +130,49 @@ export function oneEvent(element, eventName) {
     element.addEventListener(eventName, listener);
   });
 }
+
+/**
+ * Waits until the given predicate returns a truthy value. Calls and awaits the predicate
+ * function at the given interval time. Can be used to poll until a certain condition is true.
+ *
+ * @example
+ * ```js
+ * import { fixture, waitUntil } from '@open-wc/testing-helpers';
+ *
+ * const element = await fixture(html`<my-element></my-element>`);
+ *
+ * await waitUntil(() => element.someAsyncProperty, 'element should become ready');
+ * ```
+ *
+ * @param {() => boolean | Promise<boolean>} predicate - predicate function which is called each poll interval.
+ *   The predicate is awaited, so it can return a promise.
+ * @param {string} [message] an optional message to display when the condition timed out
+ * @param {{ interval?: number, timeout?: number }} [options] timeout and polling interval
+ */
+export function waitUntil(predicate, message, options = {}) {
+  const { interval = 50, timeout = 2000 } = options;
+
+  return new Promise((resolve, reject) => {
+    let timeoutId;
+
+    setTimeout(() => {
+      clearTimeout(timeoutId);
+      reject(new Error(message ? `Timeout: ${message}` : 'waitUntil timed out'));
+    }, timeout);
+
+    async function nextInterval() {
+      try {
+        if (await predicate()) {
+          resolve();
+        } else {
+          timeoutId = setTimeout(() => {
+            nextInterval();
+          }, interval);
+        }
+      } catch (error) {
+        reject(error);
+      }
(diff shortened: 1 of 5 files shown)


Westbrook:
Oooh, we have a version of this in our tools that I have on a list to contribute back, BUT this version is so much nicer. The options... the message... very exciting!

packages/testing-helpers/src/helpers.js
```diff
@@ -130,3 +130,51 @@ export function oneEvent(element, eventName) {
...
+ *
+ * await waitUntil(() => element.someAsyncProperty);
+ * ```
+ *
+ * @param {() => boolean | Promise<boolean>} asyncPredicate - predicate function which is called each poll interval.
+ *   The predicate is awaited, so it can return a promise or the actual value.
+ * @param {string} [message]
+ * @param {{ interval?: number, timeout?: number }} [options]
+ */
+export function waitUntil(asyncPredicate, message, options = {}) {
```
Westbrook: I remember some bikeshedding going on around the linguistics of other async method names...

`await waitUntil`

is not quite:

`await aTimeout`

Should this _maybe_ be

`await aPredicate`

Or similr?

packages/testing-helpers/test/helpers.test.js
```diff
@@ -68,4 +69,18 @@ describe('Helpers', () => {
...
+    let resolved = false;
+    waitUntil(() => foo).then(() => {
+      resolved = true;
+    });
+
+    await aTimeout(100);
+    foo = true;
+    await aTimeout(100);
+
+    expect(resolved).to.be.true;
```
Westbrook: As a practice question, how do you feel about `expect`ing the test to fail to begin with in order to confirm that the change actually occurred? Not a blocker in any way, just good knowledge for me to gather!

LarsDenBakker: Yeah good idea to add in an in-between check

packages/testing-helpers/src/helpers.js
LarsDenBakker: Mostly I try to pick names that read well when reading the source code. I don't think `aTimeout` is really a good name, but consistency would be nice. 

I think `aPredicate` doesn't really convey the idea that it's polling and waiting. I did some googling, I did see that `waitUntil` is used quite a bit. `until` would make sense too, but I don't want to overload the lit directive.

Open to other suggestions!

manolakis: That's an interesting problem, because `waitUtil` read better before using `async/await`.

```js
waitUntil(..)
  .then(...);
```

Also I think the name expresses very well what the function does. However with the use of `await` looks redundant reading the code `await waitUntil(...)` and names like `await until`, `await till` or event `await to` looks better reading the code, but I think the function name expresses worse what the function does.

IMHO I prefer to use names that expresses well what the function does, although we could have undesirable side effects like `await waitUntil`.

LarsDenBakker: I'm not too bothered by `await waituntil` personally

thepassle: ^ Same here

LarsDenBakker:
suggestion from @daKmoR: add a warning/note about when to use it

packages/testing-helpers/src/helpers.js
bennypowers: the param called `asyncPredicate` isn't really async, is it? like, the predicate is accessed synchronously, we're just polling it to add the async behaviour to the test.

So the sig should be `waitUntil(predicate, message?, options?)`

LarsDenBakker: The predicate can be both sync or async, we await the predicate as well. I'll call it `syncOrAsyncPredicate` :p

PR #1202, opened 2020-01-08
WIP: GitHub Actions init

DmitryMihailov:
## What was done
- Created GitHub Actions workflow, with migrated steps from [.circleci/config.yml](https://github.com/open-wc/open-wc/blob/master/.circleci/config.yml)
- CircleCI doesn’t act as build tool and now only runs all steps necessary to run BrowserStack testing

### Pros
- Executing same steps from CIrcleCI in Actions results in the same execution time (sometimes around 1 minute faster)
- Actions are closer/more-integrated with GitHub, obviously

### Cons
- `npm run test:node` and `npm run test:bs` being stuck frequently
- Secret variables aren’t available in PRs. Means `npm run test:node` being executed only in non-forked PRs and after merge to `master` branch.
- GitHub actions are in beta state. I expect performance improvements
 
## What must be fixed
- `npm run test:node` being stuck frequently 
- `npm run test:bs` being stuck frequently
- Windows `npm build` doesn’t work properly (need to switch to `npm-run-all`)


tmsns:
Would be great to see the `npm-ci-job` run on Windows as well. 😊 
The bs test is not really needed on Windows, as most of the time the npm script creation is cross-os.

.github/workflows/ci-workflow.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Restore cached NODE modules
+        uses: actions/cache@v1
+        with:
+          path: node_modules
+          key: v4-dependencies-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            v4-dependencies-${{ hashFiles('yarn.lock') }}
+            v4-dependencies- 
+      - name: "Publish packages"
+        run: "npx lerna publish --message 'chore: releasing new packages versions' --yes"
```
bennypowers: this is likely to fail when pushing tags, may also fail when publishing to npm.

I just this week (after trying longer than I'd care to admit) succeeded at [running `lerna publish` from github actions](https://github.com/apollo-elements/apollo-elements/blob/6e1bb2ab51a60cebbaed2c8aca86b2e8b2ec9336/.github/workflows/main.yml#L9-L54)

DmitryMihailov: I haven't yet reached that part though. Why should it fail?

DmitryMihailov:
@tmsns Added `windows-latest`. It fails on `npm run build`

```
Run npm run build && npm run build:types
At D:\a\_temp\05c0e193-7ab4-4c39-90af-7b86aec5be9f.ps1:2 char:15
+ npm run build && npm run build:types
+               ~~
The token '&&' is not a valid statement separator in this version.
+ CategoryInfo          : ParserError: (:) [], ParseException
+ FullyQualifiedErrorId : InvalidEndOfLine
 ```

File: .github/workflows/ci-workflow.yml (hunk 1 above)
Comment (bennypowers) [reply .github/workflows/ci-workflow.yml]: When GitHub Actions runs your workflow, the environment doesn't automatically have permissions to push commits or tags to your repository, which is what `lerna publish` does before it publishes to NPM.

Therefore, you have to create a GitHub Personal Access Token, store it in the repository's secrets, and reference it in the build step.

Similarly, in order to publish to NPM, you need to log in to the npm registry. It's not ideal to do this in CI using a username and password, so it's preferable to create an auth token on npmjs.com and reference it from the repo's secrets.

Both of these need to be configured in an incredibly specific way or they will fail.

If the process fails, the worst case scenario is that lerna will push tags to the repo and commit changes to `packages/**/package.json` (version bumps) and `packages/**/CHANGELOG.md`. These changes can be frustrating to undo in the case of failures.

**Comment by LarsDenBakker** [comment]
Running the node tests on windows would be really great indeed, it is a major blind spot for us.

I expect some tests and commands fail on windows, we will need to fix those.

**Comment by tmsns** [comment]
> @tmsns Added `windows-latest`. It fails on `npm run build`
> 
> ```
> Run npm run build && npm run build:types
> At D:\a\_temp\05c0e193-7ab4-4c39-90af-7b86aec5be9f.ps1:2 char:15
> + npm run build && npm run build:types
> +               ~~
> The token '&&' is not a valid statement separator in this version.
> + CategoryInfo          : ParserError: (:) [], ParseException
> + FullyQualifiedErrorId : InvalidEndOfLine
> ```

Seems the windows image is using Powershell as the command-line environment. Over there the `;` character is used to delimit commands. The normal CMD would have worked, though. 

In order to be trully os and command line indepent, we should switch over to JS. Some packages like `npm-run-all` or `concurrently` can help here. 😃

**Comment by DmitryMihailov** [comment]
> > @tmsns Added `windows-latest`. It fails on `npm run build`
> > ```
> > Run npm run build && npm run build:types
> > At D:\a\_temp\05c0e193-7ab4-4c39-90af-7b86aec5be9f.ps1:2 char:15
> > + npm run build && npm run build:types
> > +               ~~
> > The token '&&' is not a valid statement separator in this version.
> > + CategoryInfo          : ParserError: (:) [], ParseException
> > + FullyQualifiedErrorId : InvalidEndOfLine
> > ```
> 
> Seems the windows image is using Powershell as the command-line environment. Over there the `;` character is used to delimit commands. The normal CMD would have worked, though.
> 
> In order to be trully os and command line indepent, we should switch over to JS. Some packages like `npm-run-all` or `concurrently` can help here. 😃

So commands do must be re-written for npm or we just need to use `npm-run-all` / `concurrently` ?

**Comment by LarsDenBakker** [comment]
Yeah lets switch to npm-run-all

File: .github/workflows/ci-workflow.yml [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+        run: yarn install --frozen-lockfile
+      - name: "NPM run build"
+        run: npm run build && npm run build:types
+      - name: "Code linting"
+        run: npm run lint
+      - name: "Tests - NODE"
+        run: npm run test:node
+      - name: "Tests - headless browser"
+        run: npm run test:browser
+      - name: "Tests - BrowserStack"
```
daKmoR: I think the idea was to 
- run github actions only in the headless browser e.g. use it as the "fast lane" to get fast feedback
- run circleci with browserstack e.g. as the "slow lane" which gives "better" feedback but it takes longer

so we could remove the browserstack here right?

.github/workflows/ci-workflow.yml
daKmoR: jup deploying probably should be tested with a custom registry first 🙈 
these things are always scary to change 😱 

locally we could use https://verdaccio.org/ but not sure how to reproduce it then? can you run actions locally? or can you define a registry that points to your locally running machine? (just for testing?)

bennypowers: you can test a deploy to some local verdaccio on pull-request actions
I wouldn't draw much confidence from that, due to the finickiness of lerna, yarn, and friends

Issue #1203, opened 2020-01-08
yarn storybook:build fails


Westbrook:
When running `yarn storybook:build` I get the output below after updating the following packages:

```
"@open-wc/building-webpack": "^2.10.5" => "^2.10.8",
"@open-wc/demoing-storybook": "^1.4.0" => "^1.5.2",
"@open-wc/testing": "^2.4.2" => "^2.5.0",
"@open-wc/testing-karma": "^3.2.20" => "^3.2.28",
```

It looks like the specific upgrade that causes this is `@open-wc/demoing-storybook`, but I'm still investigating and breaking it down.

```
23:46 ~/Documents/web/repos/spectrum-web-components (westbrook/greenkeeper)!$ yarn storybook:build
yarn run v1.17.3
$ yarn storybook:stories:build && build-storybook
$ tsc --build .storybook/tsconfig.json
The "onwrite" hook used by plugin rollup-plugin-cpy is deprecated. The "generateBundle/writeBundle" hook should be used instead.
The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten
The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten
[BABEL] Note: The code generator has deoptimised the styling of /Users/westbrook/Documents/web/repos/spectrum-web-components/node_modules/@open-wc/storybook-prebuilt/dist/preview.js as it exceeds the max of 500KB.

<--- Last few GCs --->

[38609:0x102802000]    39933 ms: Scavenge 1315.5 (1422.9) -> 1314.6 (1423.4) MB, 2.6 / 0.0 ms  (average mu = 0.201, current mu = 0.156) allocation failure 
[38609:0x102802000]    39937 ms: Scavenge 1316.1 (1423.4) -> 1315.3 (1423.4) MB, 2.5 / 0.0 ms  (average mu = 0.201, current mu = 0.156) allocation failure 
[38609:0x102802000]    39941 ms: Scavenge 1316.2 (1423.4) -> 1315.4 (1423.9) MB, 3.0 / 0.0 ms  (average mu = 0.201, current mu = 0.156) allocation failure 


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x27ad5085be3d]
Security context: 0x31bddb19e6e9 <JSObject>
[...]

PR #1207, opened 2020-01-09
Fix/storybook build crash

LarsDenBakker:
This skips babel and terser for the storybook prebuilt, since that's already es5 and minified. 

It also runs the builds in sequence instead of in parallel, this is because terser takes up the majority of the build time and it's already multi threaded. Running two builds with terser in parallel is therefore not necessary, and potentially problematic.

I'm hoping this fixes https://github.com/open-wc/open-wc/issues/1203

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ dist
 build-stats.json
 stats.html
 .rpt2_cache
-static-storybook
+storybook-static
 
 ## we already store snapshots globally
 packages/**/__snapshots__
@@ -43,3 +43,4 @@ local.log
 
 ## generated codelabs
 docs/.vuepress/public/codelabs
+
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -33,7 +33,7 @@ dist
 build-stats.json
 stats.html
 .rpt2_cache
-static-storybook
+storybook-static
```
Westbrook: `-static` here but `-preview` elsewhere... is this change related to the other files?

Westbrook:
@LarsDenBakker other than copying files into my `node_modules` is there a better way to confirm this fix works for my build...patient 0 as it were.

.gitignore
LarsDenBakker: It's unrelated, `storybook-static` is the build output which was never added to the gitignore correctly

LarsDenBakker:
I think that's the best confirmation we can get

PR #1212, opened 2020-01-10
fix(testing-karma): latest karma-chrome-launcher decreases memory use

### LINKED ISSUE #1211: [testing-karma] Disconnected, because no message in 10000 ms.
I get this issue pretty regularly (maybe 40% of the time) when running my tests in CI, though almost never when running locally 😓 However, it seems there is a path forward: https://github.com/jasmine/jasmine/issues/1327#issuecomment-524281104

I added `karma-chrome-launcher@^3.1.0` to my local project and manually removed `karma-chrome-launcher@^2.0.0` from my `node_modules` as a rough test of this, and my tests ran fine on local. So, hopefully, there's no reason to worry about this change, generally. Whether it actually solves my CI issue or not 🤞 

I'll make a PR for the update of `karma-chrome-launcher@^2.0.0` -> `karma-chrome-launcher@^3.1.0`, but please respond here with any issues you think that might bring to light.

Westbrook:
fixes #1211 

Run tests with `--disable-dev-sgm-usage` as outlined in the setting for `karma-chrome-launcher@^3.1.0`.

### DIFF
--- a/packages/testing-karma/package.json
+++ b/packages/testing-karma/package.json
@@ -31,7 +31,7 @@
     "@open-wc/karma-esm": "^2.12.5",
     "axe-core": "^3.3.1",
     "karma": "^4.1.0",
-    "karma-chrome-launcher": "^2.0.0",
+    "karma-chrome-launcher": "^3.1.0",
     "karma-coverage-istanbul-reporter": "^2.0.0",
     "karma-mocha": "^1.0.0",
     "karma-mocha-reporter": "^2.0.0",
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
@Westbrook shouldn't we also add the flag to `testing-karma` ?

Westbrook:
@LarsDenBakker I’LL look a little deeper at what that might look like. I think it calls straight through the launcher _everytime_ but not fully versed.

Westbrook:
The alpha release outlined here is working: https://github.com/open-wc/open-wc/issues/1211#issuecomment-573315642 Running my CI a couple of times to see if I can run into the memory issues or not.

@LarsDenBakker being the `testing-karma` [ChromeHeadlessNoSandbox](https://github.com/open-wc/open-wc/blob/master/packages/testing-karma/src/create-default-config.js#L35) is built from the browser outlined in `karma-chrome-launcher` do you think there is a need to do anything more internal configuration setting? Happy to add it to that file as a back up if you think that's useful.

Westbrook:
I still run into the `Disconnected (0 times), because no message in 30000 ms.` message on CI, maybe I _DO_ need to include it in our actual code? 😞

PR #1214, opened 2020-01-11
fix(es-dev-server): don't look for config in root dir

LarsDenBakker:
As discussed in https://github.com/open-wc/open-wc/commit/4085898bf9fd23ab51cd86d5fe582191c30be64e#r36748834, we introduced a breaking change by looking in the root dir for a config. In hindsight, this change was not needed for achieving what we need so we should revert it.


LarsDenBakker:
@tmsns

PR #1217, opened 2020-01-12
Fix/demoing storybook build

LarsDenBakker:
Storybook builds were taking incredibly long and perceptible to crashing because it ran out of memory. 

After a lot of investigation it turned out to be a simple problem with a glob pattern for copying the `custom-elements.json` file, which was recursing into `node_modules` directories. Because `rollup-plugin-cpy` is using a deprecated API, I took the opportunity to replace it with a simpler version. 

Besides this I made some more optimization, running the bundle and build for one config. This way the previous could be garbage collected if builds take really long.

Finally I suppressed some other warnings during the build to make the output cleaner. These are warnings we can do anything about when using storybook.


LarsDenBakker:
@Westbrook this is probably the proper fix for the issues you were seeing as well

LarsDenBakker:
Great to hear :) we can make it even faster if we allow dropping the IE11 build (if you dont need to support IE11)

PR #1224, opened 2020-01-13
docs(testing-helpers): clarify oneEvent docs

### LINKED ISSUE #1173: [testing-helpers] oneEvent docs incorrect
The jsdocs for `oneEvent` should accept any `EventTarget`, not specifically a `HTMLElement`. https://github.com/open-wc/open-wc/blob/master/packages/testing-helpers/src/helpers.js#L124

bennypowers:
fixes #1173


LarsDenBakker:
Looks good to me, except now it's giving an issue with the linting :(

bennypowers:
Problem was changing the return to to Event, which is not extensible via a `detail` property, like CustomEvent

PR #1227, opened 2020-01-13
WIP: feat(demoing-storybook): add support for reading main.js

LarsDenBakker:

PR #1230, opened 2020-01-13
feat(demoing-storybook): support main.js

daKmoR:
alternative idea to 
https://github.com/open-wc/open-wc/pull/1227/

preread `main.js` and leave everything else as is... if we wanna go for that we would need to apply the same to the build

### DIFF
--- a/packages/demoing-storybook/README.md
+++ b/packages/demoing-storybook/README.md
@@ -89,7 +89,7 @@ The storybook server is based on [es-dev-server](https://open-wc.org/developing/
 
 ### Configuration file
 
-By default, `@open-wc/demoing-storybook` looks for config files called `main.js` in your config dir (default `.storybook`). You can modify this by using the `--config` flag.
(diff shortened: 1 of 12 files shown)


LarsDenBakker:
I like it, definitely a lot simpler than my approach.

daKmoR:
also closes #1231

packages/demoing-storybook/README.md
```diff
@@ -89,16 +89,19 @@ The storybook server is based on [es-dev-server](https://open-wc.org/developing/
 
 ### Configuration file
 
-By default, `@open-wc/demoing-storybook` looks for config files called `start-storybook.config.js` and `build-storybook.config.js` in your config dir (default `.storybook`). You can modify this by using the `--config` or `-c` flag.
+By default, `@open-wc/demoing-storybook` looks for config files called `main.js` in your config dir (default `.storybook`). You can modify this by using the `--config` flag.
```
LarsDenBakker:
```suggestion
By default, `@open-wc/demoing-storybook` looks for config files called `main.js` in your config dir (default `.storybook`). 
```

PR #1236, opened 2020-01-15
docs(demoing-storybook): config property name correction

Westbrook:
Porting over my implementation to `main.js` and caught what is probably just a typo?

`outputPath` -> `outputDir`


LarsDenBakker:
Thanks!

PR #1237, opened 2020-01-15
docs: expanded event documentation

### LINKED ISSUE #895: [docs] events deep dive

Westbrook:
Expand on event documentation.

closes #895 

Please review the copy here for clarity and desired context. I've only just finished the first draft and will be reviewing it tomorrow or the next myself.

### DIFF
--- a/docs/faq/events.md
+++ b/docs/faq/events.md
@@ -59,8 +59,7 @@ If you are adding a listener to your element, from inside your element, you shou
 ```js
 constructor() {
     super();
-    this.handleDone = this.handleDone.bind(this);
-    this.addEventListener('done', this.handleDone);
+    this.addEventListener('done', this.handleDone.bind(this));
 }
 handleDone(event) {
     this.done = true;
@@ -88,7 +87,7 @@ handleResize(event) {
 }
 ```
 
-Ensuring that event listeners do not stick around beyond the presence of your element in the DOM is just as straight forward with the `disconnectedCallback` method.
+Ensuring that event listeners do not stick around beyond the presence of your element in the DOM fits neatly into the `disconnectedCallback` method:
 
 ```js
 disconnectedCallback() {
--- a/docs/faq/events.md
+++ b/docs/faq/events.md
@@ -33,7 +33,7 @@ There are lots of times when words just aren't enough to really get clarity on a
 
 ## Event listening
 
-When preparing to listen for events in your code, do the least work possible. Not only because this allows you to do less work now, but it allows you also to do less work later, and most importantly it allows the browser to do less work, which means a more performant application or component. The leasst work possible changes by context, but in important to remember no matter where you are adding event listners.
+When preparing to listen for events in your code, do the least work possible. Not only because this allows you to do less work now, but it allows you also to do less work later, and most importantly it allows the browser to do less work, which means a more performant application or component. The least work possible changes per context, but in important to remember no matter where you are adding event listeners.
 
 ### Inside of your elements
 
@@ -54,7 +54,7 @@ A side benefit of relying on the `@event` syntax in `LitElement` (if you use it)
 
 ### On your elements
 
-If you are adding a listener to your element, from inside your element, you should add that listener in the `constructor`. The browser does a great job of garbage collecting that listener when the element is removed from the DOM, so there no reason to do any work to manage that yourself. See the following example where the element listeners for a `done` event.
+If you are adding a listener to your element, from inside your element, you should add that listener in the `constructor`. The browser does a great job of garbage collecting that listener when the element is removed from the DOM, so there is no reason to do any work to manage that yourself. See the following example where the element listens for a `done` event.
 
 ```js
 constructor() {
@@ -83,7 +83,7 @@ handleResize = (event) => {
 }
 ```
 
-Ensuring that event listener does not stick around beyond the presence of your element in the DOM is just as straight forward with the `disconnectedCallback` method.
+Ensuring that event listeners do not stick around beyond the presence of your element in the DOM is just as straight forward with the `disconnectedCallback` method.
 
 ```js
 disconnectedCallback() {
@@ -92,7 +92,7 @@ disconnectedCallback() {
 }
 ```
 
-Side note: again, as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the class field method syntax above. When removing event listeners is it also important to ensure that the identity of the callback bound to the event is available by reference or it won't be propertly removed.
+Side note: again, as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the class field method syntax above. When removing event listeners is it also important to ensure that the identity of the callback bound to the event is available by reference or it won't be properly removed.
 
 ## Event dispatching
 
@@ -126,7 +126,7 @@ el.dispatchEvent(
 
 Be aware that a `composed` event will be retargeted to each shadow root that it passes through. In this way, the value for `event.target` will be the first of an element with a shadow DOM or the dispatching element as you travel from the listening element towards the dispatching element. This is one way in which a shadow root provides encapsulation for the inner working of your element. If you ever need the entire path across the DOM that your event will take you can use the `event.composedPath()` method to acquire an array of elements (starting with the dispatching element) that the event will pass through.
 
-With both of these properties managed, your event will both begin its `capture` phase at the `window` as well as also a `bubble` phase which ends at `window` as well. In this case, your event will look like the following:
+With both of these properties managed, your event will both begin its `capture` phase at the `window` as well as a `bubble` phase which ends at the `window`. In this case, your event will look like the following:
 
 ```js
 el.dispatchEvent(
@@ -137,11 +137,11 @@ el.dispatchEvent(
 );
 ```
 
-Be conscious with your use of `composed: true`. With itall parent components and implementors will also have access to your event in both phases, which can lead to event polution. Only use these settings when you mean for the entire application to have access to an event.
+Be conscious with your use of `composed: true`, though. When using `composed`, all parent components and implementors will also have access to your event in both phases, which can lead to event pollution. Only use these settings when you mean for the entire application to have access to an event.
 
 ## Even more data
 
-When you'd like to pass data along with your event, you also two options: dispatch a `new CustomEvent()` or extend the `Event` class. While no native DOM element uses the `CustomEvent` constructor to create the events that they dispatch, developers are offered this contructor and its additional `detail` property as a way to pass additional information along with their event. This can be done like so:
+When you'd like to pass data along with your event, you have two options: dispatch a `new CustomEvent()` or extend the `Event` class. While no native DOM element uses the `CustomEvent` constructor to create the events that they dispatch, developers are offered this contructor and its additional `detail` property as a way to pass additional information along with their event. This can be done like so:
 
 ```js
 el.dispatchEvent(
@@ -167,7 +167,7 @@ el.didispatchEventspatch(customEvent);
 console.log(customEvent.detail.wasHeard); // true
 ```
 
-Developers also access to the `Event` constructor which can be extended to contain not only data but also functionality.
+Developers may also access to the `Event` constructor which can be extended to contain not only data but also functionality.
 
 ```js
 class MyEvent extends Event {
@@ -198,7 +198,7 @@ el.dispatchEvent(customEvent); // Something imporant
 console.log(customEvent.dataReceived); // true
 ```
 
-An additional benefit of extending the `Event` class is the ability to do `instanceof` checking in your listener. This can be a great way to disabiguate one `custom-event` from another `custom-event`.
+An additional benefit of extending the `Event` class is the ability to do `instanceof` checking in your listener. This can be a great way to disambiguate one `custom-event` from another `custom-event`.
 
 ```js
 parent.addEventListener('custom-event', event => {
--- a/docs/faq/events.md
+++ b/docs/faq/events.md
@@ -37,7 +37,7 @@ When preparing to listen for events in your code, do the least work possible. No
 
 ### Inside of your elements
 
-When adding event listening inside of your components, assuming you are using `lit-html` (or any other declarative renderer with event binding syntax) use the tools supplied therein to [add event listeners](https://lit-html.polymer-project.org/guide/writing-templates#add-event-listeners). The reasoning behind this is that for any listener added by `lit-html`, the same listener will be removed by `lit-html` when no longer needed; no ghost listeners! That means whether the DOM and its associated lister is available for the entire lifecycle of your application, or just for a small part of that, the least amount of work will be done to ensure the event is handled appropriately. See the `click` listener in the following code. It _only_ exits while `this.open` is `true`, which leaves the browser listening to less things and your code less likely to trigger handlers errantly.
+When adding event listening inside of your components, assuming you are using `lit-html` (or any other declarative renderer with event binding syntax) use the tools supplied therein to [add event listeners](https://lit-html.polymer-project.org/guide/writing-templates#add-event-listeners). The reasoning behind this is that for any listener added by `lit-html`, the same listener will be removed by `lit-html` when no longer needed; no ghost listeners! That means whether the DOM and its associated listener is available for the entire lifecycle of your application, or just for a small part of that, the least amount of work will be done to ensure the event is handled appropriately. See the `click` listener in the following code. It _only_ exits while `this.open` is `true`, which leaves the browser listening to fewer things and your code less likely to trigger handlers errantly.
 
 ```js
 render() {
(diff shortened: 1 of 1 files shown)


docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
     style="height: 100%; width: 100%; border: 0;">
   </iframe>
 </div>
+
+## Event listening
+
+When preparing to listen for events in your code, do the least work possible. Not only because this allows you to do less work now, but it allows you also to do less work later, and most importantly it allows the browser to do less work, which means a more performant application or component. The leasst work possible changes by context, but in important to remember no matter where you are adding event listners.
```
thepassle:
```suggestion
When preparing to listen for events in your code, do the least work possible. Not only because this allows you to do less work now, but it allows you also to do less work later, and most importantly it allows the browser to do less work, which means a more performant application or component. The least work possible changes per context, but in important to remember no matter where you are adding event listeners.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+### Inside of your elements
+
+When adding event listening inside of your components, assuming you are using `lit-html` (or any other declarative renderer with event binding syntax) use the tools supplied therein to [add event listeners](https://lit-html.polymer-project.org/guide/writing-templates#add-event-listeners). The reasoning behind this is that for any listener added by `lit-html`, the same listener will be removed by `lit-html` when no longer needed; no ghost listeners! That means whether the DOM and its associated lister is available for the entire lifecycle of your application, or just for a small part of that, the least amount of work will be done to ensure the event is handled appropriately. See the `click` listener in the following code. It _only_ exits while `this.open` is `true`, which leaves the browser listening to less things and your code less likely to trigger handlers errantly.
```
thepassle:
```suggestion
When adding event listening inside of your components, assuming you are using `lit-html` (or any other declarative renderer with event binding syntax) use the tools supplied therein to [add event listeners](https://lit-html.polymer-project.org/guide/writing-templates#add-event-listeners). The reasoning behind this is that for any listener added by `lit-html`, the same listener will be removed by `lit-html` when no longer needed; no ghost listeners! That means whether the DOM and its associated listener is available for the entire lifecycle of your application, or just for a small part of that, the least amount of work will be done to ensure the event is handled appropriately. See the `click` listener in the following code. It _only_ exits while `this.open` is `true`, which leaves the browser listening to less things and your code less likely to trigger handlers errantly.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+        }
+    `;
+}
+```
+
+A side benefit of relying on the `@event` syntax in `LitElement` (if you use it) is that all events handled therein are automatically bound to `this` which will prevent the need to do so manually: no `this.handleClose = this.handleClose.bind(this);` and, no `handleClose = (event) => {}` in the case above.
+
+### On your elements
+
+If you are adding a listener to your element, from inside your element, you should add that listener in the `constructor`. The browser does a great job of garbage collecting that listener when the element is removed from the DOM, so there no reason to do any work to manage that yourself. See the following example where the element listeners for a `done` event.
```
thepassle:
```suggestion
If you are adding a listener to your element, from inside your element, you should add that listener in the `constructor`. The browser does a great job of garbage collecting that listener when the element is removed from the DOM, so there is no reason to do any work to manage that yourself. See the following example where the element listens for a `done` event.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+```js
+constructor() {
+    super();
+    this.addEventListener('done', this.handleDone);
+}
+handleDone = (event) => {
```
thepassle: Prefer to use the bind syntax here. Public properties are not supported everywhere yet, people may come to this page, scan it for what they need, copy + paste code, and go "it doesnt work!"

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+    this.done = true;
+    // Do the work you'd like to complete when "done" here.
+}
+```
+
+Side note: as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the class field method syntax above.
```
thepassle: and then this note can be removed

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+### On elements outside of your element
+
+In some cases you will need to add event listener to elements outside of the element you are building: e.g. a `resize` listener of the `window`. In this case the `window` could last well beyond the last referenc to your element and it is very important that you do the work to ensure that the listener is added and removed appropriately. Thankfully, the custom elements specification outlines the [`connectedCallback` and `disconnectedCallback` lifecycle methods](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Using_the_lifecycle_callbacks). These are the perfect hooks for managing these sorts of listeners. When adding event listeners to this sort of external element, use the `connectedCallback`, like so:
```
thepassle:
```suggestion
In some cases you will need to add event listeners to elements outside of the element you are building: e.g. a `resize` listener on the `window`. In this case the `window` could last well beyond the last reference to your element and it is very important that you do the work to ensure that the listener is added and removed appropriately. Thankfully, the custom elements specification outlines the [`connectedCallback` and `disconnectedCallback` lifecycle methods](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Using_the_lifecycle_callbacks). These are the perfect hooks for managing these sorts of listeners. When adding event listeners to this sort of external element, use the `connectedCallback`, like so:
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+```js
+connectedCallback() {
+    super.connectedCallback && super.connectedCallback();
+    window.addEventListener('resize', this.handleResize);
+}
+handleResize = (event) => {
```
thepassle: same here

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+    // Do something with the resize event here.
+}
+```
+
+Ensuring that event listener does not stick around beyond the presence of your element in the DOM is just as straight forward with the `disconnectedCallback` method.
```
thepassle:
```suggestion
Ensuring that event listeners do not stick around beyond the presence of your element in the DOM is just as straight forward with the `disconnectedCallback` method.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+```js
+disconnectedCallback() {
+    window.removeEventListener('resize', this.handleResize);
+    super.disconnectedCallback && super.disconnectedCallback();
+}
+```
+
+Side note: again, as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the class field method syntax above. When removing event listeners is it also important to ensure that the identity of the callback bound to the event is available by reference or it won't be propertly removed.
```
thepassle:
```suggestion
Side note: again, as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the class field method syntax above. When removing event listeners is it also important to ensure that the identity of the callback bound to the event is available by reference or it won't be properly removed.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+## Event dispatching
+
+### Keep it simple
+
+You can dispatch a DOM event on any element in your applicaiton with `el.dispatchEvent(new Event('event'))`. With just that amount of code you'll be able to trigger the callback for any listener for that event name that is bound to that element. You'll also trigger the callback on any ancestor elements in the same DOM tree for listeners bound to the "capture" phase of an event. In a lot of cases this might be all of the functionality you really need when dispatching an event, so compare your usage needs before worrying too much about the following.
```
thepassle:
```suggestion
You can dispatch a DOM event on any element in your application with `el.dispatchEvent(new Event('event'))`. With just that amount of code you'll be able to trigger the callback for any listener for that event name that is bound to that element. You'll also trigger the callback on any ancestor elements in the same DOM tree for listeners bound to the "capture" phase of an event. In a lot of cases this might be all of the functionality you really need when dispatching an event, so compare your usage needs before worrying too much about the following.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+```js
+el.dispatchEvent(
+  new Event('event', {
+    bubbles: true,
+  }),
+);
+```
+
+This will allow your event to "bubble" up to the top of the current DOM tree. `bubbles` isn't required in order for an event to be accessible to the ancestors of the element on which is it dispatched, but in that listening to events on the "capture" phase can sometimes be confusing, involving of extended techniques to do in declarative templating, and requiring of extended documentation, it is best that you do use it in these cases.
```
thepassle: this second sentence is really long and confusing, starting at "`bubbles` isn't required .... " to "do use it in these cases". Could you reword a bit and make it multiple sentence instead of one long sentence?

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+el.dispatchEvent(
+  new Event('event', {
+    composed: true,
+  }),
+);
+```
+
+Be aware that a `composed` event will be retargeted to each shadow root that it passes through. In this way, the value for `event.target` will be the first of an element with a shadow DOM or the dispatching element as you travel from the listening element towards the dispatching element. This is one way in which a shadow root provides encapsulation for the inner working of your element. If you ever need the entire path across the DOM that your event will take you can use the `event.composedPath()` method to acquire an array of elements (starting with the dispatching element) that the event will pass through.
+
+With both of these properties managed, your event will both begin its `capture` phase at the `window` as well as also a `bubble` phase which ends at `window` as well. In this case, your event will look like the following:
```
thepassle:
```suggestion
With both of these properties managed, your event will both begin its `capture` phase at the `window` as well as a `bubble` phase which ends at the `window`. In this case, your event will look like the following:
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+```js
+el.dispatchEvent(
+  new Event('event', {
+    bubbles: true,
+    composed: true,
+  }),
+);
+```
+
+Be conscious with your use of `composed: true`. With itall parent components and implementors will also have access to your event in both phases, which can lead to event polution. Only use these settings when you mean for the entire application to have access to an event.
```
thepassle:
```suggestion
Be conscious with your use of `composed: true`, though. When using `composed`, all parent components and implementors will also have access to your event in both phases, which can lead to event pollution. Only use these settings when you mean for the entire application to have access to an event.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+## Even more data
+
+When you'd like to pass data along with your event, you also two options: dispatch a `new CustomEvent()` or extend the `Event` class. While no native DOM element uses the `CustomEvent` constructor to create the events that they dispatch, developers are offered this contructor and its additional `detail` property as a way to pass additional information along with their event. This can be done like so:
```
thepassle:
```suggestion
When you'd like to pass data along with your event, you have two options: dispatch a `new CustomEvent()` or extend the `Event` class. While no native DOM element uses the `CustomEvent` constructor to create the events that they dispatch, developers are offered this contructor and its additional `detail` property as a way to pass additional information along with their event. This can be done like so:
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+const customEvent = new CustomEvent('custom-event', {
+  detail: {
+    wasHeard: false,
+  },
+});
+el.didispatchEventspatch(customEvent);
+console.log(customEvent.detail.wasHeard); // true
+```
+
+Developers also access to the `Event` constructor which can be extended to contain not only data but also functionality.
```
thepassle:
```suggestion
Developers may also access to the `Event` constructor which can be extended to contain not only data but also functionality.
```

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+    this.importantData = importantData;
+    this.dataReceived = false;
+  }
+  logImportantData() {
+    console.log(this.importantData);
+  }
+}
+
+parent.addEventListener('custom-event', event => {
+  event.logImportantData();
```
thepassle: very nice example 😄

docs/faq/events.md
```diff
@@ -28,3 +30,182 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+    bubbles: true,
+  },
+  'Something important',
+);
+console.log(customEvent.dataReceived); // false
+el.dispatchEvent(customEvent); // Something imporant
+console.log(customEvent.dataReceived); // true
+```
+
+An additional benefit of extending the `Event` class is the ability to do `instanceof` checking in your listener. This can be a great way to disabiguate one `custom-event` from another `custom-event`.
```
thepassle:
```suggestion
An additional benefit of extending the `Event` class is the ability to do `instanceof` checking in your listener. This can be a great way to disambiguate one `custom-event` from another `custom-event`.
```

docs/faq/events.md
Westbrook: I think saying something, especially after noting not having to bind in `lit-html`, is important. But, I'll update this along with the `.bind()` addition above.

docs/faq/events.md
Westbrook: I don't see changed here. Am I missing something?

docs/faq/events.md
Westbrook: Changes?

docs/faq/events.md
Westbrook: Changes?

docs/faq/events.md
thepassle: lister to listener

docs/faq/events.md
thepassle: listener to listeners
referenc to reference

docs/faq/events.md
thepassle: you should be able to see the changes in the diff

Westbrook:
@thepassle Thanks for the close review! I've addressed all the notes I can understand. If there are further updates for those three I pointed out, let me know and I'll get right on them. 

Good point on the `.bind()` vs `method = () => {}`. However, I do wish there were a _stable/clear_ way to not imply we think people have to use the `.bind()` syntax.

I also make a Grammarly pass on this, so hopefully, that gets the general writing into an OK place...

thepassle:
Great :) Did you run grammarly yet? Or you still have to do it?

Westbrook:
Ran Grammarly and I’ll do a pass on those suggestions I missed a little later, if they aren’t already updated.

thepassle:
Ok so this can be merged then right?

docs/faq/events.md
```diff
@@ -28,3 +30,187 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+
+### On your elements
+
+If you are adding a listener to your element, from inside your element, you should add that listener in the `constructor`. The browser does a great job of garbage collecting that listener when the element is removed from the DOM, so there is no reason to do any work to manage that yourself. See the following example where the element listens for a `done` event.
+
+```js
+constructor() {
+    super();
+    this.handleDone = this.handleDone.bind(this);
+    this.addEventListener('done', this.handleDone);
```
daKmoR: if you don't need a link to the original thing then you can do
```js
this.addEventListener('done', this.handleDone.bind(this));
```
directly

docs/faq/events.md
```diff
@@ -28,3 +30,187 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+Side note: as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the `.bind()` syntax above.
+
+### On elements outside of your element
+
+In some cases, you will need to add an event listener to elements outside of the element you are building: e.g. a `resize` listener of the `window`. When doing so, the `window` could last well beyond the last reference to your element and it is very important that you do the work to ensure that the listener is added and removed appropriately. Thankfully, the custom elements specification outlines the [`connectedCallback` and `disconnectedCallback` lifecycle methods](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Using_the_lifecycle_callbacks). These are the perfect hooks for managing these sorts of listeners. When adding event listeners to this sort of external element, use the `connectedCallback`, like so:
+
+```js
+constructor() {
+    super();
+    this.handleResize = this.handleResize.bind(this);
```
daKmoR: addEventListener missing

docs/faq/events.md
```diff
@@ -28,3 +30,187 @@ There are lots of times when words just aren't enough to really get clarity on a
...
+  }),
+);
+```
+
+Remember to be conscious of your use of `composed: true`. When using `composed`, all parent components and implementors will also have access to your event in both phases, which can lead to event pollution. Only use these settings when you mean for the entire application to have access to an event.
+
+## Even more data
+
+When you'd like to pass data along with your event, you have two options: dispatch a `new CustomEvent()` or extend the `Event` class. While no native DOM element uses the `CustomEvent` constructor to create the events that they dispatch, developers are offered this constructor and its additional `detail` property as a way to pass additional information along with their event. This can be done like so:
+
```
daKmoR: add a "warning" that this can not be typed... so it would be better to use the approach below

LarsDenBakker: CustomEvent caan be typed with generics, why do people dislike it. Seems like creating a class is a lot more overhead

Westbrook: I’m with Lars on this. What’s more very little of the Open WC content is directly in reference to TS, so I’m inclined to skip this. Let me know if you _really_ want it in.

docs/faq/events.md
Westbrook: Good call. This will also strengthen the note on keeping a reference below.

Westbrook:
@thepassle I'm ready to merge this...but, I'll give @daKmoR a final word on his types request.

LarsDenBakker:
LGTM

PR #1249, opened 2020-01-16
ci: attempt to persist to workspace

Westbrook:
This PR is a work in progress of a proof of concept of using CircleCI workflows to decrease build times.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ defaults: &defaults
     - image: circleci/node:10.16.0-browsers
 
 jobs:
-  test:
+  build:
     <<: *defaults
     steps:
       - checkout
@@ -18,15 +18,6 @@ jobs:
       - restore_cache:
           keys:
             - v4-dependencies-{{ checksum "yarn.lock" }}
-      # - run:
-      #     name: Do `yarn install` on cache miss
-      #     command: |
-      #       if [ -d "./node_modules" ]; then
-      #         echo "Restored node_modules from cache";
-      #       else
-      #         echo "Fill node_modules via yarn install";
-      #         yarn install --frozen-lockfile
-      #       fi
       # seems we always need to run yarn for mono repos to get nested dependencies and bin links
       - run: yarn install --frozen-lockfile
       - save_cache:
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
For the "fast lane" we could probably run browser tests in local chrome

Westbrook:
@LarsDenBakker I think the "fast lane" is getting to be in a good place here. Thoughts?

Westbrook:
I don't have access to the `Settings` page for this repo, but with this update, we'd be able to change the branch protection for `master` so that only `build` and `test-local` are required for the PRs to go green.

LarsDenBakker:
Looks good to me, when will we run the slow lane then?

Westbrook:
Right now I’m always running the slow lane. It’s not taking that much longer than the fast lane anyways, the idea is you wouldn’t block PRs on the slow lane, just the fast one.

Who has access to tweak the branch rules?

LarsDenBakker:
ah, that makes perfect sense! I think only @daKmoR has this access on the repository

Westbrook:
@daKmoR, would love to chat through any issues you might have here so we can get the branch settings updated to let this go green and move forward with it. Let me know what works best for you!

.circleci/config.yml
```diff
@@ -40,23 +31,48 @@ jobs:
       # auto build typescript definition files (not part of build as locally JsDoc alone is all we need)
       - run: npm run build:types
 
+      # persist built files and duplicate dependencies to workspace
+      - persist_to_workspace:
+          root: ~/
+          paths:
+              - repo/packages/*/node_modules
+              - repo/packages/*/dist
```
daKmoR: should we add `- repo/node_modules` as well?

.circleci/config.yml
```diff
@@ -73,10 +89,17 @@ workflows:
...
   test-deploy:
     jobs:
-      - test
+      - build
+      - test-local:
+          requires:
+            - build
+      - test-bs:
+          requires:
+            - build
```
daKmoR: I think we should add `- test-local` here as well?

e.g. we only wanna bother browserstack once the local test succeeded right?

.circleci/config.yml
Westbrook: We could do a couple of test builds to compare performance, but right now they are being added to the cache, which can persist longer (as long as yarn.lock doesn’t change) but can’t be targeted by glob syntax. Being `node_modules` can be rather large (yay, modern JS) I’ve seen the workspace persistence go slower than caching on that number of files, but not in a scientifically proven way. I’ll check it out more closely.

.circleci/config.yml
Westbrook: I could go both ways. If we don’t block on the bs tests then it should be fine. I _do_ like that running them in parallel like this gets the whole build to less than 5 minutes pretty regularly. Confirm you want to run synchronized and I’ll update.

daKmoR: true let's try parallel and if at any point we want to reduce the load on browser stack we can easily do so 🤗

daKmoR:
I think that is good to go 👍

can do one final rebase and then we can merge 👍
I will adjust the merge requirements right after 🤗

Westbrook:
@daKmoR back up-to-date with `master`, even with Lars' approval, I don't think the repo will give me permission to merge with the branch rules not being met...

Westbrook:
@daKmoR with so much great work happening against the repo, it's tough to keep up-to-date with `master`. PTAL.

daKmoR:
it's merged now 👍 thxxxx 🤗

PR #1250, opened 2020-01-17
Fix incorrect links in README

gmlnchv:
Not sure if/how these links worked before but thinking they should be linked to (Open WC)[https://open-wc.org]


LarsDenBakker:
Thanks! These links work fine on our website (https://open-wc.org) but not on github/npm. With your changes it works on all environment, except on feature branches when we deploy them. That's fine with me, what do you think @daKmoR ? @thepassle ?

daKmoR:
sounds good 🤗 

thank you 🤗

PR #1251, opened 2020-01-17
Error in name of element

davie-robertson:
The custom element is news-app, not todo-app.


LarsDenBakker:
Thanks for the PR! 

We follow a specific convention for commits, I updated the commit message when merging so no worries about that! :)

PR #1254, opened 2020-01-17
docs: Add ZingGrid Webcomponent

zingsoftadmin:
Add the ZingGrid web component, A fully-featured, native, web-component, data table and data grid for Javascript applications.

### DIFF
--- a/docs/guide/component-libraries.md
+++ b/docs/guide/component-libraries.md
@@ -69,6 +69,10 @@ The hardest part of any project is often getting content onto that first blank p
 - [Wired Elements](https://wiredjs.com/)
 
   A set of common UI elements with a hand-drawn, sketchy look. These can be used for wireframes, mockups, or just the fun hand-drawn look.
+  
+- [ZingGrid](https://www.zinggrid.com)
+
+  A fully-featured, native, web-component, data table and data grid for Javascript applications.
 
 ::: tip Don't see your design system here?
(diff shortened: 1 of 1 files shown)


thepassle:
Seems like prettier isn't happy, could you try running prettier locally?

```
Checking formatting...
docs/guide/component-libraries.md
Code style issues found in the above file(s). Forgot to run Prettier?
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @open-wc/root@ lint:prettier: `prettier "**/*.{js,md}" "**/package.json" --check`
```

Westbrook:
Is ZingGrid part of a larger library of elements that you’d like to have listed here. I don’t think that this page should serve as a catalog of individual elements. Particularly, this format already suffers under the weight of listing all of the truly wonderful libraries listed already and moving to listing individual elements would quickly bring the benefits it brings to users to an end.

@zingsoftadmin have you taken a look at our catalog demo? Ensuring standalone elements are being distributed in such a way as to be included there would likely be more useful to you and any potential users of said element(s).

damntrecky:
@Westbrook We are only a single web component, we are not a library of components. I think we should be on the list because we are a commercial component. Free to use but the idea of a viable software as a web component is a huge step for web components. We will continue to maintain ourselves as a single product and not an entity of web components.

With that said I think the catalog demo would be nice but it's still in development.

damntrecky:
@thepassle I ran `npm run lint`  locally and all checks passed

Screen Recording 2020-01-22 at 02 20 PM

Issue #758, opened 2019-08-22
[es-dev-server] regular scripts are not compiled


LarsDenBakker:
We only compile modules with babel, if you have regular scripts they are not handled. This is currently by design, but it can break some use cases.

LarsDenBakker:
I'm working on this in a larger refactoring of compatibility mode.

e111077:
+1 on this btw. Loving `es-dev-server` so far with our mwc elements and we're using it with our glitch demo, but this is the only issue we're running into as demos tend to have inline scrips where we instinctively use es6

LarsDenBakker:
@e111077 what would you expect here in terms of behavior of polyfills and babel helpers. We load module scripts after the polyfills, and can inject babel helper imports. We could delay scripts until after polyfills, but people might rely on scripts to run at a specific time in the HTML.

e111077:
I'm sorry I'm not too familiar with the polyfill and babel injection order. Is it currently:

1. non-module scripts
2. polyfills and helpers
3. module scripts

and you're suggesting

1. polyfills and helpers
2. non-module scripts
3. modules scripts

If so, I think your suggestion would be what I would expect. If someone needs specific handling of script timing, I would provide a way for the user to disable non-transform-related polyfill injection and do

1. babel / transform helpers
2. user's non-module scripts (they must provide wc polyfills etc.)
3. module scripts

I think providing the polyfills and libraries are helpful, but sometimes a low-level escape hatch is needed. That was a common problem we ran into when we created Webcomponent Tester which provided every library you would need like sinon and lodash by default. It was useful in most situations, but made it a hassle for intricate cases.

LarsDenBakker:
@e111077 should work now with the latest es-dev-server

e111077:
Ace job! thanks!

Issue #1248, opened 2020-01-16
[es-dev-server] polyfills are not always being loaded


tmsns:
When es-dev-server is adding its polyfills (eg. for IE) the script urls for the polyfills are not always resolving. This is happening when following two conditions match:

- when no `<base />` tag to the root of the project is used in the appIndex
- when a page in a sub path is accessed (ie. http://localhost:8000/sub/sub)

I'll see if I can make a PR for this. 😃

tmsns:
I see two solutions here:
- make sure es-dev-server can resolve polyfills request in sub paths by changing [transform-index-html.js#L47](https://github.com/open-wc/open-wc/blob/master/packages/es-dev-server/src/middleware/transform-index-html.js#L47)
- change the references for the script tags [create-index-html.js#L127](https://github.com/open-wc/open-wc/blob/master/packages/building-utils/index-html/create-index-html.js#L127) and the loader script [loader-script.js#L125](https://github.com/open-wc/open-wc/blob/master/packages/building-utils/index-html/loader-script.js#L125). But as this is `building-utils`, it also affects the build version. And over there, it sometimes needs to be relative to the base path. So, we could pass the root path for the polyfills from dev server to build-utils, but that seems overkill.

What do you think @LarsDenBakker?

PS: while looking through the code I notice that by default we add polyfills for fetch and webcomponents by default. As es-dev-server is, afaik, more targeted for using regular (hence the name), shouldn't those polyfills be optional? Or maybe having a system to hook in other polyfills as well?

LarsDenBakker:
I think for es-dev-server we will need to use absolute paths to make this work. I have a bit refactor of the way we generate polyfills for the index.html on a branch, probably that's the best place to implement it.

tmsns:
> I think for es-dev-server we will need to use absolute paths to make this work. I have a bit refactor of the way we generate polyfills for the index.html on a branch, probably that's the best place to implement it.

I guess you mean server-relative paths instead of pure relative paths? This will have the same effect as absolute paths, but without the host. And I don't think you specifically need the host, right?

So based on your response, I think you want the something like the second solution, changing the reference (and probably passing a root to build-utils, if build-utils is still involved in the generation of polyfills in the new branch)

LarsDenBakker:
Yes that's the one. Started last year, been working on it on and off.

[...]

PR #1261, opened 2020-01-18
feat(polyfills-loader): create a separate polyfills loader

LarsDenBakker:
This adds a new package called `polyfills-loader` which extracts the functionality from `@open-wc/building-utils` and makes it available as a standalone package. This makes it easier for people to understand what's going on behind the scenes, and to document how to extend the logic. 

In this PR it's implemented in es-dev-server, and by extension in karma-esm since it's based on es-dev-server. Rollup and webpack will be updated separately. 

Visible benefits:
- Polyfills loader behavior can be configured, the docs have been updated to reflect this.
- Non-module scripts are now included in the polyfills loader logic, making them work on older browsers too. Fixes https://github.com/open-wc/open-wc/issues/758 
- Inline scripts aren't made into external scripts unless absolutely necessary, this especially visible when using `auto` mode on modern browsers. Currently it always has to inject the polyfills loader to handle resolving imports for inline modules. With this change, the imports are resolved in the HTML itself. This makes the `es-dev-server` less intrusive in the most common scenarios.
- No more hashes in test snapshots :)


packages/polyfills-loader/TODO.md
```diff
@@ -0,0 +1,7 @@
+Polyfills to add:
```
LarsDenBakker: need to remove this todo

LarsDenBakker:
Because we are now resolving imports in inline scripts, we need to make imports absolute from the root dir otherwise imports will be resolved incorrectly when there is no `<base href>` (related: https://github.com/open-wc/open-wc/issues/1248)

LarsDenBakker:
I've published this PR under `npm i --save es-dev-server@next`

packages/polyfills-loader/src/create-polyfills-data.js
```diff
@@ -0,0 +1,199 @@
...
+    if (polyfillConfig.minify) {
+      const minifyResult = Terser.minify(content, { sourceMap: false });
+      // @ts-ignore
+      content = minifyResult.code;
+    }
+
+    const filePath = `${path.join(
+      cfg.polyfillsDir || 'polyfills',
+      `${polyfillConfig.name}${polyfills.hash !== false ? `.${createContentHash(content)}` : ''}`,
+    )}.js`;
```
tmsns: This will only work for Linux. To support Windows, just use: `path.join(..).replace('\\', '/')`.

LarsDenBakker: I do a `toBrowserPath` later on, but not in all locations. I'll move that logic here.

PR #1265, opened 2020-01-19
feat(lit-helpers): add ReadOnlyPropertiesMixin

bennypowers:
allows `readOnly: true` in property declarations, and adds 
`set(properties: Object<string, any>): Promise<this>` to instances

This PR is based on work by @morbidick in https://github.com/morbidick/lit-element-notify/blob/master/notify.js


packages/lit-helpers/README.md
```diff
@@ -120,3 +120,86 @@ html`
     }
   }
 </script>
+
+## Privately Settable Read-Only Properties
+
+The classic Polymer library allowed defining properties that were read-only, but privately settable. By adding `readOnly: true` to the property declaration, the property could be set with a `_setPropName` method.
```
LarsDenBakker: Not many people will understand this reference, maybe we should just document the feature as it is?

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,80 @@
...
+      this.setPropEntry = this.setPropEntry.bind(this);
+    }
+
+    /**
+     * Set read-only properties
+     * @param  {Object<string, unknown>}  props
+     * @return {Promise<this>}
+     * @private
+     */
+    async set(props) {
```
LarsDenBakker: `set` is a very generic name, perhaps we can think of something more appropriate to this mixin?

Also is setting multiple properties the common case, or should we make this just key value?

bennypowers: maybe `setReadOnlyProperties` would be better.

bennypowers: A unary `set` which takes an object mapping is consistent with `React.Component#setState`, which is proof enough that the idiom is popular and well-understood. 

In the project that inspired this, seven out of ten calls to `set` used multiple props. If we didn't provide that here, users would be stuck writing boiler plate again. It's an easy abstraction for use to pick up, with negligible cost against the simple case of a single property updated. By providing an object mapping setter we cater to both case' needs

LarsDenBakker: Since you would need to provide the property name anyway that seems reasonable indeed

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Enables the nofity option for properties to fire change notification events
+ *
+ * @template TBase extends import('lit-element').LitElement
+ * @param {Constructor<TBase>} superclass
+ */
+export const ReadOnlyPropertiesMixin = superclass => {
+  /**
+   * @type {Map<string, symbol>}
```
LarsDenBakker:
```suggestion
   * @type {Map<string, Symbol>}
```

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
...
+   * @private
+   */
+  const _readOnlyPropertyNamesMap = new Map();
```
LarsDenBakker:
```suggestion
  const readOnlyPropertyNamesMap = new Map();
```

No need for a _ if it's not a  class member.

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
+/** @typedef {import('lit-element').PropertyDeclaration & { readOnly: Boolean }} PropertyDeclarationWithReadOnly */
```
LarsDenBakker:
```suggestion
/** @typedef {import('lit-element').PropertyDeclaration & { readOnly: Boolean }} ReadOnlyPropertyDeclaration */
```

Not a strong preference, but seems to make sense to me.

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
...
+   * @private
+   */
+  const _readOnlyPropertyNamesMap = new Map();
+
+  // @ts-ignore
+  return class ReadOnlyPropertiesClass extends superclass {
+    /**
+     * @inheritdoc
+     * @param  {string} name property name
+     * @param  {PropertyDeclarationWithReadOnly} options augmented property declaration with optional `readOnly` boolean.
```
LarsDenBakker:
```suggestion
     * @param  {PropertyDeclarationWithReadOnly | PropertyDeclaration} options augmented property declaration with optional `readOnly` boolean.
```

You will need to typedef the `PropertyDeclaration` above.

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * @param {[string, unknown]} entry
+     * @return {Promise<unknown>}
+     * @private
+     */
+    _setPropEntry([name, newVal]) {
+      // typescript... https://github.com/microsoft/TypeScript/issues/1863
+      /** @type {any} */
```
LarsDenBakker:
```suggestion
      /** @type {unknown} */
```

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
...
+    async setReadOnlyProperties(properties) {
+      await Promise.all(Object.entries(properties).map(this._setPropEntry));
+    }
+
+    /**
+     * @param {[string, unknown]} entry
+     * @return {Promise<unknown>}
+     * @private
+     */
+    _setPropEntry([name, newVal]) {
```
LarsDenBakker: Perhaps inline this into `setReadOnlyProperties`? Reduces the footprint of the mixin on the class.

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,90 @@
+/** @typedef {import('lit-element').PropertyDeclaration & { readOnly: Boolean }} PropertyDeclarationWithReadOnly */
+
+/**
+ * @template TBase
+ * @typedef {new (...args: any[]) => TBase} Constructor
+ */
+
+/**
+ * Enables the nofity option for properties to fire change notification events
```
LarsDenBakker:
```suggestion
 * Enables the notify option for properties to fire change notification events
```

packages/lit-helpers/src/read-only-properties.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  interface PropertyDeclaration {
+    readOnly?: boolean;
+  }
+}
+
+export interface ReadOnlyClass {
+  /**
+   * Set read-only properties
+   */
+  setReadOnlyProperties(props: {
```
LarsDenBakker: Shoud this be a public method?

bennypowers: Should probably be protected, but ts doesn't support that yet, so the best we could do would be a jsdoc `@protected` tag

packages/lit-helpers/test/read-only-properties-mixin.test.js
```diff
@@ -0,0 +1,39 @@
...
+    );
+    element = await fixture(html`<${ReadOnlyElement}></${ReadOnlyElement}>`);
+  });
+
+  it('defines a read only property', async () => {
+    const initial = element.readOnlyString;
+    element.readOnlyString = Date.now().toString(36);
+    expect(element.readOnlyString).to.equal(initial);
+  });
+
```
daKmoR: what happens if you do `element.readOnlyString = 'foo'`? will it throw? it probably should?

in any way can we test what it does?

packages/lit-helpers/src/read-only-properties-mixin.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function ReadOnlyPropertiesMixin(superclass) {
```
daKmoR: not deduped? what happens if it is applied multiple times?

we have a limitation that it can only be at the root level.. but what happens if you apply it later again? do we wanna test what happens? or prevent it with dedupeMixin?

packages/lit-helpers/src/read-only-properties.d.ts
LarsDenBakker: Do we want to follow the convention of protected with a leading `_` ?

packages/lit-helpers/test/read-only-properties-mixin.test.js
bennypowers: Silent noop.

packages/lit-helpers/src/read-only-properties-mixin.js
bennypowers: We can dedupe.

Base class requirement is a bug not a feature, a later pr should fix that one

packages/lit-helpers/src/read-only-properties.d.ts
bennypowers: None of LitElement's protected methods are prefixed so I'd prefer not to here

LarsDenBakker: hmmm, 50/50 on this one so I'll defer to your preference :)

bennypowers:
@daKmoR any objections?

packages/lit-helpers/src/read-only-properties-mixin.js
manolakis: I'm not understanding this point. Why you say this is a bug or even a limitation? I understand that you should apply the mixin in those classes or mixins in which you want to have read only properties, and that's why we need to dedupe it.

bennypowers: The limitation is the the mixin has to be applied to `LitElement`, it doesn't work if applied further up the chain

PR #1274, opened 2020-01-25
feat(demoing-storybook): update to storybook-prebuilt v1

LarsDenBakker:
Update to use the new storybook-prebuilt. See https://github.com/open-wc/storybook-prebuilt for updated docs.

This change keeps backwards compatibility by importing all addons. In the next breaking change, we will remove these imports and allow users to configure addons themselves.


LarsDenBakker:
We will need to do some extra work to support the addons in `main.js`, so that should definitely be in a separate PR. We could do some addon presets.

Issue #1059, opened 2019-11-28
[es-dev-server] Doesn't open index.html when run with app-index=index.html and url has params with a dot inside


doplumi:
Example: `http//localhost:8000/example?lat=1.0&lon=1.0` doesn't work.
For some reason `http//localhost:8000/?lat=1.0&lon=1.0` does.

LarsDenBakker:
I can't reproduce this issue, can you explain the commands you're using and the project structure?

RoCat:
Hello, I have the same problem. 
A page with path params having a dot inside cause the es-dev-server to redirect to 404:
`localhost:8888/issues?page=0&size=25&sort=&creationStart=2019-11-06T23:00:00.000Z` This page doesnt render and redirect to 404
`localhost:8888/issues?page=0&size=25&sort=&creationStart=2019-11-06T23:00:00` This one render properly.

Of course the builded app served by ngninx works with the dots.

here is a es-dev-server.config used

`const proxy = require('koa-proxies');

const proxyUrl = {
  dev: {
    api: 'XXXX',
    'private-api': 'XXXXX,
  },
};

let env = process.argv[2];
if (!Object.keys(proxyUrl).includes(env)) {
  env = 'dev';
}

module.exports = {
  port: 8888,
  watch: true,
  nodeResolve: true,
  preserveSymlinks: true,
  appIndex: 'index.html',
  moduleDirs: ['node_modules'],
  middlewares: [
    proxy('/api', {
      target: proxyUrl[env].api,
      changeOrigin: true,
      rewrite: path => path.replace(/\/api/, ''),
    }),
    proxy('/private-api', {
      target: proxyUrl[env]['private-api'],
      changeOrigin: true,
      rewrite: path => path.replace(/\/private-api/, ''),
    }),
  ],
};
`

jwehrman:
Noticed the same issue, the history-fallback has a pretty clear check for the . character.
https://github.com/open-wc/open-wc/blob/master/packages/es-dev-server/src/middleware/history-api-fallback.js#L16

Wonder if that is how it determines the request may be to a file, as if the dot means a file extension?

jwehrman:
created a pull request with a unit test to show how to reproduce and a possible solution.

LarsDenBakker:
Fixed with https://github.com/open-wc/open-wc/pull/1278#issuecomment-578933313

Sorry it took so long for something like this

PR #1278, opened 2020-01-27
fix(es-dev-server): handle url parameters with . characters (#1059)

jwehrman:
Validate only the url path, not the querystring, for . characters
to detect file paths.


LarsDenBakker:
Thanks! We just need the CLA to be signed now :)

ptc-jwehrman:
All set, figured out the confusion with my github accounts.

LarsDenBakker:
🥳

PR #809, opened 2019-09-18
WIP: feat(building-utils): adds support for the custom-styles polyfill

jrobinson01:
Adds support for the custom-styles-interface polyfill. The polyfill enables using a `<custom-style>` element wrapped around an arbitrary `<style>` element. This is usually used to provide a theme for your app, by way of css variables. This is needed to support css variables in IE11.

PR #1291, opened 2020-01-31
feat(building-utils): shady css scoped polyfill (v2)

jrobinson01:
picks up where https://github.com/open-wc/open-wc/pull/809 left off. The only change between this branch and the prior branch is that the shady-css-scoped-element is published as an independent package instead of by webcomponents/polyfills.


jrobinson01:
closing. replaced by this PR: https://github.com/open-wc/open-wc/pull/1292

PR #1292, opened 2020-01-31
feat(building-utils): shady css scoped polyfill (v3)

jrobinson01:
follow-up to #809


packages/building-utils/index-html/create-index-html.js
```diff
@@ -51,6 +51,7 @@ const { createContentHash, cleanImportPath, polyfillFilename } = require('./util
  * @property {boolean} [coreJs] whether to polyfill core-js polyfills
  * @property {boolean | string} [regeneratorRuntime] whether to add regenerator runtime
  * @property {boolean} [webcomponents] whether to polyfill webcomponents
+ * @property {boolean} [shadyCSS] whether to polyfill shadyCSS
```
LarsDenBakker: Shady CSS is loaded as part of `webcomponentjs`, right? This is only adding the custom style interface? So maybe we should call it `shadyCssCustomStyle`?

LarsDenBakker:
Sorry for the poor timing, but we separated the polyfill loading logic to a separate package: https://github.com/open-wc/open-wc/tree/master/packages/polyfills-loader 

This is only used by es-dev-server right now, but we will migrate rollup and webpack to it later as well. Would you be able to make the same changes there? The setup is pretty similar.

packages/building-utils/index-html/create-index-html.js
jrobinson01: Good point. It gets confusing fast with the separation of all the poly fills, but you're right, custom-style-interface is what we want FROM shadyCSS (I don't understand why it's a separate thing not loaded by webcomponentsjs?). https://github.com/open-wc/open-wc/pull/1292/files/d36a90f92959dae3474ef3b6cb92610bbe595159#diff-0a09ec55ca2108cb9656fd314415a8e8R187

LarsDenBakker: I *think* shady css is loaded for polyfilling styles in shadow dom, but only the parts that are strictly necessary by default.

packages/polyfills-loader/src/create-polyfills-data.js
```diff
@@ -157,14 +185,15 @@ function createPolyfillsData(cfg) {
     if (!polyfillConfig.name || !polyfillConfig.path) {
       throw new Error(`A polyfill should have a name and a path property.`);
     }
-
-    const codePath = path.resolve(polyfillConfig.path);
-    if (!codePath || !fs.existsSync(codePath) || !fs.statSync(codePath).isFile()) {
-      throw new Error(`Could not find a file at ${polyfillConfig.path}`);
+    let content = '';
+    if (Array.isArray(polyfillConfig.path)) {
+      content = polyfillConfig.path.reduce((acc, p) => {
```
LarsDenBakker: This is more readable as a map + join

PR #1302, opened 2020-02-02
Feat/storybook dynamic stories

### LINKED ISSUE #1179: [demoing-storybook] tracking adding/removing/renaming files
When using storybook with webpack, it watches the filesystem for added stories and reloads the browser when they are added. We should add the same capability to `@demoing/storybook`

### LINKED ISSUE #1275: [demoing-storybook] don't use dynamic imports for stories
When we generate the `preview.js` file, we use dynamic imports to import all the stories. This is fine for development, but it causes rollup to create a lot of separate chunks. We should use static imports here.

LarsDenBakker:
This changes looking for stories dynamic, doing a lookup for stories each time iframe.html is served. This way when you refresh, you always get the latest stories which are added/removed/renamed.

It also makes es-dev-server trigger a reload on file names, and makes the story imports into static imports to prevent rollup from creating many separate chunks.

Fixes #1275
Fixes #1179

### DIFF
--- a/packages/demoing-storybook/src/build/build.js
+++ b/packages/demoing-storybook/src/build/build.js
@@ -10,6 +10,7 @@ const { createOrderedExports } = require('../shared/createOrderedExports');
 const createAssets = require('../shared/getAssets');
 const listFiles = require('../shared/listFiles');
 const toBrowserPath = require('../shared/toBrowserPath');
+const { injectStories } = require('../shared/injectStories');
 
 const injectOrderedExportsPlugin = storyFiles => ({
   async transform(code, id) {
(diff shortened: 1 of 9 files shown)


packages/demoing-storybook/src/build/build.js
```diff
@@ -151,12 +152,26 @@ async function buildManager({ outputDir, assets }) {
   await rollupBuild(configs[1]);
 }
 
-async function buildPreview({ outputDir, assets, storyFiles, rollupConfigDecorator }) {
+async function buildPreview({
+  outputDir,
+  assets,
```
bennypowers: nit: looks like we only take the `iframeHTML`. Can we destructure here in the params?

packages/demoing-storybook/src/start/transformers/createServeStorybookTransformer.js
```diff
@@ -1,15 +1,53 @@
...
+const { injectStories } = require('../../shared/injectStories');
+const { createOrderedExports } = require('../../shared/createOrderedExports');
+
+/**
+ * @param {object} args
+ * @param {object} args.assets
+ * @param {string} args.assets.indexHTML
+ * @param {string} args.assets.iframeHTML
+ * @param {string} args.previewImport
+ * @param {string[]} args.storiesPatterns
```
bennypowers: would have liked to see more docs
```suggestion
 * @param {string[]} args.storiesPatterns glob patterns of story files
```

packages/demoing-storybook/src/start/transformers/createServeStorybookTransformer.js
```diff
@@ -1,15 +1,53 @@
...
+ * @param {string} args.rootDir
+ */
+function createServeStorybookTransformer({ assets, previewImport, storiesPatterns, rootDir }) {
+  const servedStoryUrls = new Set();
+  const { indexHTML, iframeHTML } = assets;
```
bennypowers: nit: can we destructure this in params?

LarsDenBakker: I don't like it when a function declaration goes multiline, it's hard to see where things start and end.

packages/demoing-storybook/src/build/build.js
LarsDenBakker: I think deeply nested destructuring is not very readable :)

LarsDenBakker: In this function the parameters are already multiline, so I guess it's fine here. added it :)

Issue #1303, opened 2020-02-02
[Codelabs intermediate] Google news API alternative


tkgroot:
Is there a reasonable reason why the codelabs is using the Google news API? 
Wouldnt it be better to use open resources instead? The Hackernews API is publically available, hence no registration step required. It reduces complexity (yet another registration form with captcha), completely omits the construction of the url with the API key and doesnt interrupt the flow of the codelabs.

I did implement the codelabs with the Hackernews API instead. With some minor modifications (see code below), it has the same outcome, no real value though (just a bunch of comments to an "unknown" topic).

```javascript
async fetchArticles() {
          this.loading = true
          let idx = 0
          const array = []
          const path = "https://hacker-news.firebaseio.com/v0/item"
          const maxItem = 22115410

          while (idx < 5) {
            await fetch(`${path}/${maxItem - idx}.json`)
            .then(res => res.json())
            .then(data => array.push(data))  
            
            idx++
          } 
          
          this.articles = array
          this.loading = false
        }
```

LarsDenBakker:
The registration for the google news API is a bit annoying indeed. I think the hackernews API is a good alternative, I don't mind switching over to that. Open for PRs :)

LarsDenBakker:
This was implemented.

PR #1310, opened 2020-02-04
Feature hackernews api

tkgroot:
reference issue #1303 
- changed the intermediate codelabs to use the hackernews api instead of the google news api
- change occurances of articles -> comments, article -> comment
- change the fetchArticles function and rename it to fetchComments
- add additional api information to step 3
- add additional tip to the step 4

- create npm run-scripts for development on codelabs
- fix codelabs readme generated output path
- add codelabs readme additional npm start script option

I used only the comments from the HackerNews api since they are the most frequently and thus not that error-prune (the hackernews api can be buggy at times tbh). On the downside they dont carry that much information in terms of object properties. I tried to touch as little as possible of the current available codelab and just made changes where there were necessary.
Let me know if the changes are inline with you ideas of the codelab.

Additionally to that i created npm run-scripts to make life easier during developing the codelabs. They include a nodemon watching for file changes and execute the buildCodelab.js and browser-sync to serve the static generated content. I dont know, since its not part of the issue referenced if this is something you want, and if so if you want that combined with this PR. Here the same applies as above: let me know.

### DIFF
--- a/packages/codelabs/package.json
+++ b/packages/codelabs/package.json
@@ -26,8 +26,11 @@
     "@types/cheerio": "^0.22.13",
     "@types/fs-extra": "^8.0.1",
     "@types/marked": "^0.6.5",
+    "browser-sync": "^2.26.7",
     "cheerio": "^1.0.0-rc.3",
+    "concurrently": "^5.1.0",
     "fs-extra": "^8.1.0",
-    "marked": "^0.7.0"
+    "marked": "^0.7.0",
+    "nodemon": "^2.0.2"
   }
 }
--- a/packages/codelabs/package.json
+++ b/packages/codelabs/package.json
@@ -15,7 +15,10 @@
   "author": "open-wc",
   "homepage": "https://github.com/open-wc/open-wc/",
   "scripts": {
-    "site:build": "node build-codelabs.js"
+    "start": "concurrently npm:site:build:watch npm:site:build:serve",
+    "site:build": "node build-codelabs.js",
+    "site:build:watch": "nodemon -w './src/**/*.*' -e 'md' --exec 'npm run site:build'",
+    "site:build:serve": "browser-sync start -s '../../_site/codelabs/' -w '../../_site/**/*.*'"
   },
   "keywords": [
     "open-wc",
--- a/packages/codelabs/src/intermediate/lit-html.md
+++ b/packages/codelabs/src/intermediate/lit-html.md
@@ -39,7 +39,7 @@ At the bottom of each section, there is a "View final result" button, this will 
 
 ## Setup
 
-In this codelab, we will build a news app. This is a great exercise to learn the intermediate parts of lit-html and lit-element.
+In this codelab, we will build a hackernews app. This is a great exercise to learn the intermediate parts of lit-html and lit-element.
 
 You can follow this codelab using anything that can display a simple HTML page. For the best editing experience, we recommend setting this up using your favorite IDE. Alternatively, you can use an online code editor like [jsbin](https://jsbin.com/?html,output), [stackblitz](https://stackblitz.com/) or [webcomponents.dev](https://webcomponents.dev/).
 
@@ -63,7 +63,7 @@ Make sure that you add `type="module"` to the script tag.
 In this example, we are using `unpkg`, a CDN from which we can easily import any modules that are available on NPM. When working on a real project, it is a good idea to use an actual package manager such as NPM or yarn.
 </aside>
 
-You should already know how to create a web component using `LitElement`. Go ahead and create one which renders '_My news app_' to the screen. When it works, you're ready to move on to the next step.
+You should already know how to create a web component using `LitElement`. Go ahead and create one which renders '_My hackernews app_' to the screen. When it works, you're ready to move on to the next step.
 
 ---
 
@@ -74,20 +74,20 @@ You should already know how to create a web component using `LitElement`. Go ahe
 <!DOCTYPE html>
 <html>
   <body>
-    <news-app></news-app>
+    <hackernews-app></hackernews-app>
 
     <script type="module">
       import { LitElement, html } from 'https://unpkg.com/lit-element?module';
 
-      class NewsApp extends LitElement {
+      class HackerNewsApp extends LitElement {
         render() {
           return html`
-            My news app
+            My hackernews app
           `;
         }
       }
 
-      customElements.define('news-app', NewsApp);
+      customElements.define('hackernews-app', HackerNewsApp);
     </script>
   </body>
 </html>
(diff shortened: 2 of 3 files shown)


packages/codelabs/src/intermediate/lit-html.md
```diff
@@ -39,7 +39,7 @@ At the bottom of each section, there is a "View final result" button, this will
 
 ## Setup
 
-In this codelab, we will build a news app. This is a great exercise to learn the intermediate parts of lit-html and lit-element.
+In this codelab, we will build a comment overview app. This is a great exercise to learn the intermediate parts of lit-html and lit-element.
```
thepassle: Hackernews app?

packages/codelabs/src/intermediate/lit-html.md
```diff
@@ -74,79 +74,106 @@ You should already know how to create a web component using `LitElement`. Go ahe
...
+      customElements.define('hn-comment-app', HackerNewsCommentApp);
     </script>
   </body>
 </html>
 ```
 
 </details>
 
-## Google news API
+## External API
```
thepassle: Hackernews API?

packages/codelabs/src/intermediate/lit-html.md
```diff
@@ -74,79 +74,106 @@ You should already know how to create a web component using `LitElement`. Go ahe
...
 
-For this codelab, we will be fetching live data from the google news API. You will need to retrieve an API key first:
+For this codelab, we will be fetching live data from the HackerNews API. To retrieve an
```
thepassle: Why is there a hard line break here? Its in some other places too

packages/codelabs/src/intermediate/lit-html.md
```diff
@@ -74,79 +74,106 @@ You should already know how to create a web component using `LitElement`. Go ahe
...
+</aside>
 
 ### Tips
 
+The HN API doesn't distinguish its items, thus it might be difficult to know if a
+received object is indeed a comment. To verify that the data fetched from the HackerNews
+API is a comment you can simply use an if-conditional and check against its data.type.
+
+```js
+if(data.type !== "comment") return false;
```
thepassle: Can we add more code context around this code snippet? Where do people add/use this code?

thepassle:
Intuitively, to me, it feels a bit weird to fetch comments, can't we fetch for threads instead? Im not so sure about that while loop, too

I feel like it makes more sense to have an overview of hackernews _threads_ that you can then filter on which ones you've read and which ones you havent. We can omit loading/displaying the actual comments of a thread, thats fine. If we list threads instead of comments, we can probably get rid of the while loop, too.

packages/codelabs/src/intermediate/lit-html.md
```diff
@@ -1070,36 +1221,49 @@ html`
...
-          this.articles = jsonResponse.articles;
+          const path = "https://hacker-news.firebaseio.com/v0/item";
+          const itemId = 22115410;
+
+          let idx = 0;
+          const maxItems = 5;
+          const array = [];
+
+          while (array.length < maxItems) {
+            await fetch(`${path}/${itemId - idx}.json`)
```
thepassle: I'd prefer to stay consistent and either use async await everywhere, or use .thens, so we should probably go with async await

packages/codelabs/src/intermediate/lit-html.md
tkgroot: yes that´s better. i update this

packages/codelabs/src/intermediate/lit-html.md
tkgroot: I break my code around like 80, just a habbit i started. I will update these lines to not wrap.

packages/codelabs/src/intermediate/lit-html.md
tkgroot: consistency is better i agree. I will update it to the async/await schema

tkgroot:
> Intuitively, to me, it feels a bit weird to fetch comments, can't we fetch for threads instead? Im not so sure about that while loop, too

That´s the things about the hackernews api. It doesnt support fetching threads. One could change the item type from comments to stories. There you would have a title and text property available. So it might be worth to read/unread the stories instead of the comments. 
But for that to work one must create more requests to the hackernews api, since stories are less frequent then comments. So simply querying /api/v0/items/<id> doesn't work as good anymore. Or query a different api to get the frequent story ids, but not their content. This approach requires a lot more logic around the fetching of data, even fetching two different apis.

> I feel like it makes more sense to have an overview of hackernews _threads_ that you can then filter on which ones you've read and which ones you havent. We can omit loading/displaying the actual comments of a thread, thats fine. If we list threads instead of comments, we can probably get rid of the while loop, too.

We cannot get rid of the while loop im afraid. The hackernews api doesnt offer any bulk get of stories, or any other type of that matter.

packages/codelabs/src/intermediate/lit-html.md
tkgroot: Yes i can add more context. Stay tuned

thepassle:
What I like about the API we used before is that its conceptually simple; Its a news app, and it fetches news articles. The code for fetching that data is simple, too. The only 'burden' is getting an API key, which imo is not that much work.

So at this point in time I'm not entirely convinced this makes things much more clear. 

@LarsDenBakker what do you think?

tkgroot:
I can also offer to use a different api, which is public and the query would be as simple as the google news api. It would list public avaiable apis, with title, description, auth, https, cors and category.
If that is something you are more comfortable with. As always let me know, im here to help not to make it worst 😄

packages/codelabs/package.json
```diff
@@ -15,7 +15,10 @@
   "author": "open-wc",
   "homepage": "https://github.com/open-wc/open-wc/",
   "scripts": {
-    "site:build": "node build-codelabs.js"
+    "start": "concurrently npm:site:build:watch npm:site:build:serve",
+    "site:build": "node build-codelabs.js",
+    "site:build:watch": "nodemon -w './src/**/*.*' -e 'md' --exec 'npm run site:build'",
+    "site:build:serve": "browser-sync start -s '../../_site/codelabs/' -w '../../_site/**/*.*'"
```
LarsDenBakker: adding dependencies to the monorepo is expensive, we don't need to use browser-sync since we already have es-dev-server with `--watch` available

tkgroot: okay i will remove this and use it for my local development only.

LarsDenBakker:
I agree with @thepassle that the API exposed by hackernews isn't great, ideally it's just one request to get the data you need without additional checks and mappings.

thepassle:
> As always let me know, im here to help not to make it worst 😄

I know, and sorry if my comment came across as harsh, I really do appreciate you taking the time and effort to contribute 😊 

Also, to give a little bit more context, we've given this codelab as a workshop before for quite some people, not all of whom are mainly frontend developers. Many of them didnt even know about the fetch api, so I really feel we should keep that part of the code as close to (if not exactly the same) as we have it right now; just a simple fetch call.

I'd be open to hear your thoughts about a different API you mentioned though, which API specifically did you have in mind?

tkgroot:
> I know, and sorry if my comment came across as harsh, I really do appreciate you taking the time and effort to contribute 😊
> 
> Also, to give a little bit more context, we've given this codelab as a workshop before for quite some people, not all of whom are mainly frontend developers. Many of them didnt even know about the fetch api, so I really feel we should keep that part of the code as close to (if not exactly the same) as we have it right now; just a simple fetch call.
> 
> I'd be open to hear your thoughts about a different API you mentioned though, which API specifically did you have in mind?

Its fine. I myself found the HackerNews API not really feasable for many things, so i totally understand the concerns you raised.

I have looked for an public api and i found the following, which i thing would be way better than the hackernews api and very close to the google news api (simple and bulk gettable). With [Open Brewery DB](https://www.openbrewerydb.org/documentation/01-listbreweries) you can list breweries, which is also very nice for tutorials and explainable i think.

thepassle:
The brewery API looks fine to me. The 'app' could track wether you visited a brewery or not, so instead of 'read' and 'unread' it could be 'visited' and 'unvisited'

tkgroot:
Okay, then i will close this feature since its not an hackernews implementation anymore and create a new pull request with the open brewery api.

daKmoR:
nice 👍  looking forward to this brewery api pull request 🤗

PR #1324, opened 2020-02-08
docs: add adopted stylesheets workaround to FAQ

Westbrook:


thepassle:
Text LGTM, looks like prettier isnt happy though

Westbrook:
Yeah. Wrote it on GitHub as I’m traveling a bit. Will clean up later.

docs/faq/README.md
```diff
@@ -12,10 +12,75 @@ In this section you can find answers to frequently asked questions regarding jav
...
+```
+
 ## Custom elements render life cycle
 
 <iframe src="https://andyogo.github.io/custom-element-reactions-diagram/" style="width: 940px; border: none;margin-left: -100px; height: 1350px;"></iframe>
 
+## Debugging styles in adopted stylesheets with Chrome Dev Tools
+
+If your custom element leverages the performance benefits of [constructable stylesheets](https://developers.google.com/web/updates/2019/02/constructable-stylesheets), you will likely run into [this Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=946975) that makes it difficult to debug those styles in Dev Tools. In order to make those styles accessible in Dev Tools again, you can prepend the following snippet to the beginning of your development environment to remove the `adoptedStyleSheets` API and have access to those styles in the inspector. When you find yourself needing this work around, it would be beneficial to the community for you to visit the linked Chrome bug and star it to help drive the priority of a fix in the Chrome team.
+
```
LarsDenBakker: Perhaps we can mention here that lit-element uses adopted stylesheets?

Westbrook: Added.

Westbrook:
Added some reference to what actually happens when using this snipper as per conversations on Slack. Having written this now while being in conversation around recent updates to `demoing-storybook`, I wonder if we should make it a default injection or option in our build there... 🤔 but, a conversation for another day.

PR #1327, opened 2020-02-09
Point package.json homepages to specific directories in monorepo

aomarks:
This way the "homepage" at at npmjs.com (e.g. https://www.npmjs.com/package/es-dev-server) will lead to the package-specific directory instead of the top-level monorepo. Useful for e.g. finding the CHANGELOG.


LarsDenBakker:
Makes sense, thanks! Commitlint didn't like the capital in the commit name, I fixed that when merging :)

Issue #1293, opened 2020-01-31
[scoped-elements] mangling elements class names makes dom tags unrecognizable


manolakis:
Often, class names are mangled in the bundling process. That makes the tags used when defining an item as `MyButton`, instead of having a meaningful name like` my-button`, to get a meaningless name like `c-a`,` c-b`, `c-c`, etc., which makes it very difficult to identify which element is used in the DOM.

LarsDenBakker:
We could use the keys for setting the name instead of the class.

manolakis:
Agree. I have a PR almost ready

manolakis:
Fixed by https://github.com/open-wc/open-wc/commit/b6667ef88c79e6440d5804874fbf641452f1a63e

Closing

PR #1329, opened 2020-02-10
fix(scoped-elements): unrecognizable tag names (#1293)

manolakis:
This commit uses the keys instead the class names for setting the tag names.


daKmoR:
niiicccceee 💪

PR #1335, opened 2020-02-12
ci: include node_modules in workspace

Westbrook:
🙈 Missed a dependency on the deploy step!

Running the CI now to see whether it looks better to have this in the workspace or to load the cache in the deploy step.


Westbrook:
@daKmoR perf on the CI for this looks good and should get runs of `deploy` against `master` back to green. Sorry for the miss here!

Issue #1319, opened 2020-02-07
resolve-module-imports give issue on multiline


Anubis012:
https://github.com/open-wc/open-wc/blob/9b36974bfc81bafb11373ae7bfd527da485063e3/packages/es-dev-server/src/utils/resolve-module-imports.js#L190

When you import some html with a long path prettier puts the string on a new line so it gives an issue within the es-dev-serv resolve-module-imports
example : 
works fine
{
    pattern: ':lang/:segment/borrowing/mortgage-early-repayment',
    // prettier-ignore
    import: () =>
      import('ing-orange-sf-mortgage-early-reimbursement/ing-orange-sf-mortgage-early-reimbursement.html'),
    render: { component: 'ing-orange-sf-mortgage-early-reimbursement' },
 }

but prettier doesn't like the long path and transforms it to : 
{
    pattern: ':lang/:segment/borrowing/mortgage-early-repayment',
    import: () =>
      import(
            'ing-orange-sf-mortgage-early-reimbursement/ing-orange-sf-mortgage-early-reimbursement.html'
      ),
    render: { component: 'ing-orange-sf-mortgage-early-reimbursement' },
 }

PR #1336, opened 2020-02-12
Fix/resolve module imports

Anubis012:
First fix for 
https://github.com/open-wc/open-wc/issues/1319

something like : 
`import(
        'module'
);`
will work now but when a litteral char is used in a comment it will still cause issues (didn't want to resort currently to regex for speed reasons).
Still failing example :
`import(
     /* a comment with a literal ' inside*/
     'module''
)`

### DIFF
--- a/packages/es-dev-server/src/utils/resolve-module-imports.js
+++ b/packages/es-dev-server/src/utils/resolve-module-imports.js
@@ -184,11 +184,16 @@ async function resolveModuleImportsWithConfig(importer, source, cfg) {
       lastIndex = end;
     } else if (dynamicImportIndex >= 0) {
       // dynamic import
-      const dynamicStart = start + 1;
-      const dynamicEnd = end - 1;
+      const subSource = source.substring(start, end);
+      let stringSymbol;
+      if (subSource.includes('`')) stringSymbol = '`';
+      if (subSource.includes('"')) stringSymbol = '"';
+      if (subSource.includes("'")) stringSymbol = "'";
+
+      const dynamicStart = source.indexOf(stringSymbol, start) + 1;
+      const dynamicEnd = source.indexOf(stringSymbol, dynamicStart);
 
       const [importee, importeeSuffix] = getImportee(source.substring(dynamicStart, dynamicEnd));
-      const stringSymbol = source[dynamicStart - 1];
       const isStringLiteral = [`\``, "'", '"'].includes(stringSymbol);
       const concatenatedString =
         stringSymbol === `\`` || importee.includes("'") || importee.includes('"');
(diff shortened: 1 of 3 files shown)


packages/es-dev-server/src/utils/resolve-module-imports.js
```diff
@@ -184,18 +184,17 @@ async function resolveModuleImportsWithConfig(importer, source, cfg) {
       lastIndex = end;
     } else if (dynamicImportIndex >= 0) {
       // dynamic import
-      const dynamicStart = start + 1;
-      const dynamicEnd = end - 1;
-
+      const literals = [`\``, "'", '"'];
+      const stringSymbol = literals.find(char => source.substring(start, end).includes(char));
```
LarsDenBakker: `source.substring(start, end)` should be extracted to a separate var, otherwise we're executing it on each iteration. You can then perform all the operations below on this substring, since it's a smaller subset it will be faster.

LarsDenBakker: The find operation should check the first character which matches any of the literals, instead of checking the first literal that matches any of the characters in the string.

packages/es-dev-server/src/utils/resolve-module-imports.js
```diff
@@ -184,18 +184,17 @@ async function resolveModuleImportsWithConfig(importer, source, cfg) {
...
+      const dynamicStart = source.indexOf(stringSymbol, start) + 1;
+      const dynamicEnd = source.indexOf(stringSymbol, dynamicStart);
```
LarsDenBakker:
```suggestion
      const dynamicEnd = source.lastIndexOf(stringSymbol, dynamicStart);
```

packages/es-dev-server/src/utils/resolve-module-imports.js
LarsDenBakker: The result of this can be `dynamicStart`, then you can derive the `stringSymbol` from that. That saves a bit of work :)

```js
const dynamicStart = .. whatever magic ...
const stringSymbol = source[dynamicStart];
```

We will however need handle the case when `dynamicStart === -1` because then there is no string literal in the import. In that case, we should skip a lot of the work below.

packages/es-dev-server/src/utils/resolve-module-imports.js
```diff
@@ -184,18 +184,17 @@ async function resolveModuleImportsWithConfig(importer, source, cfg) {
...
       const [importee, importeeSuffix] = getImportee(source.substring(dynamicStart, dynamicEnd));
-      const stringSymbol = source[dynamicStart - 1];
-      const isStringLiteral = [`\``, "'", '"'].includes(stringSymbol);
+      const isStringLiteral = literals.includes(stringSymbol);
```
LarsDenBakker: `stringSymbol` is taken from the list of literals, so there is no need to check if it's in the list again. probably the clearest would be to check if `dynamicStart !== -1`

LarsDenBakker:
@Anubis012 I forgot about this on friday :( Overall it looks good, but we need to be careful to handle the case when the dynamic import isn't a string literal and do a few optimizations.

Issue #1331, opened 2020-02-10
Add sideEffects: false flag to lit-helpers package


bschlenk:
I just noticed in my webpack bundle that we're pulling in the `live` directive from lit-helpers, even though we only use `spreadProps`. Looks like [this package](https://github.com/open-wc/open-wc/blob/master/packages/lit-helpers/package.json) doesn't set [sideEffects: false](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) which would allow webpack to treeshake these other modules.

I'm open to sending a PR, just want to make sure everyone would be onboard with this.

daKmoR:
that would be awesome 🤗 

actually most of our packages are side effect free 👍 (only testing and testing-helpers have side effects by default as far as I can think) 

care to "check" and add `sideEffects: false` everywhere if appropriate?

bschlenk:
Sure I can check! Although if I’m not mistaken are most of the other packages build utilities? It really only matters for the ones that would potentially be bundled into a production app.

daKmoR:
true makes sense - I think only `scoped-elements` might also end up in a production app 🤗

PR #1337, opened 2020-02-12
fix: add setting so webpack can apply tree shaking on it

bschlenk:
Closes https://github.com/open-wc/open-wc/issues/1331


daKmoR:
thxx 🤗

PR #1339, opened 2020-02-12
Update modern-config.js

### LINKED ISSUE #1338: webpack, outputpath could not be adjusted.
I have a webpack config like this:

    const path = require('path');
    const { createDefaultConfig } = require('@open-wc/building-webpack');

        input: path.resolve(__dirname, './index.html'),
        webpackIndexHTMLPlugin: {
            polyfills: {
                fetch: false,
                intersectionObserver: false,
                polyfillsHash: false,
                coreJs: false,
                regeneratorRuntime: false,
                webcomponents: false,
            },
        },
        plugins : {
            workbox: false
        },
        output: {
            filename: '[name].bundle.js',
            path: path.resolve(__dirname, 'webpack-dist'),
        },
    });

but the path und er output will not be used, cause in here:

https://github.com/open-wc/open-wc/blob/1dcbe3c35906a75ff92eafa9777da9813fa223c0/packages/building-webpack/modern-config.js#L49

the option from outside is not used!

jogibear9988:
fix #1338

### DIFF
--- a/packages/building-webpack/modern-and-legacy-config.js
+++ b/packages/building-webpack/modern-and-legacy-config.js
@@ -43,9 +43,9 @@ function createConfig(options, legacy) {
     entry: options.input || options.entry,
 
     output: {
-      filename: outputFilename,
-      chunkFilename: outputChunkFilename,
-      path: path.resolve(process.cwd(), `dist`),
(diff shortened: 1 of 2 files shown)


jogibear9988:
maybe someone could help with the prettier... I don't know whats the error, but also could not run it localy

PR #1341, opened 2020-02-13
chore(scoped-elements): write valid html in example

dancarlocanlas:
Removed `}` from item number 4 in `Usage` section of `README.md`


daKmoR:
nice catch 👍 

thxxx 🤗

Issue #1272, opened 2020-01-22
[scoped-elements] demos/docs code does not support extending elements


daKmoR:
we have the following code in 

`page-a./indexjs`
```js
const html = createScopedHtml({
  'feature-a': FeatureA,
  'feature-b': FeatureB,
});

export class PageA extends HTMLElement {
  // ...
  connectedCallback() {
    render(html`
      <feature-a></feature-a>
      <feature-b></feature-b>
    `, this.shadowRoot);
  }
}
```

this is basically a side effect and registeres `feature-a`...

now it is not possible to extend without registering `feature-a` 🙈 

`ExtendedPage.js`
```js
import { PageA } from 'page-a'; // triggers side effect already
import { FeatureExtended } from '...';

const html = createScopedHtml({
  'feature-extended': FeatureExtended,
  'feature-b': FeatureB, // do I need this?
});

export class ExtendedPage extends PageA {
  // ...
  connectedCallback() {
    render(html`
      <feature-extended></feature-extended>
    `, this.shadowRoot);
  }
}
```

# Strawman proposal Mixin

Ideally, you could use a static get

```js
export class ExtendedPage extends ScopedElementsMixin(PageA) {
  static get scopedElements() {
    const parentEls = super.scopedElements;
    delete parentEls['feature-b'];
    return {
       ...parentEls, 
       'feature-extended': FeatureExtended,
    }
  }
}
```

### Strawman proposal convention

alternatively, you could have a convention for how to get the scoped html

```js
let html;
export class ExtendedPage extends PageA {
  constructor() {
    // ...
    this._registerSubElements();
  }

  _registerSubElements() {
    html = createScopedHtml({
      'feature-a': FeatureA,
      'feature-b': FeatureB,
    });
  }
````

LarsDenBakker:
It's a side effect in the sense that importing a module will add an entry to the module graph, it doesn't interfere with anything. We could defer registration until the first render of a template perhaps.

LarsDenBakker:
A static class field is quite interesting too from build-time optimization perspective, but it can't be used by sub templates.

The ideal would be if we have a way to plug into lit-html, so that all lit templates used by a web component are rewritten and can use the static field. Then it really behaves the way scoped custom element registries also work. lit-html and lit-element do something similar here, where it passes the element's context to the lit templates to use as `this` when calling event listeners.

manolakis:
[...]

PR #1342, opened 2020-02-13
fix(scoped-elements): support extending elements (#1272)

manolakis:
This commit defers the definition of web components until they are used, so it allows to extend parent components without defining unused inherited web components.


daKmoR:
that is improving nicely 🤗

Issue #1270, opened 2020-01-21
[scoped-elements] add option to retrieve the tags used in WC definition


manolakis:
There are some use cases in which we have to obtain the tags generated by `scoped-elements` in the web-components definition and the current API doesn't provide any method to get them.

As an example, in Lion library two components create inner ones using `document.createElement`:
- [LionSelectRich](https://github.com/ing-bank/lion/blob/master/packages/select-rich/src/LionSelectRich.js#L116)
- [LionSwitch](https://github.com/ing-bank/lion/blob/master/packages/switch/src/LionSwitch.js#L22)

Two different approaches come to mind:
```js
const [ lionSwitchButton ] = createScopedTags([ LionSwitchButton ]);
...
document.createElement(lionSwitchButton);
```
```js
const { html, tags } = createScopedHtml({  'lion-switch-button': LionSwitchButton });
...
document.createElement(tags['lion-switch-button']);
```

but it would be great to discuss a proper API for it.

daKmoR:
how about
```js
const html = createScopedHtml({ ... });
const tags = createScopedTags({ ... });

const { html, tags } = createScope({ ... });
```
?

@LarsDenBakker would that still be possible to optimize (for production) via a babel plugin?

LarsDenBakker:
It's hard to say right now. We won't be able to compile away using the tags, but we dont really have to either. The main point is making sure it doesn't make it harder to compile away the `html` tags. 

We really need to get some numbers on the impact of the template rewriting, to see if it's worth worrying about it.

manolakis:
I've been playing with some internal shared libraries and also with Lion. Applying the scoped elements is really easy but I found two things:

- It seems necessary to have some method to obtain the tagName for the use of `document.createElement`. Other uses of the tag name can be replaced easily by css classes, ids or attributes as Lars pointed some time ago.
- Complex components use to have a template composition, so not all `html` calls should be done with the Html function provided by us (in terms of performance). In the end, we probably are going to have a mix of `html` and, for example, `scopedHtml` in the same component.

I like `const { html, tags } = createScope({ ... });`. Users can obtain what they need with just one function.

tmsns:
You could also have the following exports:
```javascript
const { html, createElement } = createScope({ ... });

createElement(...)
```
This still hides (more or less) the tags themselves, if that's what is worrying.

daKmoR:
uh interesting idea about `createElement` 🤗 
I think that is worth a shot 💪 

[...]

PR #1343, opened 2020-02-14
feat(scoped-elements): add createElement capability (#1270)

manolakis:
Adds the capability to create scoped elements by it's tag name.

```js
import { createScope } from '@open-wc/scoped-elements';

const { createElement } = createScope({
  'my-button': MyButton,
});

const $button = createElement('my-button');
```


manolakis:
Closing because it's not relevant anymore

PR #1349, opened 2020-02-17
feat(scoped-elements): implement as ScopedElementsMixin

LarsDenBakker:
TODO:
- [x] Re-implement https://github.com/open-wc/open-wc/pull/1342
- [x] Re-implement https://github.com/open-wc/open-wc/pull/1329
- [x] Update demos
- [x] Update tests
- [x] Ensure tagnames are *always* random to avoid users depending on the names
- [ ] Update types


packages/scoped-elements/package.json
```diff
@@ -35,14 +35,14 @@
     "utils",
     "helpers"
   ],
-  "peerDependencies": {
-    "lit-html": "^1.0.0"
+  "dependencies": {
+    "@open-wc/dedupe-mixin": "^1.1.0",
+    "lit-element": "^2.0.1"
```
manolakis: Not sure if `lit-element` should be added as dependency or peerDependency. :thinking:

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -0,0 +1,81 @@
...
+
+  // @ts-ignore
+  const { scopedElements = {} } = element.constructor;
+
+  const transformedTemplate = transformTemplate(template, scopedElements, templateCache);
+
+  return litRender(transformedTemplate, container, options);
+};
+
+export const ScopedElementsMixin = dedupeMixin(superclass => {
```
manolakis: I couldn't add a valid jsDoc. Any help is welcome

packages/scoped-elements/src/transform.js
```diff
@@ -1,8 +1,28 @@
...
+ * @description allowed tag name chars
+ * @type {string}
+ */
+const chars = `-|\\.|[0-9]|[a-z]`;
 
+/**
+ * @description Regular Expression to find a custom element tag
+ * @type {RegExp}
+ */
+const re = new RegExp(`<\\/?([a-z](${chars})*-(${chars})*)`, 'g');
```
manolakis: This regular expression avoids the processing of native tags like `button`, `p`, etc and only custom tags are going to be processed.

packages/scoped-elements/package.json
LarsDenBakker: We depend on lit element, do we not?

packages/scoped-elements/src/transform.js
LarsDenBakker: It will give false positives for comments and attributes though, right?

LarsDenBakker: It will give false positives for comments and attributes though, right?

packages/scoped-elements/package.json
manolakis: right 👍

packages/scoped-elements/src/transform.js
manolakis: I don't think so... in reality is searching elements like `<my-button` or `</my-button` and attributes can't start with `<` or `</`

Issue #1360, opened 2020-02-22
workbox precaching legacy and polyfill files by default


vikerman:
The default workbox config glob seems to be '**/*.{html,js,css}' which is precaching all JS including the polyfill and legacy JS. For the starter app with legacy config it comes to 368KB over 10 URLs which seems pretty big.

Is this behavior intentional? I would think just precaching the modern ES modules JS without the polyfills would be better since the chances of service worker being present in older browser(IE11 and friends) is anyway lower.

LarsDenBakker:
You are right, the legacy files should not be cached.

thepassle:
I think I know why this is happening, workbox should not run when [`legacy` is set to false](https://github.com/open-wc/open-wc/blob/master/packages/building-rollup/modern-and-legacy-config.js#L126), but the legacy config is ran first, and _then_ the modern config runs, which also runs workbox, meaning that the legacy files also get precached.

I think this can be fixed by switching:

```js
return [createConfig(options, true), createConfig(options, false)];
```

to 

```js
return [createConfig(options, false), createConfig(options, true)];
```
I'll investigate further tomorrow, but would also be happy to look at a MR 😊

vikerman:
The above fix could potentially break people extending the legacy config to add plugins based on index(config[0], config[1] etc.).

For now I have created a local workbox config of just *.{html|css|js} which could also work since the legacy and polyfills JS go into subdirectories. Let me know how you would want to fix and I can try to create a PR.

LarsDenBakker:
If we cant find a nom breaking fix, we have a breaking change upcoming for the rollup configs, we can take along this change then.

vikerman:
This would handle only the legacy files and not the polyfills?

LarsDenBakker:
Correct, our polyfill setup doesnt only handle features for really old browser but also modern features for browsers which do support service workers. For example intersection observer, resize observer, dynamic imports. There will also be new web component apis that will need polyfills in the future.

thepassle:
We could probably improve on that and add runtime caching for the polyfills now that I think of it actually. It doesnt make sense to aggressively precache _all_ polyfills, when maybe not all of them will be used, but when they get used, they'll still get cached.

I think we may be able to do this as a non breaking change, too.

LarsDenBakker:
Ah yes, interesting. Should we open a separate issue for that?

PR #1365, opened 2020-02-23
fix: don't precache legacy files and polyfills

thepassle:
fixes https://github.com/open-wc/open-wc/issues/1360

This should be a non breaking change, since the legacy files/polyfills should never be requested in the first place (since those browsers dont support service worker), and on a new deploy workbox will clean outdated caches

### DIFF
--- a/packages/building-utils/get-workbox-config.js
+++ b/packages/building-utils/get-workbox-config.js
@@ -5,6 +5,7 @@ const path = require('path');
 module.exports = function getWorkboxConfig(outputDir) {
   const workboxConfigPath = path.join(process.cwd(), 'workbox-config.js');
   const defaultWorboxConfig = {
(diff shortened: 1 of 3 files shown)


packages/building-utils/get-workbox-config.js
```diff
@@ -5,6 +5,7 @@ const path = require('path');
 module.exports = function getWorkboxConfig(outputDir) {
   const workboxConfigPath = path.join(process.cwd(), 'workbox-config.js');
   const defaultWorboxConfig = {
+    globIgnores: ['/polyfills/*.js', '/legacy/*.js'],
```
LarsDenBakker: I think we should cache polyfills, some of them will be loaded on modern browsers too. What do you think?

LarsDenBakker:
Can you make this two separate commits, one for webpack and one for rollup? That way they end up in the changelog of those projects.

PR #1376, opened 2020-02-26
feat(scoped-elements): add support for lazy elements

manolakis:
Adds support for lazy elements

```js
      class extends ScopedElementsMixin(LitElement) {
        static get scopedElements() {
          return {
            'feature-a': FeatureA,
          };
        }

        constructor() {
          super();

          setTimeout(() => this.constructor.defineScopedElement('feature-b', FeatureB), 100);
        }

        render() {
          return html`
            <feature-a></feature-a>
            <feature-b></feature-b>
          `;
        }
      }
```

Also the function `getScopedTagName` has been moved as a Mixin method.


packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -2,35 +2,44 @@
...
  * @type {WeakMap<Function, Map<TemplateStringsArray, TemplateStringsArray>>}
  */
 const templateCaches = new WeakMap();
 
+/**
+ * Lambda caches
+ *
+ * @type {WeakMap<object, Map<string, function(): Promise<typeof HTMLElement>>>}
+ */
+const lambdaCaches = new WeakMap();
```
manolakis: I had to create this lambda cache because each time we call scopedElements new lambdas were created and I nedd to maintain a reference to link it with the scoped tag name.

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -2,35 +2,44 @@
...
+
 /**
  * Transforms an array of TemplateResults or arrays into another one with resolved scoped elements
  *
  * @param {ReadonlyArray} items
  * @param {Object.<string, typeof HTMLElement>} scopedElements
  * @param {Map<TemplateStringsArray, TemplateStringsArray>} cache
+ * @param {Map<string, function(): Promise<typeof HTMLElement>>} lambdaCache
```
LarsDenBakker: Type can be (() => Promise<...>).

packages/scoped-elements/src/ScopedElementsMixin.js
LarsDenBakker: I suggest calling it functionCache :)

packages/scoped-elements/src/registerElement.js
```diff
@@ -1,55 +1,104 @@
 import { createUniqueTag } from './createUniqueTag.js';
 
 /**
- * Global cache for tag names
+ * @typedef {function(): Promise<typeof HTMLElement>} Lambda
```
LarsDenBakker: I suggest calling this LazyElement or LazyElementFn

packages/scoped-elements/src/ScopedElementsMixin.js
manolakis: nice

packages/scoped-elements/src/ScopedElementsMixin.js
manolakis: I know lambdaCache was not a good name for JS world 😆 

functionCache it's ok for me 👍

packages/scoped-elements/src/registerElement.js
manolakis: +1 to LazyElementFn

manolakis:
>Sorry for questioning our original design, but seeing all the extra caching work we need to do makes me rethink this a bit.

No worries @LarsDenBakker, the intention of this PR was having some working code to discuss the design :)

>What if instead when we encounter custom element tags which dont have a scoped class set we scope it regularly and remember the tag to scoped tag mapping.
>
>Then, at any point the user can call defineScopedElement and pass the class. The user can do the import themselves.

So basically we would have something like this?

```js
class extends ScopedElementsMixin(LitElement) {
  static get scopedElements() {
    return {
      'feature-a': FeatureA,
      'feature-b',
    };
  }

  constructor() {
    super();

    setTimeout(
      () => import('./FeatureB.js').then(module =>
        this.defineLazyScopedElement('feature-b', module.FeatureB),
      100,
    );
  }

  render() {
    return html`
      <feature-a></feature-a>
      <feature-b></feature-b>
    `;
  }
}
```

LarsDenBakker:
Yes I was thinking something like that. We will need to think about some way to indicate that a custom element will be lazy. Or should we just assume that for any unknown custom element definition that it will be lazy? That's probably what the browser will do, not render any unknown elements.

manolakis:
I guess browser will render the tag in the shadon dom and at some point the user will define the element in the registry so the element would be upgraded. As we must scope the tag we should know it in advance to render it scoped in dom.

manolakis:
As a first approach we could recognize as a lazy element every tag without an HTMLElement

packages/scoped-elements/src/ScopedElementsMixin.js
manolakis: Intellij marks it as invalid type syntax 🤔 
Screenshot 2020-02-27 at 17 52 51

LarsDenBakker: Try wrapping it in `( )` ?

packages/scoped-elements/test/ScopedElementsMixin.test.js
```diff
@@ -256,8 +256,40 @@ describe('ScopedElementsMixin', () => {
...
+  });
 
-    expect(getFromTagsCache(FeatureD)).to.be.undefined;
+  it('supports lazy loaded elements', async () => {
+    const tag = defineCE(
+      class extends ScopedElementsMixin(LitElement) {
+        static get scopedElements() {
+          return {
+            'feature-a': FeatureA,
+            'feature-b': none,
```
LarsDenBakker: what if we just default to this behavior for an unscoped tag? so there is no need to explicitly declare it, any tag we can't find we scope it and wait for when it's defined later

packages/scoped-elements/src/ScopedElementsMixin.js
manolakis: yes, also wrapping it says Invalid type syntax. Anywaay, with the new approach is not needed.

packages/scoped-elements/test/ScopedElementsMixin.test.js
manolakis: The problem with that is that it would break compatibility with self defined components.

I thought to not scope them until the element is defined, changing the tag in the dom at that moment, but it would cause problems with `getScopedTagName` function if it's called before the element is defined 🤷‍♂️

LarsDenBakker: If you don't change the tag, it can accidentally take on a definition from a global element with that name. 

When using scoped elements, I think you should not inherit anything globally so breaking compatibility with self defined components is fine.  If a component doesn't ship it's class you will need to do `'feature-b': customElements.get('feature-b'),`.

Changing the tag name at a later point in time would probably be pretty trick to do?

manolakis: Yes, good point. So then I think we could scope all custom elements in template by default so then we could remove the lazy ones from `static get scopedElements` and even `getScopedTagName` would continue working as expected.

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -60,17 +71,54 @@ export const ScopedElementsMixin = dedupeMixin(
...
         return super.render(transformedTemplate, container, options);
       }
+
+      /**
+       * Defines a scoped element
+       *
+       * @param {string} tagName
+       * @param {typeof HTMLElement} klass
+       */
+      defineScopedElement(tagName, klass) {
```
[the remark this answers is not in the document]
manolakis: That's true. I'll change it to static.

LarsDenBakker:
Looks good to me, but we need to update it with master.

manolakis:
I'm going to do the rebase

PR #1389, opened 2020-03-01
docs: make navbar more easily navigable

LarsDenBakker:
- Make the nested menus non-collapsible. This makes it easier to see what's available and to jump between sections.
- Don't show page headings in the nav bar. It's a cool feature, but on some pages you really lose an overview of where you are. Our pages are usually pretty concise.
- Rename Tools to Tools & Libraries
- Move DedupeMixin to Tools & Libraries

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -16,7 +16,6 @@ const sidebar = [
       '/developing/routing',
       ['/developing/ide', 'IDE'],
       '/guide/component-libraries',
-      '/guide/dedupe-mixin',
     ],
   },
   {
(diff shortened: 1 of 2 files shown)


thepassle:
should we move the dedupemixin in this PR as well?

LarsDenBakker:
@thepassle done!

thepassle:
In the linting pages we still mention:

```
# Upgrade > Linting
```

we should probably remove those too while we're here (im not home rn but i can do it when i get home if you want)

LarsDenBakker:
@thepassle I see upgrade is still an option in `npm init @open-wc`

thepassle:
yeah, i was wrong, disregard what i said :p

this lgtm 👍

PR #1392, opened 2020-03-01
Feat/create simplify app

LarsDenBakker:
This simplifies the app scaffolding. It's a combination of things we've discussed here and there, and some new things. It's a big change, I'd like to get an approval from the whole core team one this one :)

The goal for me is to reduce the amount of files and package.json scripts which are generated, and to make it a general use case for all types of projects.

Most of the generated files comes from reusing the component generator for each of the app components. This creates a readme, index.js etc. which don't really fit the components in an app project. I changed this to just static files in the app generator.

Overview of changes:

- index.js is removed, we just keep `MyElement.js` and `my-element.js`
- README.md is removed
- `components/my-component/src` is removed, apps don't have public/private files. 
- `stories/my-element.stories.mdx` is changed to `stories/my-element.stories.js` since that fits app components better
- `demo/index.html` is removed, there is already an app-level demo

The final generated file structure if you turn on all options:
```
├── owc-app/
│   ├── .storybook/
│   │   ├── main.js
│   │   └── preview.js
│   ├── components/
│   │   ├── owc-app/
│   │   │   ├── stories/
│   │   │   │   └── owc-app.stories.js
│   │   │   ├── test/
│   │   │   │   └── owc-app.test.js
│   │   │   ├── aboutTemplate.js
│   │   │   ├── open-wc-logo.js
│   │   │   ├── owc-app.js
│   │   │   └── OwcApp.js
│   │   ├── page-main/
│   │   │   ├── stories/
│   │   │   │   └── page-main.stories.js
│   │   │   ├── test/
│   │   │   │   └── page-main.test.js
│   │   │   ├── page-main.js
│   │   │   └── PageMain.js
│   │   ├── page-one/
│   │   │   ├── stories/
│   │   │   │   └── page-main.stories.js
│   │   │   ├── test/
│   │   │   │   └── page-one.test.js
│   │   │   ├── page-one.js
│   │   │   └── PageOne.js
│   ├── .editorconfig
│   ├── .gitignore
│   ├── custom-elements.json
│   ├── index.html
│   ├── karma.conf.js
│   ├── LICENSE
│   ├── package.json
│   ├── README.md
│   └── rollup.config.js
```

I also made some other changes in this PR:

- Removed snapshot testing scaffolding. Snapshot testing is useful, but not for all types of projects. Having it in by default doesn't feel right, we have it in our docs so people can choose to add it when it makes sense for their project. Removing it reduces the amount of package.json scripts by 2.
- Removed browserstack testing scaffolding. I don't feel 100% certain about this one. Browserstack testing is very useful, but not every project can or will make use of it. It feels a bit overkill to generate the files for it when a lot of people need to remove it later on. Removing this reduces 1 generated file in the root of the project, and one package.json script. 

@daKmoR @thepassle @bennypowers @Westbrook


LarsDenBakker:
You can try out the changed in the generator by checking out the branch, doing `cd packages/create` and `yarn build && node dist/create.js`

thepassle:
Very nice 😊

Could we simplify even more? E.g. create-react-app just scaffolds a App.js, App.css, and App.test.js. I think it would be really nice to cut down the total amount of files to:

```
├── owc-app/
│   ├── .storybook/
│   │   ├── main.js
│   │   └── preview.js
│   ├── components/
│   │   ├── owc-app/
│   │   │   ├── stories/
│   │   │   │   └── owc-app.stories.js
│   │   │   ├── test/
│   │   │   │   └── owc-app.test.js
│   │   │   ├── open-wc-logo.js
│   │   │   ├── owc-app.js
│   │   │   └── OwcApp.js
│   ├── .editorconfig
│   ├── .gitignore
│   ├── custom-elements.json
│   ├── index.html
│   ├── karma.conf.js
│   ├── LICENSE
│   ├── package.json
│   ├── README.md
│   └── rollup.config.js
```

I think it makes things a lot more clear. We could also then remove the 'pseudo router' which is now causing a11y tests to fail for skip-links. As another added bonus, it looks even less like a "fake monorepo", which is a real issue that users still get confused by

LarsDenBakker:
@thepassle so we wouldn't have pages anymore in that case?

thepassle:
@LarsDenBakker correct

LarsDenBakker:
It feels like we would miss out on some guidance to users. Perhaps we should put effort again into `owc-starter-app` with more opinionated project structure, library usage etc.?

thepassle:
My personal experience is that whenever I use our scaffolding, the first thing I do is remove the pseudo router, and remove the PageOne and PageMain. I dont think the pseudo router is really that great of a pattern that we should be recommending as a default. I'm also not sure what guidance they'd miss out on exactly by removing those pages

It would be interesting to hear some feedback from users to see what they think and how they use it.

I also dont think we should separate things again by having a separate `owc-starter-app` _and_ the generator, keeping the generator up to date proves to be hard enough already.

petecarapetyan:
There are so many things that have come to mind, in the history of our little team's usage. Here are some out of order observations:

- super nice to be able to generate a more complete organization of an app, even just to see how to organize our own. In that view, I select everything.
- ditto on each comment/change above
- There is no ts flavor, that would be nice to be able to select that and generate ts style instead
- We seem to have two realities for our wc. Simple for a project unshared unstoried, and more canonical storybook like that has the fully featured. Some wc are just dumb containers for other wc or such
- Our simple unshared wc has no src folder, no separate index, just a folder named for the wc and an index.ts which holds everything. No testing or storybook or anything else, just that one file.
- the act of removing generated files seems innocuous to me but others on the team get weird as if "owc sucks" so there is that illogical reality - so now you know.
- a folder named components would be better for us as wc - shorter and more accurate
- what about non-wc libraries such as redux, router etc. We really don't like placing them as siblings to web components, instead they go in a [project-name]/src folder
- the reality for us is that TS, a different router, redux/thunk, pwa tweaks including service worker etc and similar issues make usage of the generator no longer a part of our standard setup, which is sad.
- issue 575 closed is a total frkn pita for me because my stupid 2013 always updated mac still won't let me get around this. Yes, I know it works for everyone else in the world, I wistfully look over my shoulder and see others using the generator without this problem. Call the whaaaambulance.

LarsDenBakker:
After thinking about it some more, I agree with @thepassle . The generator should be a way to set up the project and you shouldn't need to delete a lot of stuff before getting started. 

I think it's useful to provide guidance on structuring your application, but there are so many ways to do that there is no one size fits all.

bennypowers:
I'm for @thepassle 's minimal case

packages/create/src/generators/app-lit-element/templates/static/README.md
```diff
@@ -21,4 +21,10 @@ npm init @open-wc
 - `start:build` runs your app after it has been built using the build command
 - `build` builds your app and outputs it in your `dist` directory
 - `test` runs your test suite with Karma
-- `lint` runs the linter for your project
\ No newline at end of file
+- `lint` runs the linter for your project
+
+## Tooling configs
+
+For most of the tools, the configuration is in the `package.json` to reduces the amount of files in your project.
```
thepassle: To reduce

(I guess you cant post suggestions on mobile? 😕 sorry)

packages/create/src/generators/wc-lit-element/templates/static/README.md
```diff
@@ -18,6 +18,12 @@ npm i <%= tagName %>
 
 <%= featureReadmes %>
 
+## Tooling configs
+
+For most of the tools, the configuration is in the `package.json` to reduces the amount of files in your project.
```
thepassle: To reduce

Issue #1368, opened 2020-02-24
[es-dev-server] check accept header before injecting into HTML


LarsDenBakker:
es-dev-server doesn't work for projects using html imports, because html imports request html files with an accept */* header. This causes a false positive in the dev server, which thinks it's serving a regular HTML file.

Instead of only checking the server response content-type header, we should also double check the browser is requesting a text/html file.

dancarlocanlas:
> Instead of only checking the server response content-type header, we should also double check the browser is requesting a text/html file.

Does this refer to the checking being done in `isIndexHTMLResponse` under `utils`?

LarsDenBakker:
Yes indeed

LarsDenBakker:
Closing since this was fixed.

davidmaxwaterman:
I'm just come back to the project that was the reason for this issue to be raised, and I notice the problem is still there. I see that the fix has been merged, so I'm not sure if the fix didn't fix it for me (or the polymer init test case), or if the fix has made it into the most recent release. Any ideas?

Here's my comment on slack:

Hi, I just moved back to this project, and it seems to be having the same problem again...ie requests for message_channel as hanging. I updated the globally installed es-dev-server for my other project (if you recall), and now it doesn't work for this project again. I updated to the very latest, but it still fails. Furthermore, I downgraded to 1.38.2 and it no longer works:
```
  TypeError: (0 , _polyfillsLoader.findJsScripts) is not a function
      at findScripts (/home/davidmaxwaterman/.nvm/versions/node/v12.14.1/lib/node_modules/es-dev-server/dist/utils/inject-polyfills-loader.js:100:58)  TypeError: (0 , _polyfillsLoader.findJsScripts) is not a function
      at findScripts (/home/davidmaxwaterman/.nvm/versions/node/v12.14.1/lib/node_modules/es-dev-server/dist/utils/inject-polyfills-loader.js:100:58)
```
but I think that's a different problem since it fails on the very first request (for index.html).
With the latest es-dev-server:
```
$ npm ls -g es-dev-server
/home/davidmaxwaterman/.nvm/versions/node/v12.14.1/lib
└── es-dev-server@1.54.1 
```
it serves the first 6 requests for message-channel (dev tools says the first is initiated in index.html, and the other 5 from 'other'), but the subsequent (25 - 6) requests remain (all from 'other') as 'pending'.
Es-dev-servers 'debug' output mentions 1 request early on:
```
[...]

PR #1400, opened 2020-03-04
fix(es-dev-server): updated checking for index html request

dancarlocanlas:
[1368](https://github.com/open-wc/open-wc/issues/1368)


packages/es-dev-server/src/utils/utils.js
```diff
@@ -126,13 +126,19 @@ export async function isIndexHTMLResponse(ctx, appIndex) {
   }
 
   // if we're serving the app index, it's an index html response
-  if (ctx.url === appIndex) {
+  if (ctx.url === appIndex || ctx.url.includes('index.html')) {
```
LarsDenBakker: these kinds of assumptions keep biting us,  so I don't want to assume that `index.html` means a HTML file.

dancarlocanlas: If that's the case, do we need to add checking for the scenario where no `appIndex` is provided? The tests for `polyfills-loader` failed after adding the checking for accept type, as it is now giving out a false negative, unless we explicitly add `appIndex` in the tests' config.

LarsDenBakker: We will need to add the header in fetch requests in the tests probably.

The app index check is just a nice to have, but not strictly required.

dancarlocanlas: Updated the tests with the latest commit :)

LarsDenBakker:
Thanks for the contribution!

PR #1406, opened 2020-03-08
WIP: feat(es-dev-server): allow passing chokidar options as `watch` option

### LINKED ISSUE #1398: [es-dev-server]: expose chokidar options
https://polymer.slack.com/archives/CE6D9DN05/p1583315939403900

I propose changing the type of the `watch` option from `boolean` to `boolean|WatchOptions`

https://github.com/paulmillr/chokidar/blob/6217706d11cac734c437fa4757e4b002a1d90b43/types/index.d.ts#L67-L167

bennypowers:
closes #1398

### DIFF
--- a/packages/es-dev-server/src/command-line-args.js
+++ b/packages/es-dev-server/src/command-line-args.js
@@ -64,7 +64,16 @@ export const commandLineOptions = [
   {
     name: 'watch',
     alias: 'w',
-    type: Boolean,
+    type: x => {
+      switch (x) {
+        case 'true':
+          return true;
+        case 'false':
+          return false;
+        default:
+          return JSON.parse(x);
+      }
+    },
     description: 'Reload the browser when files are edited.',
   },
   {
--- a/packages/es-dev-server/src/utils/utils.js
+++ b/packages/es-dev-server/src/utils/utils.js
@@ -115,19 +115,13 @@ export class SSEStream extends Stream.Transform {
 }
 
 /**
+ * Typeguard which asserts that the arg is of type `boolean`.
+ * Used to disambiguate union types e.g. in function params
  * @param  {*}  x
  * @return {x is boolean}
  */
 export function isBoolean(x) {
   return typeof x === 'boolean';
-}
-
-/**
- * @param  {boolean|import('chokidar').WatchOptions}  watch
- * @return {watch is WatchOptions}
- */
-export function isWatchOptions(watch) {
-  return typeof watch === 'object';
 }
 
 /**
--- a/packages/es-dev-server/src/utils/utils.js
+++ b/packages/es-dev-server/src/utils/utils.js
@@ -115,6 +115,22 @@ export class SSEStream extends Stream.Transform {
 }
 
 /**
+ * @param  {*}  x
+ * @return {x is boolean}
+ */
+export function isBoolean(x) {
+  return typeof x === 'boolean';
+}
+
+/**
+ * @param  {boolean|import('chokidar').WatchOptions}  watch
+ * @return {watch is WatchOptions}
+ */
+export function isWatchOptions(watch) {
+  return typeof watch === 'object';
+}
+
+/**
  * Returns the index html response, or null if this wasn't an index.html response
  * @param {import('koa').Context} ctx
  * @param {string} appIndex
--- a/packages/es-dev-server/src/config.js
+++ b/packages/es-dev-server/src/config.js
@@ -34,7 +34,7 @@ import { compatibilityModes } from './constants.js';
  * @property {boolean} debug whether to log debug messages
  *
  * Development help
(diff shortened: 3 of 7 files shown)


bennypowers:
- missing tests (master skips tests)
- disconcertingly, using this branch solved my original problem (`Not Found` in browser on watch-reload) *without* passing @thepassle 's recommended options to config.

packages/es-dev-server/src/command-line-args.js
```diff
@@ -64,7 +64,16 @@ export const commandLineOptions = [
   {
     name: 'watch',
     alias: 'w',
-    type: Boolean,
+    type: x => {
```
LarsDenBakker: I didn't know about this feature, very useful for some other cases too!

packages/es-dev-server/src/utils/utils.js
```diff
@@ -114,6 +114,22 @@ export class SSEStream extends Stream.Transform {
   }
 }
 
+/**
+ * @param  {*}  x
+ * @return {x is boolean}
+ */
+export function isBoolean(x) {
```
LarsDenBakker: Not sure if it's worth creating a helper function for this? `typeof x === 'boolean'` is fairly straightforward?

packages/es-dev-server/src/utils/utils.js
```diff
@@ -114,6 +114,22 @@ export class SSEStream extends Stream.Transform {
...
+  return typeof x === 'boolean';
+}
+
+/**
+ * @param  {boolean|import('chokidar').WatchOptions}  watch
+ * @return {watch is WatchOptions}
+ */
+export function isWatchOptions(watch) {
```
LarsDenBakker: This isn't used anywhere?

packages/es-dev-server/src/utils/utils.js
bennypowers: except typescript makes the straightforward crooked.

the return type `x is boolean` is a necessary typeguard to let options be passed as `WatchOptions` and not `boolean|WatchOptions`

packages/es-dev-server/src/utils/utils.js
bennypowers: true. see previous comment. seems typescript preferred the opposite predicate 🤷‍♂

bennypowers:
>  create the filtewatcher with the config from the cli before we call `createServer`.

The goal is to have 
```js
// es-dev-server.config.js
module.exports = {
  // ...
  watch: { ...chokidarWatchOptions }
}
```

If we can do that the way you suggest, then sure

packages/es-dev-server/src/utils/utils.js
LarsDenBakker: ok, makes sense.

LarsDenBakker:
Initially I thought the config is cli-only, but that's not true we want this for people who use the config programmatically as well.

In that case, can we call the function parameter `watcherOrWatcherConfig` or something of that nature? And deconstruct `const fileWatcher = watcher || chokidar.watch([], ...(!isBoolean(watch) ? [watch] : []));` a little bit so that it's easier to grok at first glance :P

bennypowers:
> Initially I thought the config is cli-only, but that's not true we want this for people who use the dev server programmatically as well.
> 
> In that case, can we call the function parameter `watcherOrWatcherConfig` 

is this what you had in mind?
```js
  let fileWatcher = watcherOrWatchOptions;
  if (!fileWatcher) {
    const chokidarOptionsArgs = !isBoolean(cfg.watch) ? [cfg.watch] : []
    fileWatcher = chokidar.watch([], ...chokidarOptionsArgs);
  }
```
in which case, `maybeWatcher` is a better name, since the config was passed in the first param.

What do we do when the config in the first param is different to the existing watcher's config? AFAIK there's no chokidar API to add or update config, so should we close the watcher and create a new one each time?

packages/es-dev-server/src/utils/utils.js
bennypowers: removed

LarsDenBakker:
Hmmm, indeed that's problematic. Introducing a breaking change for this would be unfortunate, and recreating a file watcher can create problems with people who expect their instance to be used.

Should we ignore `cfg.watch` then if it's passed as a parameter? We can just revert back to the original function parameter structure then.

bennypowers:
@LarsDenBakker I'd appreciate some guidance on where to take this next.

LarsDenBakker:
@bennypowers 

I think we need to go back to the original behavior of the watcher parameter where it is only a watcher, not either a watcher or options. And then we only read the options from `cfg.watch`.

daKmoR:
are we still pursuing this? 🤔

bennypowers:
would be nice but I don't have time to go after it atm

Issue #1409, opened 2020-03-09
[testing-helpers] fixtureCleanup is only run for bdd


thepassle:
It would be very nice to support `teardown` as well

https://github.com/open-wc/open-wc/blob/master/packages/testing-helpers/src/fixture.js#L13

PR #1410, opened 2020-03-10
feat(testing-helpers): support fixture cleanup in teardown

thepassle:
fixes https://github.com/open-wc/open-wc/issues/1409

Not sure if this is a breaking change since the check for `afterEach` changed?

Alternatively I could do it like this:

```js
/**
 * This registers the fixture cleanup as a side effect
 */
try {
  // we should not assume that our users load mocha types globally
  // @ts-ignore
  if (afterEach) {
    // @ts-ignore
    afterEach(() => {
      fixtureCleanup();
    });
  }
} catch (error) {
  /* do nothing */
}

try {
  // @ts-ignore
  if (teardown) {
    // @ts-ignore
    teardown(() => {
      fixtureCleanup();
    });
  }
} catch (error) {
  /* do nothing */
}
```

but it seemed a little bit ugly. Thoughts?


packages/testing-helpers/src/fixture.js
```diff
@@ -8,12 +8,19 @@ export { fixture, fixtureSync } from './fixture-no-side-effect.js';
 try {
   // we should not assume that our users load mocha types globally
   // @ts-ignore
-  if (afterEach) {
+  if (typeof afterEach !== 'undefined') {
```
LarsDenBakker: Can we change these to 
```
if ('afterEach' in window') {

}
```

thepassle: Done

LarsDenBakker:
LGTM, should it still be WIP?

PR #1412, opened 2020-03-10
feat(es-dev-server): auto compat for browsers without stage4 features

LarsDenBakker:
`es-dev-server` skips compatibility mode for the latest versions of modern browsers. This assumes that people don't use syntax which isn't supported by browsers, and that browsers typically implement features before they become a standard.

This assumptions holds for most features, and for most browsers. For example class fields is being implemented in most browsers now and it's not yet a stage 4 feature. 

Unfortunately this wasn't the case with optional chaining and nullish coalescing, they became a standard before they were available unflagged in any browser. Additionally the safari release schedule is painfully slow, so while they might implement something early on it's not available for developers.

I'd like to improve the detection with an additional check for certain browser versions which have not implemented all stage 4 features yet. Today Firefox 74 is released, so in effect this means we continue to not do any work on the latest Chrome, Edge, Firefox, Brave etc. 

We would need to maintain this list by hand, but only to add support/checks for new features. There is no danger in the list getting out of date and giving incorrect results.


bennypowers:
TBH I'd rather see ES2017 and/or Stage 4 as the standard with anything beyond (even if widely implemented) done by users in `.babelrc`

LarsDenBakker:
Thats what im going for, I want people to be able to use stage 4 features with es dev server taking care of compatibility. Right now you cant, because safari hasnt implemented it.

Anything beyond is for the user to configure, because they are not standards.

thepassle:
Sounds good to me 👍

daKmoR:
it is unfortunate that we need to do this 🙈 but probably worth it 🤔

the irony that chrome 79 is not modern is kinda funny 😬

but yeah fine with me 👍

PR #1415, opened 2020-03-10
ci: save all package files to workspace

Westbrook:
Types weren't being cached across the workspace and thus were not being published. This changes that, hopefully.

Use `- repo/packages` as opposed to `- repo/packages/*/node_modules` and `- repo/packages/*/dist` so that ALL the files in packages are cached to the workspace. There is likely a similarly successful, but slightly more optimized, list of files to cache, but want to make sure we get this fixed without a lot of side testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,12 +39,12 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - attach_workspace:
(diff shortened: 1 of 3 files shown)


Westbrook:
Can get this to build for some reason... similar changes are working in #1416

PR #1422, opened 2020-03-11
docs(core): add link to slick-router routing library

blikblum:
slick-router is a well tested, feature rich router with first class web component support

BTW: there's not an appropriate commit message scope for such PRs

### DIFF
--- a/docs/developing/routing.md
+++ b/docs/developing/routing.md
@@ -33,6 +33,11 @@ In alphabetical order:
 
   <hr>
 
+- [slick-router](https://github.com/blikblum/slick-router)
+  <NpmUsage routerName="slick-router"/>
+
+  <hr>
+
 - [vaadin-router](https://github.com/vaadin/vaadin-router)
   <NpmUsage routerName="@vaadin/router"/>
   <hr>


LarsDenBakker:
We already have quite a few routers on our recommendations website, so we try to be conservative when adding now ones. It's not meant as a package listing, but rather as a list of recommendations. 

We also show a warning when the weekly downloads are below 400, slick-router would fall into that category directly. I'm not sure if it's appropriate to add in that case.

blikblum:
> It's not meant as a package listing, but rather as a list of recommendations.

That's the reason why i opened this PR: i think is good candidate to be recommended 

> We also show a warning when the weekly downloads are below 400, slick-router would fall into that category directly. I'm not sure if it's appropriate to add in that case.

slick-router is pretty new (although descending of an old project - [cherrytree](https://github.com/QubitProducts/cherrytree)) with few advertisement, so is penalized with this regard while the others were exposed to web component devs by open-wc for sometime.  

Anyway, those metrics tell more about popularity and, IMO, should not be the main reason to do a recommendation. The technical aspects must be considered also.

Some of technical aspects:
- Well tested, covering unit and functional tests, both to core routing and component rendering
- Distributed as ES modules, fully tree shackable 
- web component library aware (use with vanilla, lit-element, skate etc)
- Flexible, covering many use cases (component aware of routing or decoupled of it  / opinionated rendering or custom render )
- Feature rich: async lifecycle hooks, nested route tree, streamlined support for code splitting, powerful router animation (written as an standalone, decoupled web component, embracing the spirit of web components)
- Many demos, from simple demos to full application covering aspects like authentication, redirection, animation, automatic link handling
- The documentation exists but is really lacking some love, organization

So, lets look at the current recommended routers:

https://github.com/hamedasemi/lit-element-router: limited tests (test just utilities), no support for hash change routing, no link handling (a problem with pushstate), very simple demos (the advanced demos are behind a todo flag). There are other questionable implementation like setting window.onpopstate and adding a window event and not unregistering it

https://github.com/danevans/lit-router [limited testing](https://github.com/danevans/lit-router/blob/master/test.js), no docs, no demos. I can't even evaluate properly

https://github.com/Polymer/pwa-helpers#routerjs not a proper library (either you use the entire project or copy the code), lack of features (seems by design) limited docs, limited tests (fine for it scope)

https://github.com/mksarge/redux-first-routing well tested, tied to redux (both a con and a pro), more a pattern than a library (needs a third party router to work)

https://github.com/Matsuuu/simplr-router: well tested, limited documentation, opinionated component structure (routing tied to components), supports animation but with little customization

So, considering the above options slick-router clearly offers advantages in many aspects.

blikblum:
The last option is clearly the best recommendation: vaadin-router - well tested, good docs, feature rich.

Even so, slick-router provides some advantages warranting be considered as an option by devs:

- smaller API surface providing more or less the same functionality
- first class nested component / route support (vaadin have some [problems](https://github.com/vaadin/vaadin-router/issues/419))
- allow, through opt-in option, to reuse component instance (vaadin [lacks this feature](https://github.com/vaadin/vaadin-router/issues/250))
- supports router link automatic handling (setting href, active class) (vaadin [still planning](https://github.com/vaadin/vaadin-router/issues/250))
- provides primitive to decouple component from route specific properties through [properties](https://github.com/blikblum/slick-router/blob/master/docs/middlewares/wc.md#properties) and `paramValue` helper. This makes testing easier
- more powerful, still simple, animation support. slick-router uses a system adapted from vue one that provides more granular control. Its possible to use also third party code like animate.css or Velocity
See demos:
https://codesandbox.io/s/slick-router-css-animations-q0fzs
https://codesandbox.io/s/slick-router-animate-css-zpg96
https://codesandbox.io/s/slick-router-gsap-animations-oqbp5
https://codesandbox.io/s/animated-outlet-page-transitions-7vgcy (standalone usage of animation component)

thepassle:
You're right that popularity metrics are not the best metrics to measure a packages quality. However, it does help us maintain the list of routers we currently have, if a package suddenly drops to low usage, that might signal us that the package is not well maintained anymore, or people have moved on to other solutions, and it may not make sense anymore to recommend that router.

As you mentioned, lit-router and simplr-router can indeed be considered for removal. 

Routing seems to be a very hot topic, and many people are making routers, and are hoping to add them to our list of routers which is simply not tenable. This is why we implemented the rule that in order to be considered to be listed, we'd like to have some kind of quality check in place and make sure that a router is actually used by _some_ people (our metric is not that high), and is somewhat battletested. As Lars mentioned, the list of routers should be recommendations, not a package listing.

blikblum:
Hi, just tried to make the point that a recommendation should not rely on popularity only, otherwise we all were recommending React and webpack.

slick-router is technically superior to vaadin-router in many aspects, some of them described above, but unfortunately did not get the proper advertising. I will improve docs and promote it more.

BTW: slick-router is not an "opportunistic" project born of a "let's write a router" idea of an inexperienced developer. 

It is the successor of veteran project cherrytree which has a solid routing engine (i tried to keep its development but [was not possible](https://github.com/QubitProducts/cherrytree/issues/167), so the rename).

The API design, was also influenced by the experience acquired developing and using, in actual projects, two other routing libraries (Marionette.Routing and nextbone-routing)

Finally, the project was stale when i needed a router for a web component project. I looked at open-wc routing guide. 

For each project i evaluated the code and  issues. The first ones, just looking to the code i discarded. Only one i considered was vaadin-router but the some issues (nesting routes bug, lack of link handling) and very, very opinionated design give the impulse to continue slick-router development.

I also did not bothered initially to apply it here, but looking at people asking for router advice, i think that they were loosing the opportunity to know a as good or even better library them what is commonly recommended.

thepassle:
Hi Blikblum,

> BTW: slick-router is not an "opportunistic" project born of a "let's write a router" idea of an inexperienced developer.

I absolutely did not mean to insinuate this at all, I'm sure slick-router is a perfectly capable router, and im sorry if my response came off that way.

We've had some discussion about our router recommendations yesterday, and we feel like we may have stepped outside of our bounds as open-wc as far as router recommendations go, and we'll most likely be removing router recommendations from our docs completely. The router recommendations have been somewhat of a headache to maintain since we've introduced them, and we see in our analytics that pageviews for the routers page are very low compared to other pages.

anlexN:
[open-wc](https://open-wc.org/) website use what router? i review [open-wc repo](https://github.com/open-wc/open-wc) source code, i can not find what router.

thepassle:
We dont use a router 🙂 open-wc is built with 11ty, which outputs good ol' html files

thepassle:
Since we've moved to 11ty, we also got rid of the routing recommendations for now, so I'll close this

anlexN:
what is 11ty? i can't find it in open-wc website.

thepassle:
11ty is a static site generator, you can find more information here https://www.11ty.dev/

anlexN:
oh, no. why use static site generator? you should use pure web component, html, css, javascript to make website so that demonstrate web standard's power, strong, beauty.
can you give me demos made by web standard technology?
how to make a router by myself without framework?

thepassle:
> you should use pure web component, html, css, javascript to make website

11ty gives us exactly that 🙂

> can you give me demos made by web standard technology?
> how to make a router by myself without framework?

Not sure what exactly you mean or expect from this, but I think this is not really in scope for this issue. Maybe this is better discussed on slack?

anlexN:
1. i need some example website source code written by web component to learn.
2.  how to set up router with web component?
3. where is slack channel?

PR #1424, opened 2020-03-12
Feat/rollup html improvements

LarsDenBakker:
- Allow setting index.html as regular input. This makes it easier to use our `@open-wc/building-rollup` config, as consumers can just set the rollup input and we don't need to make extra options.

- With rollup v2 the order in which build outputs are executed is random, while I was relying on them to be excuted in order. We need to know which build output to emit the generated HTML file from because they may have different output directories. I fixed this by checking if all outputs have the same directory, which is usually the case, and picking the last build to run in that case. Otherwise the user needs to specify a `buildOutputName` option to let us know which one to pick.

- In the polyfills loader when there is only one HTML plugin, we can just pick that one and not require the user to specify the HTML filename. This makes it easier to use the polyfills loader from the rollup config, since we don't know which name might be used.

### DIFF
--- a/packages/rollup-plugin-html/rollup-plugin-html.js
+++ b/packages/rollup-plugin-html/rollup-plugin-html.js
@@ -18,7 +18,12 @@ const { getInputHtmlData } = require('./src/getInputHtmlData');
 const { getEntrypointBundles } = require('./src/getEntrypointBundles');
 const { getOutputHtml } = require('./src/getOutputHtml');
 const { extractModules } = require('./src/extractModules');
-const { createError, getMainOutputDir, addRollupInput } = require('./src/utils');
+const {
+  createError,
+  getMainOutputDir,
+  addRollupInput,
+  shouldReadInputFromRollup,
+} = require('./src/utils');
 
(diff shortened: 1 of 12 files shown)


packages/rollup-plugin-html/rollup-plugin-html.js
```diff
@@ -208,10 +209,63 @@ function rollupPluginHtml(pluginOptions) {
...
+                  });
+              }
+            }
+
+            if (generatedBundles.length === multiOutputNames.length) {
+              // this is the last build, emit the HTML file
+              if (deferredEmitHtmlFile) {
+                // emit to another build output's asset tree
+                deferredEmitHtmlFile().then(() => {
+                  resolve();
```
thepassle: Is this `resolve()` redundant?

thepassle:
One question, otherwise lgtm 🙂

packages/rollup-plugin-html/rollup-plugin-html.js
LarsDenBakker: no, because the promise is not returned. this is to satisfy the linter to have a consistent return type.

Issue #1425, opened 2020-03-12
Testing with sinon


cnagadya:
The mock/stub/spies/fakes are not called. [This](https://open-wc.org/testing/testing-sinon.html) example given in the testing docs also fails
```
import { stub } from 'sinon';
import { expect, fixture, html } from '@open-wc/testing';

describe('my component', () => {
  it('calls myFunction when a button is clicked', () => {
    const el = fixture(html`
      <my-component></my-component>
    `);

    // stub a function
    const myFunctionStub = stub(el, 'myFunction');

    // click a button
    el.shadowRoot.getElementById('myButton').click();

    // check if the function was called
    expect(myFunctionStub).to.have.callCount(1);
  });
});
```

`AssertionError: expected myFunction to have been called exactly once, but it was called 0 times`

jaredmahan:
@cnagadya I am experiencing the same issue and this is currently blocking my testing. This example doesn't work in our code base.

jaredmahan:
Can someone help us out and let us know what is going on here?

thepassle:
I can reproduce, I will take a look at this today

thepassle:
We figured out why this is happening:

#1 Assuming your code looks like this, the test will pass:
```js
render() {
  return html`<button @click=${() => this.myFunction()}>click</button>`;
}
```

#2 And this will fail:
```js
render() {
  return html`<button @click=${this.myFunction}>click</button>`;
}
```

The reason is that with #2 the stubbing(/wrapping) of myFunction happens AFTER you give it to lit-html, whereas with #1 you make fixture, stub `el.myFunction`, and _then_ when you click it, it will already be wrapped.

So you can either do: `@click=${() => this.myFunction()}`, and the test as is will pass.

Or go with `@click=${this.myFunction}`, and then in your test:

```js
  it('does the thing', async () => {
    const el = await fixture(html`<my-component></my-component>`);
    const myFunctionStub = sinon.stub(el, 'myFunction');
    el.requestUpdate();
    await el.updateComplete;
    el.shadowRoot.getElementById('myButton').click();
    expect(myFunctionStub).to.have.callCount(1);
  });
```

Alternatively, instead of testing the function has been called, you could test the sideeffects of that function.

We should definitely improve the documentation on this though for sure. I will keep this issue open to track it.

jaredmahan:
Awesome thanks @thepassle for the update that worked for me! If you guys can improve the documentation around this it would help a ton!

[...]

PR #1433, opened 2020-03-17
docs: clarify sinon documentation

thepassle:
closes https://github.com/open-wc/open-wc/issues/1425

### DIFF
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -186,3 +186,77 @@ If you're interested in reading more about this, you can check out these resourc
 
 - [Why is my web component inheriting styles?](https://lamplightdev.com/blog/2019/03/26/why-is-my-web-component-inheriting-styles/)
 - [Web Fundamentals: Web Components](https://developers.google.com/web/fundamentals/web-components/shadowdom#reset)
+
+## Stubs on my element aren't called
+
+Given the following code:
+
+`my-component.js`:
+
+```js
+// example #1
+class MyComponent extends LitElement {
+  myFunction() {
+    // ...
+  }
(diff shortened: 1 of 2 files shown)


docs/faq/README.md
```diff
@@ -186,3 +186,77 @@ If you're interested in reading more about this, you can check out these resourc
...
+  );
+  const myFunctionStub = sinon.stub(el, 'myFunction');
+  el.requestUpdate();
+  await el.updateComplete;
+  el.shadowRoot.querySelector('button').click();
+  expect(myFunctionStub).to.have.callCount(1);
+});
+```
+
+- You could pass an anonymous function that calls `this.myFunction()`, like so: `@click=${() => this.myFunction()}`. Note that this usually only makes sense if you pass some arguments to the function, e.g.: `@click=${(e) => this.myFunction(e)}`, or `@click=${() => this.myFunction(someOtherData)}`
```
thepassle: @Westbrook do you think theres a better way we can explain this?

Westbrook: I'd prefer _not_ to list this as an option, and not just for the realities of that bind. Testing documentation just doesn't seem like it should be telling you how to write your functionality...

To that end, I'd suggest we list "Test Side Effects" first and then list the forced rerender approach as an "escape hatch".

thepassle: I dont consider this testing documentation as much as a technical explanation of a gotcha though, and in reality, there _can_ be uses for this kind of bind.

What if I move this point below the 'test side effects', and then add something like "As a last resort escape hatch, you could pass an anonymous ...."?

PR #1436, opened 2020-03-18
feat(rollup-plugin-html): add service worker registration

thepassle:


thepassle:
decided to do this as an inline rollup plugin in the build config instead

PR #1441, opened 2020-03-19
fix(es-dev-server): add directly depended package to package.json

### LINKED ISSUE #1439: Dependency Graph in es-dev-server seems to be missing "systemjs"
The following line in the `es-dev-server` references `systemjs`:

https://github.com/open-wc/open-wc/blob/master/packages/es-dev-server/src/utils/inject-polyfills-loader.js#L43

but, it doesn't seem to be available in the [`package.json`](https://github.com/open-wc/open-wc/blob/master/packages/es-dev-server/package.json) for that package, which I think is confusing PNPN when it attempts to resolve that package at runtime... It seems kind of made up, but when I manually copy the package into the directory involved things seem to work as expected. If this sounds like it would apply, I'd be happy to supply the PR for this, or for any related fix that you might be able to suggest for this issue.

Westbrook:
Adds dependency to package as listed in #1439. If this isn't the right path, let's discuss next steps in the issue.

fixes #1439

### DIFF
--- a/packages/es-dev-server/package.json
+++ b/packages/es-dev-server/package.json
@@ -97,6 +97,7 @@
     "polyfills-loader": "^1.5.1",
     "portfinder": "^1.0.21",
     "strip-ansi": "^5.2.0",
+    "systemjs": "^4.0.0",
     "useragent": "^2.3.0",
     "whatwg-url": "^7.0.0"
   },


LarsDenBakker:
Thanks!

Issue #1282, opened 2020-01-29
Rollup non index.html entrypoint conflicting documentation


abraham:
I'm trying to use a non index.html entrypoint and there is conflicting documentation and each initially don't seem to work. Neither example has worked for me so far.

- https://open-wc.org/building/building-rollup.html#non-index-html-entrypoint
- https://open-wc.org/building/rollup-plugin-index-html.html#non-index-html-entrypoints

daKmoR:
hey there 🤗 

what do you mean by a `non index.html entrypoint`?
e.g. use a `my-app.js`? then you will still need to define the "default html around" somewhere

this could be an example 🤗 

```js
import { createDefaultConfig } from '@open-wc/building-rollup';
import { indexHTMLPlugin } from 'rollup-plugin-index-html';
import deepmerge from 'deepmerge';

// use js as entry point => disable default indexHTML 
// (which only supports a index.html file entry point)
const basicConfig = createDefaultConfig({
  input: './my-app.js',
  plugins: {
    indexHTML: false,
  },
});

// add the html plugin yourself with some hardcoded html
export default merge(basicConfig, {
  plugins: [
    indexHTMLPlugin({
      indexHTML: `
        <html>
          <head></head>
          <body></body>
        </html>
      `,
    }),
  ],
});
```

if that does not work - can you post your errors?

abraham:
Yes, using HTML that's in a string and not using an index.html file.

Using the sample code you included and that is in my first link (and fixing the `deepmerge` vs `merge` bug) on a brand new scaffolded app. The error is "rollupPluginIndexHtml.indexHTMLPlugin is not a function"

> [!] TypeError: rollupPluginIndexHtml.indexHTMLPlugin is not a function
> TypeError: rollupPluginIndexHtml.indexHTMLPlugin is not a function
>     at Object.<anonymous> (/Users/abraham/Sandbox/rollup-html-test/app-demo/rollup.config.js:25:27)
>     at Module._compile (internal/modules/cjs/loader.js:959:30)
>     at Object.require.extensions.<computed> [as .js] (/Users/abraham/Sandbox/rollup-html-test/app-demo/node_modules/rollup/dist/bin/rollup:830:24)
>     at Module.load (internal/modules/cjs/loader.js:815:32)
>     at Function.Module._load (internal/modules/cjs/loader.js:727:14)
>     at Module.require (internal/modules/cjs/loader.js:852:19)
[...]

PR #1446, opened 2020-03-19
fix(building-rollup): ensuring input is a string before checking exte…

dlockhart:
I also ran into #1282 when using a configuration that has an array input. Something like this trips it up:

```
createDefaultConfig({
	input: ['file1.js', 'file2.js']
});
```

You get a `TypeError: _options.input.endsWith is not a function` error, since `input` in this case isn't a string, it's an array -- which is a valid Rollup thing to want to do.


LarsDenBakker:
Thank you!

PR #1447, opened 2020-03-20
chore: fix examples in Setting up a snapshot section

kavitaasiwal:
The examples in `Setting up a snapshot` section are missing the `async` keyword in function declaration. This PR fixes that.


LarsDenBakker:
Thank you!

PR #1448, opened 2020-03-20
chore: run tests on windows

LarsDenBakker:
I'm currently working on the new rollup config, and the plugins that go along with it. I want to make sure that they work on windows, so I added a windows step in the CI.

A lot of tests fail on windows, not necessarily because of bad code but often because of path and newline differences. For now I scoped it to the packages I'm working on, over time we can add more packages.

### DIFF
--- a/packages/rollup-plugin-html/test/src/extractModules.test.js
+++ b/packages/rollup-plugin-html/test/src/extractModules.test.js
@@ -1,4 +1,5 @@
 const { expect } = require('chai');
+const path = require('path');
 const { extractModules } = require('../../src/extractModules');
 
 describe('extractModules()', () => {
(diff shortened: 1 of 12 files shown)


packages/rollup-plugin-html/test/src/extractModules.test.js
```diff
@@ -59,7 +63,10 @@ describe('extractModules()', () => {
     );
 
     expect(inlineModules.size).to.equal(0);
-    expect(moduleImports).to.eql(['/base-1/base-2/foo.js', '/base-1/bar.js']);
+    expect(moduleImports).to.eql([
+      `${path.sep}base-1${path.sep}base-2${path.sep}foo.js`,
```
bennypowers, line 67: small nit, but I might have preferred path.resolve or some other `f(['base-1', 'base-2', 'foo.js'])` path building function here

PR #1452, opened 2020-03-22
chore: use "bugfixes: true" in babel-preset-env

Westbrook:
Using `bugfixes: true` as available in `@babel/preset-env@7.9.0` https://babeljs.io/blog/2020/03/16/7.9.0 will help to ensure that the smallest possible output when building with Babel in es-dev-server, Rollup, webpack, and more!


Westbrook:
@LarsDenBakker these tests that don't pass after the update (https://app.circleci.com/pipelines/github/open-wc/open-wc/5418/workflows/af48d8a7-d752-497c-87de-c0c9832a1e8e/jobs/6951/parallel-runs/0/steps/0-104), I expect them to be different, that's sort of the win here, but is there a way I can know the output still serves the purpose of the test before I update the snapshots?

LarsDenBakker:
Yeah those need to be updated. Snapshots can be updated with yarn test:updates-snapshots, but there are also compatibility mode e2e tests which check if the babel compile did what it needed to do. They're a bit annoying to maintain, but valueable :)

Westbrook:
As long as they're backed up in the e2es so that I don't have to parse them for "correctness" as much as make sure they're tuned to pass, then I'm fine with it. Got this change passing CI and am looking forward to doing some before and afters of the possible byte savings...

packages/es-dev-server/test/middleware/compatibility-transform.test.js
```diff
@@ -117,7 +117,7 @@ describe('compatibility transform middleware', () => {
     it('does not do any transformation for standard features', async () => {
       await expectCompatibilityTransform(userAgents['Chrome 78'], {
         objectSpread: false,
-        templateLiteral: true,
+        templateLiteral: false,
```
Westbrook: @LarsDenBakker the three changes in this file are the ones I'd be the most obliged that you confirmed. I think this falls into the realm of "there are e2e tests for that" as I understood your previous comment, but wanted to double-check as everything else I changes was just package numbers and API properties.

LarsDenBakker: This is a correct and expected cange!

Issue #1462, opened 2020-03-27
[scoped-elements] Doesn't register the element when using lit-html "until" directive. 


carloviviano:

LarsDenBakker:
Do you have some more information? Can you create a reproduction with  https://webcomponents.dev/create/issue?

manolakis:
I'm trying to reproduce it in webcomponents. Seems a problem with promises.

manolakis:
Reproduction https://webcomponents.dev/edit/yWaR2kJsf06QNnw6nUBv

It's not scoping the element when the promise is resolved

PR #1464, opened 2020-03-28
fix(scoped-elements): elements not scoped using directives

manolakis:
Instead transforming the ResultTemplates before render, this PR creates a scopedElementsTemplateFactory that is going to be used also by directives.

fixes: #1462

### DIFF
--- a/packages/scoped-elements/src/shadyTemplateFactory.js
+++ b/packages/scoped-elements/src/shadyTemplateFactory.js
@@ -1,0 +1,47 @@
+import { templateCaches } from 'lit-html/lib/template-factory.js';
+import { marker, Template } from 'lit-html/lib/template.js';
+
+const getTemplateCacheKey = (type, scopeName) => `${type}--${scopeName}`;
+
+let compatibleShadyCSSVersion = true;
+
+// @ts-ignore
+const { ShadyCSS } = window;
+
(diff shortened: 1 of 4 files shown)


packages/scoped-elements/src/shadyTemplateFactory.js
```diff
@@ -0,0 +1,47 @@
...
+  compatibleShadyCSSVersion = false;
+} else if (typeof ShadyCSS.prepareTemplateDom === 'undefined') {
+  compatibleShadyCSSVersion = false;
+}
+
+/**
+ * Template factory which scopes template DOM using ShadyCSS.
+ * @param scopeName {string}
+ */
+export const shadyTemplateFactory = scopeName => result => {
```
manolakis: It would be great if lit-html exported this function. By the moment this is a workaround to continue supporting shadyCSS

LarsDenBakker: Can we file a feature request at lit-html to expose this?

manolakis: https://github.com/Polymer/lit-html/issues/1134

LarsDenBakker: Great!

PR #1467, opened 2020-03-28
docs: updates rollup plugin url and usage

gangadharjannu:
# Updates rollup package documentation
* Updates URL for commonjs plugin. Link for the rollup plugin commonjs is outdated as recently it has been moved into scoped packages.
* Updates code example.
* Also adds additional information on why do we want to use commonjs module.


packages/building-rollup/README.md
```diff
@@ -268,11 +268,11 @@ export default deepmerge(basicConfig, {
 
 #### Resolve commonjs modules
 
-CommonJS is the module format for NodeJS, and not suitable for the browser. Rollup only handles es modules by default, but sometimes it's necessary to be able to import a dependency. To do this, you can add [rollup-plugin-commonjs](https://github.com/rollup/rollup-plugin-commonjs):
+CommonJS is the module format for NodeJS, and not suitable for the browser. Rollup only handles es modules by default, but sometimes it's necessary to be able to import a dependency if you want to use third party libraries in your web component. To do this, you can add [@rollup/plugin-commonjs](https://github.com/rollup/plugins/tree/master/packages/commonjs):
```
LarsDenBakker: I don't think adding this sentence is correct, not all third party libraries use commonjs.

LarsDenBakker:
Thanks for the contribution!

packages/building-rollup/README.md
gangadharjannu: I explicitly added this line because I struggled to add a third party library however, you are correct. May be we can rewrite this as 'if you want to use third party libraries which only supports commonjs'.

LarsDenBakker: Yes that would be fine.

gangadharjannu: Ok, then I will update the content

gangadharjannu: I just updated with the discussed change.

LarsDenBakker:
We had some other pending updates to rollup, and  the docs could use some improvements in general. I've picked up this change in https://github.com/open-wc/open-wc/pull/1500.

PR #1469, opened 2020-03-29
feat(create): copyTemplates resolves with the copied files

jorenbroekema:
This change allows me to do the following:

```js
participants.forEach(name => {
  copyTemplates(
    'templates/**/*',
    path.resolve(process.cwd(), `./participants/${name}`),
    { participantName: name },
  ).then(files => {
    files.forEach(file => {
      writeFileToPathOnDisk(file.toPath, file.processed, {
        override: opts.override, // option that gets passed from my middleware, sometimes I need to override, sometimes I don't :P
        ask: false,
      });
    });
  });
});
```

Which is super awesome for generic purposes where I don't need a cli or the other cool features that, I assume, made you guys opt for a `virtualFiles` array that keeps the state in between functions.

Let me know if the test makes sense.. I couldn't really figure out a way to test copyTemplates without creating an actual template for testing. But perhaps you guys have a helper for this kinda thing?

### DIFF
--- a/packages/create/src/core.js
+++ b/packages/create/src/core.js
@@ -1,13 +1,11 @@
 /* eslint-disable no-console, import/no-cycle */
 
-import fs from 'fs';
-import path from 'path';
 import { spawn } from 'child_process';
-
-import prompts from 'prompts';
-import glob from 'glob';
 import deepmerge from 'deepmerge';
-
+import fs from 'fs';
+import glob from 'glob';
+import path from 'path';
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
Looks good to me, @daKmoR double check please?

daKmoR:
not really the usecase it was built for 😅

but code looks fine so we can merge it 👍

daKmoR:
thaaaankkksss 🤗

PR #1470, opened 2020-03-29
fix(testing-helpers): constrain the type of defineCE

bennypowers:
it should take a subclass of HTMLElement

### DIFF
--- a/packages/testing-helpers/src/helpers.js
+++ b/packages/testing-helpers/src/helpers.js
@@ -11,12 +11,12 @@ let defineCECounter = 0;
  * const el = fixture(`<${tag}></${tag}>`);
  * // test el
  *
- * @param {typeof HTMLElement | function} klass Class which extends HTMLElement
+ * @template {HTMLElement} T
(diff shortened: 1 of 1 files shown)


packages/testing-helpers/src/helpers.js
```diff
@@ -11,12 +11,12 @@ let defineCECounter = 0;
  * const el = fixture(`<${tag}></${tag}>`);
  * // test el
  *
- * @param {typeof HTMLElement | function} klass Class which extends HTMLElement
+ * @template {HTMLElement} T
+ * @param {import("@open-wc/dedupe-mixin").Constructor<T>} klass Class which extends HTMLElement
```
daKmoR, line 15: wouldn't that require a hard dependency on `@open-wc/dedupe-mixin`?

bennypowers: LitElement also exports this type

bennypowers: also, that hard dependency is introduced by scopedelementstestwrapper

PR #1479, opened 2020-03-31
docs(scoped-elements): update documentation

manolakis:


packages/scoped-elements/README.md
```diff
@@ -2,88 +2,7 @@
 
 [//]: # 'AUTO INSERT HEADER PREPUBLISH'
 
-:::warning
-This is an experimental feature, use it at your own risk and be sure to understand its [limitations](#limitations).
-No big applications are using `scoped-elements` yet - so there is no proof yet if it works in production.
-This page focuses on in depth explanations as it should help [foster a discussion on scoping](https://github.com/open-wc/open-wc/issues/1262).
-:::
-
-Complex Web Component applications are often developed by several teams across organizations. In that scenario it is common that shared component libraries are used by teams to create a homogeneous look and feel or just to avoid creating the same components multiple times, but as those libraries evolve problems between different versions of the same library may appear, as teams may not be able to evolve and update their code at the same velocity. This causes bottlenecks in software delivery that should be managed by the teams and complex build systems, to try to alleviate the problem.
```
LarsDenBakker, line 11: Can we keep the text from this paragraph, down to....

packages/scoped-elements/README.md
```diff
@@ -2,88 +2,7 @@
...
-
-[Scoped Custom Element Registries](https://github.com/w3c/webcomponents/issues/716) is a proposal that will solve this problem, but until it is ready, or a polyfill becomes available, we have to _scope_ custom element tag names if we want to use different versions of those custom elements in our code. This package allows you to forget about how custom elements are defined, registered and scopes their tag names if it is necessary, and avoids the name collision problem.
-
-## Installation
-
-```bash
-npm i --save @open-wc/scoped-elements
-```
-
```
LarsDenBakker, line 20: ... here?

packages/scoped-elements/README.md
```diff
@@ -166,6 +85,141 @@ export class MyElement extends ScopedElementsMixin(LitElement) {
...
+    // do whatever you need with the scopedTagName
+  }
+
+  // ...
+}
+```
+
+## Motivation
+
+Complex Web Component applications are often developed by several teams across organizations. In that scenario it is common that shared component libraries are used by teams to create a homogeneous look and feel or just to avoid creating the same components multiple times, but as those libraries evolve problems between different versions of the same library may appear, as teams may not be able to evolve and update their code at the same velocity. This causes bottlenecks in software delivery that should be managed by the teams and complex build systems, to try to alleviate the problem.
```
manolakis, line 152: @LarsDenBakker I didn't removed the text, I just moved it here, after the API docs and I title it as Motivation.

I removed `How to install` as we don't have it in any other docs.

LarsDenBakker: A small explanation of what the library does I think is good as a start, but indeed we should move most of the in-depth explanation to the bottom.

manolakis: Ok. I just restored the installation and I add a little explanation at the top. Let me know if it's better now.

thanks!

daKmoR:
thaaankkks guys 😍

PR #1491, opened 2020-04-05
fix(create): set default value for variable

amitport:
Previous code set a specific variable value on `:host`, which does not make sense for variables since we probably want to allow overrides from ancestors.


LarsDenBakker:
Makes sense, thanks a lot!

PR #1496, opened 2020-04-08
fix: arrow function to function for legacy browser

qpitlove:
- avoid don't support arrow function syntax in legacy browser

### DIFF
--- a/packages/building-utils/index-html/create-index-html.js
+++ b/packages/building-utils/index-html/create-index-html.js
@@ -115,7 +115,7 @@ function createScripts(polyfillsConfig, polyfills, entries, needsLoader) {
     scripts.push(
       createScript(
         null,
-        "window.importShim = src => import(src.startsWith('.') ? new URL(src, document.baseURI) : src)",
+        "window.importShim = function(src) { return import(src.startsWith('.') ? new URL(src, document.baseURI) : src); }",
       ),
     );
   }


LarsDenBakker:
These browsers dont support dynamic imports either :)

qpitlove:
@LarsDenBakker 
### Step by Step
1. `npm init @open-wc`, and select rollup setup 
2. update rollup.config.js `createCompatibilityConfig` for IE11 Support
```
import { createCompatibilityConfig } from '@open-wc/building-rollup';
export default createCompatibilityConfig({ input: './demo/index.html' });
```
3. after `npm run build`, below image is shown. `(`dist/index.html` )`
image

legacy browser ( like`android 4.4 default brower` ) or ie 11 below version will be exception, and stop all script eval

please check it again.

LarsDenBakker:
My point is that even if you fix the arrow function, the script still won't work because IE11 doesn't support dynamic imports. 

When it throws, only execution of that small script is stopped. This is intentional, it's there to feature detect dynamic import which is only possible in this way.

qpitlove:
Yes. i know what you said.
But when I use with `createCompatibilityConfig` options and build and exec in ie 11, small exeception is always occurred. 

As caniuse site. Arrow function is not support 
https://caniuse.com/#search=Arrow%20function

I just don't want to show exception. That's all

LarsDenBakker:
As I explained, this is the only way to feature detect dynamic imports. If you make it an arrow function, there is still an error thrown.

LarsDenBakker:
With https://github.com/open-wc/open-wc/pull/1500 the feature detection is moved inside a dynamic function that is created, having the same effect but hiding the error.

qpitlove:
@LarsDenBakker 
ok Thanks.

PR #1500, opened 2020-04-11
feat(building-rollup): refactor and update to rollup v2

LarsDenBakker:
TODO:

- [x] Typescript demo
- [x] Review
- [x] Update storybook
- [x] clientsClaim etc. for service worker (@thepassle ?)
- [x] Create v1 for rollup plugins
- [x] breaking changes for storybook
- [x] Fix storybook rollup warnings


thepassle:
Re: serviceworker, I think a good default would be:

- apply sw by default false
- skipWaiting and clientsClaim true (workbox config)

So the user gets the service worker either way, and if they want to actually make use of it, they only have to toggle a boolean in the config

packages/building-rollup/src/createSpaConfig.js
```diff
@@ -0,0 +1,95 @@
...
+          userOptions.workbox,
+          {
+            globIgnores: ['legacy-*.js'],
+            navigateFallback: '/index.html',
+            // where to output the generated sw
+            swDest: path.join(process.cwd(), outputDir, 'sw.js'),
+            // directory to match patterns against to be precached
+            globDirectory: path.join(process.cwd(), outputDir),
+            // cache any html js and css by default
+            globPatterns: ['**/*.{html,js,css}'],
```
thepassle, line 87:
```suggestion
            globPatterns: ['**/*.{html,js,css}'],
            skipWaiting: true,
            clientsClaim: true
```

packages/building-rollup/test/integration.test.js
```diff
@@ -36,56 +35,53 @@ describe('integration tests', () => {
     );
   });
 
-  testCases.forEach(testCase => {
-    ['modern', 'modern-and-legacy'].forEach(mode => {
-      describe(`testcase ${testCase}-${mode}`, function describe() {
-        this.timeout(10000);
-        let page;
+  // ['js/rollup.spa.config.js', 'js/rollup.spa-nomodule.config.js'].forEach(testCase => {
```
thepassle, line 38: can this comment be removed?

packages/building-rollup/README.md
```diff
@@ -1,151 +1,164 @@
...
+// use createBasicConfig to do regular JS to JS bundling
+// import { createBasicConfig } from '@open-wc/building-rollup';
+
+const baseConfig = createSpaConfig({
+  // set legacyBuild to add an additional build for build legacy,
+  // the build for modern browsers stays the same
+  // legacyBuild: 'nomodule',
+  // development mode creates a non-minified build
+  developmentMode: process.env.ROLLUP_WATCH === 'true',
+  // set to true to inject service worker into HTML
```
thepassle, line 53:
```suggestion
  // set to true to inject the service worker registration into the index.html
```

packages/building-rollup/README.md
```diff
@@ -1,151 +1,164 @@
...
-  - Both:
-    - resolve bare imports ( `import { html } from 'lit-html'` )
-    - preserve `import.meta.url` value from before bundling
-    - load polyfills when needed:
-      - dynamic import
-      - webcomponents
-      - fetch
-    - minify + treeshake js
-    - minify html and css in template literals
+When you use `createSpaConfig` a service worker is generated automatically. It is not injected into your `index.html` by default to prevent unexpected results. You can turn this on by enabling the `injectServiceWorker` option:
```
thepassle, line 139:
```suggestion
When you use `createSpaConfig` a service worker is generated automatically with Workbox. The service worker registration is not injected into your `index.html` by default to prevent unexpected results. You can turn this on by enabling the `injectServiceWorker` option:
```

packages/building-rollup/README.md
```diff
@@ -155,342 +168,209 @@ For example to add support for class properties:
 }
 ```
 
-## Extending the rollup config
+## Customizations
```
thepassle, line 171: Maybe we can 'go the extra mile' here and explain that the `createConfig` functions really just return a rollup configuration object that people can still manipulate when they need to add/customize stuff. Maybe it helps remove some of the 'magic'?

EDIT: nvm, you explain this further down below

packages/building-rollup/README.md
```diff
@@ -155,342 +168,209 @@ For example to add support for class properties:
...
 
-A rollup config is just a plain object. It's easy to extend it using javascript:
+Our config generator sets you up with good defaults for most projects. It's easy to extend and customize this config further for your requirements.
 
-```javascript
-import { createDefaultConfig } from '@open-wc/building-rollup';
+If you find yourself disabling a lot of the default functionality we recommend forking from the default config and taking control yourself. Rollup is not that hard to configure, it's better to be in full control if you know what you're doing.
```
thepassle, line 175:
```suggestion
If you find yourself disabling a lot of the default functionality we recommend forking from the default config and taking control yourself. Rollup is relatively easy to configure compared to other build tools, it's better to be in full control if you know what you're doing.
```

packages/building-rollup/README.md
```diff
@@ -155,342 +168,209 @@ For example to add support for class properties:
...
+    myPlugin(),
+  ],
+});
 ```
 
-### Overriding the workbox config
+</details>
 
-If you want to override the default config with your own workbox configuration, you can disable the default workbox configuration by setting `options.plugins.workbox` to false in the `options` object that you pass to `createDefaultConfig`, and then you can override the plugins
+#### Support for Typescript
```
thepassle, line 332:
```suggestion
#### Support for Commonjs
```

packages/building-rollup/test/integration.test.js
LarsDenBakker: No, need to re enable this test.

abdonrd:
What is the `Typescript demo` from the TODO?

An TS application demo in the CLI?

LarsDenBakker:
In the `building-rollup` package we have some demos to do a manual regression test whether things work correctly.

LarsDenBakker:
@daKmoR will do a final review of the storybook changes

daKmoR:
as this will release basically all packages this now also includes

- [chore: move most devDependencies to the root](https://github.com/open-wc/open-wc/pull/1500/commits/04a58460ffca3ba0addeb662f70d8857568e7421)

packages/building-rollup/src/createSpaConfig.js
```diff
@@ -0,0 +1,97 @@
...
+            skipWaiting: true,
+            clientsClaim: true,
+          },
```
abdonrd, line 90: @LarsDenBakker @thepassle about the default Workbox config... 

- Why the `globIgnores: ['legacy-*.js'],`?
- Why the `skipWaiting` and `clientsClaim` as `true`?

thepassle: legacy browsers dont support service worker, so it doesnt make sense to cache the legacy build

skipwaiting and clientsclaim because its the safest default for people getting started with service workers

abdonrd: - But now the legacy build are with the `nomodule-hash.js` files, right?

- Ok, thanks! Playing with your `pwa-update-available` I have a few issues with this default config.

PR #1501, opened 2020-04-11
chore: run yarn-deduplicate command

web-padawan:
TIL about [yarn-deduplicate](https://github.com/atlassian/yarn-deduplicate) tool, so here is a PR with some `yarn.lock` cleanup.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -32,16 +32,7 @@
   dependencies:
     "@babel/highlight" "^7.8.3"
 
-"@babel/compat-data@^7.8.6":
-  version "7.8.6"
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
Thanks, looks pretty useful. We did some big changes on master now, so there are conflicts. Is this something in yarn itself, or is it a separate project?

web-padawan:
@LarsDenBakker rebased and force-pushed. 

This is a separate tool by Atlassian, not part of yarn itself. You can run it like this:

```
npx yarn-deduplicate yarn.lock
```

LarsDenBakker:
Unfortunately it's causes a change in snapshots, we have file hashes in snapshots which is pretty cumbersome. Looking to get red of it.

LarsDenBakker:
I ran the tool, did some cleanup and updates the snapshots in https://github.com/open-wc/open-wc/pull/1509

Will close hthis one, Thanks for the tip!

Issue #1371, opened 2020-02-24
[building-rollup] workbox add runtime caching for polyfills


thepassle:
as discussed: https://github.com/open-wc/open-wc/issues/1360

We should probably add runtime caching for anything in `polyfills/`, because it doesnt make sense to aggressively precache them when they may not all be used. Using runtime caching, we can cache them on demand.

Investigate if we can make this a non-breaking change, if not, it should go with https://github.com/open-wc/open-wc/issues/1153

LarsDenBakker:
@thepassle Was this implemented?

PR #1504, opened 2020-04-12
feat(building-rollup): runtime cache polyfills

thepassle:
closes https://github.com/open-wc/open-wc/issues/1371


daKmoR:
I am just curious - is there any way we could actually test this? I have now clue on how to  test service worker things without actually running a browser 🤔 but as there can only be one service worker per folder... that would be crazy to test...

how does workbox does it?

LarsDenBakker:
@daKmoR maybe with puppeteer? Workbox is mostly configuration, so I'm not *too* worried about that.

daKmoR:
agreed 💪

Issue #1473, opened 2020-03-30
eslint-config-prettier: bump to Prettier 2.0.2


theolavaux:
Any plans on this ?
Here are the changes: [Prettier 2.0](https://prettier.io/blog/2020/03/21/2.0.0.html)

PR #1525, opened 2020-04-16
Chore/update prettier

LarsDenBakker:
- Deprecates `@open-wc/prettier-config`, recommending to use prettier v2 directly
- Updates `create` to use prettier v2
- Update docs to reflect changes to prettier v2
- Format whole repository with prettier v2

Fixes https://github.com/open-wc/open-wc/issues/1473

Because of the linting configuration in our repository, I did have to bump prettier to v2 in pretter-config package.


docs/faq/README.md
```diff
@@ -200,9 +200,7 @@ class MyComponent extends LitElement {
     // ...
   }
   render() {
-    return html`
-      <button @click=${this.myFunction}>click</button>
-    `;
+    return html` <button @click=${this.myFunction}>click</button> `;
```
Westbrook, line 203:
```suggestion
    return html`<button @click=${this.myFunction}>click</button>`;
```

Nit: the way this collapsed the previous white space, there are now space at the beginning and end of all the templates. It's the _same_, but is it what we _want_/what we want to exhibit for others?

LarsDenBakker: These will need to be trimmed manually groomed now indeed :)

PR #1532, opened 2020-04-18
chore: undo wrong releases

daKmoR:


daKmoR:
we force reset master instead

PR #1542, opened 2020-04-20
fix(storybook-addon-markdown-docs): no babel user config for jsxToJs

daKmoR:


LarsDenBakker:
Thanks for your contribution!

Issue #1541, opened 2020-04-20
[scoped-elements] Wrongfully replaces non-scoped tags


lazka:
Using @open-wc/scoped-elements@1.0.8

Example:

```js
import {html, LitElement} from 'lit-element';
import { ScopedElementsMixin } from '@open-wc/scoped-elements';

class Button extends LitElement {
    render() {
        return html`button`;
    }
}

class OtherThing extends LitElement {
    render() {
        return html`other thing`;
    }
}

customElements.define('my-other-thing', OtherThing);

class Container extends ScopedElementsMixin(LitElement) {

    constructor() {
        super();
    }

    static get scopedElements() {
        return {
          'my-button': Button,
        };
      }

    render() {
        return html`
            Button: <my-button></my-button>
            OtherThing: <my-other-thing></my-other-thing>
        `;
    }
}
```

This results in

```html
Button: <my-button-55790></my-button-55790>
OtherThing: <my-other-thing-55789></my-other-thing-55789>
```

I would have expected instead

```html
Button: <my-button-55790></my-button-55790>
OtherThing: <my-other-thing></my-other-thing>
```

because I've never set `my-other-thing` as a scoped element and it's defined somewhere else.

I've worked around this for now by doing

```js
static get scopedElements() {
    return {
        'my-button': Button,
        'my-other-thing': customElements.get('my-other-thing'),
        'my-other-thing2': customElements.get('my-other-thing2'),
        'my-other-thing3': customElements.get('my-other-thing3'),
        'my-other-thing4': customElements.get('my-other-thing4'),
    };
}
```

but that's ugly and error prone..

lazka:
In case this is in purpose, my use case is that I'm porting things one by one to scoped elements, and having this all or nothing makes things complicated.

daKmoR:
hey @lazka 👋

this is indeed on purpose. The currently proposed spec says every scoped registry is "new". There is no inheritance from the parent or global registry.

Hence all components you want to use in your scoped registry need to be registered.

We definitely should make this more clear in the docs 😅

@manolakis can you make sure that we have a "warning/info" in the docs about it... and that 
```
'my-other-thing': customElements.get('my-other-thing'),
```
is the recommended way to "reregister" from the global registry if needed.

lazka:
Thanks, having this in the docs like you described would have helped me.

PR #1553, opened 2020-04-23
feat(scoped-elements): self-registering components compatibility

manolakis:
Reusing the global tag name can be useful in some situations in which not all component libraries are exporting the component classes or when components are looking in dom for specific tag names. In those cases, although this change doesn't remove the limitation of having just one version of those elements in our app, it makes them work properly inside our scoped elements as we are going to use the tag name they use in their definition.

E.g. If we want to migrate our app to scoped-elements but we are using `vaadin-grid` we are going to have problems because of vaadin components. We already have two options:

1- Ask or make a PR against vaadin components to change the way they interact with other components and making them "scoped friendly" (in the future, with scoped custom element registries users could use any tag name inside their components, so custom elements shouldn't impose a specific one).
2- Wrap the vaadin components into another unscoped component. This workaround, although it is feasible, sometimes could make our apps hard to migrate to scoped elements.

With this change, we could continue using those unscoped friendly libraries eventually while we search a feasible scoped friendly substitution for them or they release a new scoped friendly version.

### How to use it

```js
// file item-a.js
// unscoped friendly component
class ItemA extends LitElement {
   render() {
     return html` <div>Item A</div> `;
  }
}

customElements.define('item-a', ItemA);


// file ContainerElement.js
import './item-a.js'; // imports the unscoped friendly component

const ItemA = customElements.get('item-a');  // we must obtain the component class reference

export class ContainerElement extends ScopedElementsMixin(LitElement) {
  static get scopedElements() {
    return {
      ...super.scopedElements,
      'item-a': ItemA,  // <-- If element is not declared here the tag would be scoped as usual
    };
  }

  render() {
    return html` <item-a></item-a> `; // <-- the tag in DOM would be 'item-a'
  }
}
```

There is a limitation to this solution. **Self-defined components can't be lazily defined.**

### DIFF
--- a/packages/scoped-elements/README.md
+++ b/packages/scoped-elements/README.md
@@ -50,6 +50,8 @@ npm i --save @open-wc/scoped-elements
    >    };
    >  }
    > ```
+   >
+   > If you try to register the same element globally AND locally with the exact same name AND class instance it will reuse the global tag name and NOT scope it.
 
 4. Use your components in your html.
 
--- a/packages/scoped-elements/src/registerElement.js
+++ b/packages/scoped-elements/src/registerElement.js
@@ -22,6 +22,26 @@ const defineElement = (tagName, klass, registry = customElements) => {
 };
 
 /**
+ * Stores a lazy element in the cache to be used in future
+ *
+ * @param {string} tagName
+ * @param {CustomElementRegistry} registry
+ * @param {Map<string, string>} tagsCache
+ * @returns {string}
+ */
+const storeLazyElementInCache = (tagName, registry, tagsCache) => {
+  const tag = createUniqueTag(tagName, registry);
+
+  if (!tagsCache) {
+    throw new Error('Lazy scoped elements requires the use of tags cache');
+  }
+
+  tagsCache.set(tagName, tag);
+
+  return tag;
+};
+
+/**
  * Define a scoped custom element storing the scoped tag name in the cache
  *
  * @param {string} tagName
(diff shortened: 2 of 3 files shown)


LarsDenBakker:
I think it's a clever way to tackle this problem, I'm in favor of this solution.

Can we add a clarification in the docs why this extra check is needed.

mihaisavezi:
Can I ask why the default behaviour is to scope all elements in the component, and not just the ones defined in scopedElements static method ?

```
export class MyElement extends ScopedElementsMixin(LitElement) {
  static get scopedElements() {
    return {
      'my-button': MyButton,
    };
  }

  connectedCallback() {
    super.connectedCallback();
    this.addEventListener('onMenuItemClicked', this.onMenuItemClicked);
  }

  render() {

    return html`
      <div class="home-page">
        <div class="home-page__left_container">
          <div class="home-page__title">
            <my-button @click=${() => {'click on my custom button'}}>TEXT</my-button>
            <global-defined-checkbox
              id="myEntries"
              @model-value-changed=${onCheckboxChanged}
              .checked=${isCheckboxChecked}
            ></global-defined-checkbox>
          </div>
      </div>
    `;
  }
}
```

Here i expect 'my-button' to be scoped but not my `global-defined-checkbox`. But they are both scoped, and unless I have my `global-defined-checkbox` included in the scopedElements method, it won't work.

The problem is I have a dozen or so global custom components just like `global-defined-checkbox` and including them one by one seems the opposite of how this should be working.

Was this a decision or is it a constraint ?

mihaisavezi:
Ok, I see this is trying to follow the spec as closely as possible and this is why this works this way.

https://github.com/open-wc/open-wc/issues/1541

PR #1554, opened 2020-04-23
feat(building-rollup): support overriding built-in babel plugins

LarsDenBakker:
This allows overriding the configuration of babel plugins we add in the config in a lightweight manner. For example:

```js
const baseConfig = createSpaConfig({
  babel: {
    plugins: [
      [
        require.resolve('babel-plugin-template-html-minifier'),
        {
          modules: {
            'cool-html': ['html'],
          },
          htmlMinifier: {
            removeComments: false,
          },
        },
      ],
    ],
  },
});
```


packages/building-rollup/src/utils.js
```diff
@@ -6,6 +6,28 @@ const Terser = require('terser');
...
+    return undefined;
+  }
+
+  const config = merge(defaultConfig, typeof userConfig === 'object' ? userConfig : {});
+
+  const newPlugins = [];
+  const addedPlugins = new Set();
+  for (const plugin of [...config.plugins].reverse()) {
+    const name = Array.isArray(plugin) ? plugin[0] : plugin;
+    const resolvedName = require.resolve(name);
```
daKmoR, line 20: smart move 👍

PR #1555, opened 2020-04-23
fix(building-rollup): don't read user .browserslist

LarsDenBakker:
We take quite some assumptions in our config about browser support for the two different build outputs. Reading a user's browserslist is incompatible with that and can produce unexpected results or bugs.


bennypowers:
I'm uncomfortable with clobbering user config without logging a warning

LarsDenBakker:
What do you mean by clobbering?

Basically this is a bug, we read a user's browserslist which can completely break the application.

bennypowers:
To elaborate, I think a user's basic assumption will (and **SHOULD**) be that "if I specify a `.browserslist`, tools will respect it"

What I understood from your messages is that you think that would be a bug for this tool.

So my comment is not on design principles for `building-rollup`, it's on user expectations. If indeed you think `building-rollup` should in-all-cases-and-on-purpose ignore (in other words clobber) a user's config - and it's fine to take that approach - I think we should print a warning to the console to the effect that:
> You are using a `.browserslist` config file, which `@open-wc/building-rollup` ignores. see https://owc.url/4jkd87 for more info

(and on that explainer doc we could give instruction on how to turn off the warning)

I'm certain that otherwise we're inviting our users to spend hours and hours debugging their toolchains when it was actually a design decision on our part to ignore their config.

Please correct me if I misunderstood anything

LarsDenBakker:
I see what you mean now, but I'm not sure that's a reasonable expectation. A browserslist is a very crude metric, not all tools can produce a valid outcome based on a browserslist alone. Especially when it comes to es modules, polyfilling browsers features etc. 

The main clash is that we recommend users to create a separate for older browsers which don't support modern features. This means that we can't work with a single browserslist, and when people add one for eslint it breaks the build.

PR #1556, opened 2020-04-23
fix(testing-karma): fixes test coverage reporter

### LINKED ISSUE #1164: [testing-karma] switch to karma-coverage
We are using `karma-coverage-istanbul-reporter`, but it's using an outdated istanbul API and doesn't accept community feature requests / contributions unless you are a donator. [karma-coverage](https://github.com/karma-runner/karma-coverage) seems to be a good replacement.

ryuhhnn:
move from karma-coverage-istanbul-reporter to karma-coverage

Closes #1164


ryuhhnn:
This would be a breaking change since the `coverageIstanbulReporter` object in the karma config would now become `coverageReporter`. There are also a couple of other API changes to this. Since it's hard to figure out the last version of this package that worked, I'm unsure if the output of the coverage reports has regressed, so do let me know if I need to make any changes!

coreyfarrell:
packages/karma-esm/package.json also uses karma-coverage-istanbul-reporter as a devDependency.  I think this is the only package which uses old versions of istanbul libs but it's difficult to know for sure without eliminating that dependency.  This also causes weird hoisting in the monorepo.

bennypowers:
This is indeed a breaking change, and [commit message should reflect that](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#summary)

Docs should also be updated inline and on the site

Thanks :D

ryuhhnn:
@coreyfarrell addressed the karma-esm package

@bennypowers added in the BREAKING CHANGE message to commit and updated documentation

Let me know if I missed anything else!

yarn.lock
```diff
@@ -11010,36 +11041,15 @@ isstream@~0.1.2:
...
-    istanbul-lib-instrument "^3.3.0"
-    istanbul-lib-report "^2.0.8"
-    istanbul-lib-source-maps "^3.0.6"
-    istanbul-reports "^2.2.4"
-    js-yaml "^3.13.1"
-    make-dir "^2.1.0"
-    minimatch "^3.0.4"
-    once "^1.4.0"
-
 istanbul-lib-coverage@^2.0.5:
```
coreyfarrell, line 11044: I'm not currently at a computer so I cannot see why this is still installed but it is concerning as 2.x is incompatible with 3.x.  I'll have a closer look once I can but wanted to mention this as soon as I saw it.

coreyfarrell: Sorry I missed it in the first look, karma-esm also depends on babel-plugin-istanbul@5.  This needs to be upgraded to v6.

ryuhhnn: Upgrading babel-plugin-istanbul to 6.0.0 in the karma-esm package works, but `@storybook/addon-docs` depends on `@jest/transform` which relies on `babel-plugin-istanbul@^5.1.0` and it looks like storybook addon docs hasn't been updated yet. I'm unsure how to proceed here.

LarsDenBakker: I'll see if we can lose this dependency

LarsDenBakker: https://github.com/open-wc/open-wc/pull/1590 removes the dependency

LarsDenBakker:
Do you think it is at all possible to avoid a breaking change? Specifically, could we carry over the coverage thresholds from the old config?

bennypowers:
> Do you think it is at all possible to avoid a breaking change? Specifically, could we carry over the coverage thresholds from the old config?

with the same thresholds, it's still a breaking change, unless we copy `coverageIstanbulReporter` options to `coverageReporter` options

bennypowers:
I'd still like to see this land

ryuhhnn:
My apologies, I've been caught up in actual work work, so haven't had much time to address this. I should have some time this week to make this backwards compatible!

packages/karma-esm/demo/basic/karma.config.js
```diff
@@ -78,11 +78,9 @@ module.exports = config => {
       },
     },
 
-    coverageIstanbulReporter: {
-      reports: ['html', 'lcovonly', 'text-summary'],
+    coverageReporter: {
+      reporters: [{ type: 'html' }, { type: 'lcovonly' }, { type: 'text-summary' }],
```
Westbrook, line 82: This might be the extend of it, but should we list a `migration.md` file in this PR so that users can easily make the version change here?

ryuhhnn: `migration.md` file added. Hope I put it in the right place

bennypowers:
Is this ready to go?

LarsDenBakker:
Looks good to me!

pndewit:
I'd love to see this land a.s.a.p. 👍 
I'm currently running into issues that my tests are ran twice (or at least reports twice). As far as I can see this change is fixing my issues too.

NOTE:
I noticed it when I add the following pieces of code to one of my test files:
```
  before(() => console.log('before')); // Will print twice
```

or...

```
Object.keys(window.__karma__.files).forEach(file => console.log(file));
```
This will list all files Karma loads and all files will be listed twice. Removing `karma-coverage-istanbul-reporter` fixed the duplication, adding `karma-coverage` did not result in the same duplication as I saw before.

pndewit:
* gentle bump *
Any updates?

LarsDenBakker:
Rebased the PR, and fixed some linting errors. Sorry this one took so long to get through, we had some discussions within the team about whether we want to do a breaking change to the karma config at this point.

PR #1557, opened 2020-04-23
feat(testing-helpers): add scoped-elements support

manolakis:
This PR allows to use scopedElements in fixtures with the following syntax:

```js
await fixture('<test-class foo="bar"></test-class>', {
  scopedElements: {
    'test-class': TestClass,
  },
});
```

or

```js
await fixture(html`<test-class foo="bar"></test-class>`, {
  scopedElements: {
    'test-class': TestClass,
  },
});
```

### DIFF
--- a/packages/testing-helpers/src/fixture-no-side-effect.js
+++ b/packages/testing-helpers/src/fixture-no-side-effect.js
@@ -5,6 +5,8 @@ import { isValidRenderArg } from './lib.js';
 /**
  * @typedef {object} FixtureOptions
  * @property {Element} [parentNode] optional parent node to render the fixture's template to
+ * @property {import('@open-wc/scoped-elements').ScopedElementsMap} [scopedElements] optional scoped-elements
+ * definition map
  */
 
 /**
--- a/packages/testing-helpers/src/scopedElementsWrapper.js
+++ b/packages/testing-helpers/src/scopedElementsWrapper.js
@@ -1,6 +1,5 @@
 import { ScopedElementsMixin } from '@open-wc/scoped-elements';
 import { html, LitElement, TemplateResult } from 'lit-element';
-import { nothing } from 'lit-html';
 import { isIterable } from './lib.js';
 
 /** @typedef {import('@open-wc/scoped-elements').ScopedElementsMap} ScopedElementsMap */
@@ -31,13 +30,13 @@ class ScopedElementsTestWrapper extends ScopedElementsMixin(LitElement) {
     /** @type {ScopedElementsMap} */
     this.scopedElements = {};
 
-    /** @type {TemplateResult|{}} */
-    this.template = nothing;
+    /** @type {import('./litFixture').LitHTMLRenderable} */
+    // eslint-disable-next-line babel/no-unused-expressions
+    this.template;
   }
 
-  async firstUpdated() {
-    // @ts-ignore
-    await super.firstUpdated();
+  firstUpdated(_changed) {
+    super.firstUpdated(_changed);
 
     Object.keys(this.scopedElements).forEach(key =>
       this.defineScopedElement(key, this.scopedElements[key]),
@@ -49,7 +48,6 @@ class ScopedElementsTestWrapper extends ScopedElementsMixin(LitElement) {
   }
 }
 
-// @ts-ignore
 customElements.define('scoped-elements-test-wrapper', ScopedElementsTestWrapper);
 
 /**
--- a/packages/testing-helpers/src/scopedElementsWrapper.js
+++ b/packages/testing-helpers/src/scopedElementsWrapper.js
@@ -1,0 +1,69 @@
+import { ScopedElementsMixin } from '@open-wc/scoped-elements';
+import { html, LitElement, TemplateResult } from 'lit-element';
+import { nothing } from 'lit-html';
+import { isIterable } from './lib.js';
+
+/** @typedef {import('@open-wc/scoped-elements').ScopedElementsMap} ScopedElementsMap */
+
+const transform = template => {
+  if (isIterable(template)) {
+    return [...template].map(v => transform(v));
+  }
+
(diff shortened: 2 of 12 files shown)


packages/testing-helpers/src/fixture-no-side-effect.js
```diff
@@ -5,6 +5,7 @@ import { isValidRenderArg } from './lib.js';
 /**
  * @typedef {object} FixtureOptions
  * @property {Element} [parentNode] optional parent node to render the fixture's template to
+ * @property {Object<string, typeof HTMLElement>} [scopedElements] optional scoped-elements definition map
```
bennypowers, line 8: second type param will probably need to be a generic
something like 

```
@template {HTMLElement} T
@property {Object<string, T>}
```

But likely better expressed directly in TypeScript

packages/testing-helpers/src/scopedElementsWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  if (template instanceof TemplateResult) {
+    return html(template.strings, ...template.values);
+  }
+
+  return template;
+};
+
+class ScopedElementsClass extends ScopedElementsMixin(LitElement) {
```
bennypowers, line 18: small nit: since this is actually an element, maybe a name like `ScopedElementsHost`?

packages/testing-helpers/src/scopedElementsWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+
+  constructor() {
+    super();
+
+    this.scopedElements = {};
+    this.template = nothing;
+  }
+
+  async firstUpdated() {
+    // @ts-ignore
```
bennypowers, line 34: why?

packages/testing-helpers/src/scopedElementsWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+      this.defineScopedElement(key, this.scopedElements[key]),
+    );
+  }
+
+  render() {
+    return transform(this.template);
+  }
+}
+
+// @ts-ignore
```
bennypowers, line 47: why?

packages/testing-helpers/src/scopedElementsWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+customElements.define('scoped-elements-test-wrapper', ScopedElementsClass);
+
+/**
+ * Wraps the template inside a scopedElements component
+ *
+ * @param {import('./litFixture').LitHTMLRenderable} template
+ * @param {Object<string, typeof HTMLElement>} scopedElements
```
bennypowers, line 54: could use same exported type

packages/testing-helpers/src/scopedElementsWrapper.js
```diff
@@ -0,0 +1,64 @@
...
+ * @returns {import('lit-html').TemplateResult}
+ */
+export function getScopedElementsTemplate(template, scopedElements) {
```
bennypowers, line 57: does this need to be a function?

packages/testing-helpers/src/fixture-no-side-effect.js
manolakis: Honestly, I'm a noob regarding TypeScript 😅

First of all thanks for your comments @bennypowers 🤗

I'm not sure about that change because `typeof HTMLElement` refers to any class element that extends HTMLElement but using `@template {HTMLElement} T` I "think" that it refers to an instance of HTMLElement, not the class. Am I wrong?

packages/testing-helpers/src/scopedElementsWrapper.js
manolakis: It's because `ScopedElementsMixin` doesn't have `firstUpdated` method and the linter is not inferring LitElement methods.

Screenshot 2020-04-24 at 15 39 27

packages/testing-helpers/src/scopedElementsWrapper.js
manolakis: Same as before but in this case the is not inferring the `CustomElementConstructor`.

Screenshot 2020-04-24 at 15 43 19

packages/testing-helpers/src/scopedElementsWrapper.js
manolakis: As it is exported I'm using a function because `tsc` doesn't export the types in case using an arrow function.

packages/testing-helpers/src/scopedElementsWrapper.js
manolakis: I just changed to `ScopedElementsTestWrapper` which I think is more aligned with what is doing.

packages/testing-helpers/src/scopedElementsWrapper.js
manolakis: I changed it to use the type exported by `@open-wc/scoped-elements`

packages/testing-helpers/src/fixture-no-side-effect.js
manolakis: I created the types.d.ts file in `@open-wc/scoped-elements` to export the type
```typescript
export type ScopedElementMap = {
  [key: string]: typeof HTMLElement;
}
```

LarsDenBakker:
LGTM, @bennypowers you wanna check the review comments?

packages/testing-helpers/src/scopedElementsWrapper.js
bennypowers: we should fix the mixin definition then. will pm on slack

packages/testing-helpers/src/scopedElementsWrapper.js
bennypowers: see above

PR #1561, opened 2020-04-24
fix(semantic-dom-diff): allow assertion message

### LINKED ISSUE #1477: [semantic-dom-diff] Allow specifying assertion message
I'm used to being able to set the assertion message as the last argument to the assertion function:
```ts
assert.equal(a, b, 'a and b are not the same');
```
`assert.lightDom.equal` doesn't appear to accept a message argument.

bennypowers:
also fixes types

fixes #1477

### DIFF
--- a/packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
+++ b/packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
@@ -1,19 +1,37 @@
 /// <reference types="chai" />
 
-declare namespace Chai {
-  interface Assertion {
-    dom: Assertion;
-    lightDom: Assertion;
-    shadowDom: Assertion;
-    notEqual(actual: Object, expected: Object, message?: string): void;
-    equalSnapshot(options?: Object): Assertion;
-    notEqualSnapshot(options?: Object): Assertion;
-  }
+import { DiffOptions } from "./get-diffable-html";
 
-  interface Assert {
-    dom: Assertion;
-    lightDom: Assertion;
-    shadowDom: Assertion;
-    equalSnapshot(fixture: any, options?: Object): Assertion;
+declare global {
+  namespace Chai {
+    interface Assertion extends LanguageChains, NumericComparison, TypeComparison {
+      dom: Assertion;
+      lightDom: Assertion;
+      shadowDom: Assertion;
+      notEqual(actual: Object, expected: Object, message?: string): void;
+      equalSnapshot(options?: Object): Assertion;
+      notEqualSnapshot(options?: Object): Assertion;
+    }
+
+    interface Assert {
+      dom: Pick<Assertion, 'equal'|'notEqual'|'equalSnapshot'|'notEqualSnapshot'>;
+      lightDom: Pick<Assertion, 'equal'|'notEqual'|'equalSnapshot'|'notEqualSnapshot'>;
+      shadowDom: Pick<Assertion, 'equal'|'notEqual'|'equalSnapshot'|'notEqualSnapshot'>;
+      equalSnapshot(fixture: any, options?: DiffOptions): Assertion;
+      notEqualSnapshot(fixture: any, options?: DiffOptions): Assertion;
+
+      equal<T>(actual: T, expected: T, message?: string, options?: DiffOptions): void;
+      equal<T>(actual: T, expected: T, message?: string): void;
+      equal<T>(actual: T, expected: T, options?: DiffOptions): void;
+
+      notEqual<T>(actual: T, expected: T, message?: string, options?: DiffOptions): void;
+      notEqual<T>(actual: T, expected: T, message?: string): void;
+      notEqual<T>(actual: T, expected: T, options?: DiffOptions): void;
+    }
+
+    interface Equal {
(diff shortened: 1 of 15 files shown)


packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
```diff
@@ -1,19 +1,37 @@
...
+      dom: Assertion;
+      lightDom: Assertion;
+      shadowDom: Assertion;
+      notEqual(actual: Object, expected: Object, message?: string): void;
+      equalSnapshot(options?: Object): Assertion;
+      notEqualSnapshot(options?: Object): Assertion;
+    }
+
+    interface Assert {
+      dom: Pick<Assertion, 'equal'|'notEqual'|'equalSnapshot'|'notEqualSnapshot'>;
```
bennypowers, line 17: I think a "truer" approach would be to implement the entire interface in JavaScript, just throw errors on nonsense assertions. 

cc: @LarsDenBakker

packages/semantic-dom-diff/chai-dom-diff.d.ts
```diff
@@ -1,7 +1,24 @@
...
+/// <reference path="./chai-dom-diff-plugin.d.ts" />
 
 /**
- * @param {any} chai
- * @param {any} utils
+ * @param chai
+ * @param utils
  */
-export const chaiDomDiff: (chai: any, utils: any) => void;
+export const chaiDomDiff: (chai: typeof Chai, utils: any) => void;
```
bennypowers, line 7: should maybe be a type assertion of ChaiPlugin

packages/semantic-dom-diff/chai-dom-diff.d.ts
```diff
@@ -1,7 +1,24 @@
...
+
+// No idea what this is
+type Snapshot = any;
```
bennypowers, line 10: seriously, what is this?

packages/semantic-dom-diff/chai-dom-diff.js
```diff
@@ -37,21 +57,31 @@ export const chaiDomDiff = (chai, utils) => {
...
     } else {
-      assertion.equal(expectedDiffableHTML);
+      assertion.equal(expectedDiffableHTML, message);
     }
   };
 
   /** DOM assertion for `should` and `expect` interfaces. */
   const domEquals = _super =>
+    /**
+     * @this {any}
```
bennypowers, line 83: what is `this` here?

packages/semantic-dom-diff/chai-dom-diff.js
```diff
@@ -80,8 +110,15 @@ export const chaiDomDiff = (chai, utils) => {
   const context = window.__mocha_context__;
   const snapshotState = window.__snapshot__;
 
-  /** Base HTML snapshot assertion for `assert` interface. */
-  function assertHtmlEqualsSnapshot(actual, options, negate = false) {
+  /**
+   * Base HTML snapshot assertion for `assert` interface.
+   * @this {any}
```
bennypowers, line 115: What is `this` here?

packages/semantic-dom-diff/chai-dom-diff.js
```diff
@@ -119,7 +156,10 @@ export const chaiDomDiff = (chai, utils) => {
     }
   }
 
-  /** Snapshot assertion for `should` and `expect` interfaces. */
+  /**
+   * Snapshot assertion for `should` and `expect` interfaces.
+   * @this {any}
```
bennypowers, line 161: what is `this` here?

bennypowers:
Notes to reviewers: I'm pretty sure I didn't break any public apis here, hence the patch version, but please double-check

packages/semantic-dom-diff/chai-dom-diff.js
daKmoR: chai itself... chai works by binding its plugin function to itself... that's also the reasons that `handleDom` can not be an arrow function... as it would break chains plugin "system"...

it's a very old school system... because it's old like really old 😅

bennypowers: is that Chai or Chai.ChaiStatic?

bennypowers: I think this instance is actually `Chai.AssertionStatic`

packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
LarsDenBakker: I didn't write this code, so I'm not sure :p

packages/semantic-dom-diff/chai-dom-diff.d.ts
LarsDenBakker: this comes from https://www.npmjs.com/package/mocha-snapshots and https://www.npmjs.com/package/karma-mocha-snapshot

packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
bennypowers: YOLO

Issue #1484, opened 2020-04-02
Extract "rel=stylesheet" data from HTML files


Westbrook:
I'd like for for `rollup-plugin-html` to `extractStylesheets` as well as `extractModules`.

I'm not sure if wanting to put the page CSS in the HTML is a byproduct of my stubbornness to support old practices, the desire to force most (hopefully all) `import styles from './styles.css'` to support LitElement (though `rollup-plugin-postcss` also doesn't seem to support the `include`/`exclude` options common to Rollup that I think I'd want to start with to be able to work out conditional processing there) or something else that I'm missing, but I think I want to put `<link rel="stylesheet">` tags in my HTML and have it extract them both for possible `postCSS` processing and to ensure the build takes any files referenced and puts them into dist automatically.

If this made sense to you too, I'd be happy to start putting together a PR with a little guidance on how you thought that could best be added to the plugin.

LarsDenBakker:
I agree this would be good to add to the HTML plugin. I think it's good to start with some use cases we have in mind here, for example some people may want to concatenate multiple stylesheet links into one file while others may want to just inline the stylesheet into the HTML.

Westbrook:
Good call! Do you keep a map of how we're managing various `<script>` tag realities? I think this would probably line up 1 to 1 with that, but not sure. Some possibilities off the top of my head:

**A single `<link rel="stylesheet">`**
In the context of a "componentized" front-end, this would seem to be the default as an application's individual components would manage their own styles, and the page level styles would be more about load management and static HTML orchestration.
- do nothing beyond ensuring the file is taken along for the ride (would this automatically put it into the rest of the Rollup pipeline?). if so, this might nullify some of the following options, as we could just allow the user to manage a CSS file themselves.
  - probably opt-out
- surface a `transform` relative option so that custom processing can occur
  - does this nullify the `minify` option?
- surface a `minify` relative option
- surface an `inline` option to inject the CSS into a `<style>` tag

**Multiple `<link rel="stylesheet">` tags**
What are the goals of a user that does this? Even a user that doesn't rely on SASS could leverage `@import` in raw CSS to get this same functionality, right?
- options like the above
- surface a `bundle` option to merge these files
  - nullified by the `inline` option?

Other concepts that might deserve a little thought:
- Above the fold CSS injection
  - I've never used tools for this, and am unsure what sort of guardrails would be needed for a tool like ours to support aligning with a tool like that
- Unused CSS clean up
[...]

PR #1562, opened 2020-04-24
docs: move docs site to 11ty

### LINKED ISSUE #1421: Exchange VuePress for a static site generator and web components ⁉️
I see you're using VuePress to ship your documentation site... 😏 What a strange decision for a set of recommendations around web components. You should probably change that!

What if you used a static site generator (I [hear](https://github.com/Polymer/lit-html/pull/1076) [11ty](https://11ty.dev) is cool), and then leverage actual web components to deliver your content? 🤔 

Let's chat about it here to make plans. I know that people would be interested in taking part in the process...as well as seeing the success in this area. 🎉

Westbrook:
Copy to site almost verbatim from Vuepress to 11ty.

fixes #1421


thepassle:
I took a look at this, and I'm just writing down some thoughts/todos/bears on the road that I see. Pretty nice how easily you were able to reuse most of the layout and styling! That already helps a lot I think. Here's a list:

- layout for landing page
- layout for doc pages
- include prism styling for code snippets in the docs
- toc/menu bar (left)
  - this can differ depending on which kind of page you are on (for example: guide/faq/about have different menu items)
- the hamburger menu button for mobile view is triggered with JS, we should make a solution for that
- search functionality (how do we index?  11ty docs point to this for reference: https://www.hawksworx.com/blog/adding-search-to-a-jamstack-site/)
- Vuepress uses... Vue, and does spa routing, how should we approach this? Just go full MPA im assuming?
- Vue components in our docs should be turned into web components, and then figure out how to include those in 11ty
  - I guess this is where we can use the rollup-html-plugin? How would something like that look like, run 11ty first, and then run rollup against it? How do we get the html files that are output by 11ty in the rollup config? I dont really have a good vision/overview on this yet
- Figure out how to include the codelabs
- Service worker integration
- How do we achieve something like this?
image
- I'm assuming we'll have to implement these ourself as well:
image
- opening/closing of the menu items requires some scripting as well (when you open one, the other opened item closes)
menu

All in all I think it looks doable, but I think it will take quite some work, and, unless im missing something, we'll have to roll quite some solutions ourself (or consider going with a premade template? But most of the 11ty themes I found don't look like documentation sites, mostly blogs). 

I'm also a bit unsure how extensible this approach would be, as an example: we currently have the vuepress config file that makes it easy to shuffle content around etc, im unsure what this would look like when we go with 11ty.

LarsDenBakker:
We could also reconsider some of the UI elements, for example I'm not really sure the backward/forward navigation really adds anything. And has anyone ever used the edit on github link?

Westbrook:
@LarsDenBakker I've been leveraging this pipeline: https://github.com/adobe/spectrum-web-components/blob/westbrook/11ty/projects/documentation/package.json#L11 which would of course be majorly simplified by the work we're discussing in https://github.com/open-wc/open-wc/issues/1484

It's a little extra complicated with the TS and PostCSS processes and some lack of experience with the Rollup copy plugin.

thepassle:
> We could also reconsider some of the UI elements

I agree, im not sure if we really need them, and we can always iterate on it and add it later.

> I could dig into the 11ty -> rollup situation if noone has thought of a solution here yet.

Sounds good

I've been looking into 11ty, it seems fairly easy, so I can help westbrook with the implementation of 11ty

docs/faq/README.md
```diff
@@ -1,3 +1,7 @@
+---
+permalink: 'faq/index.html'
```
Westbrook, line 2: This is the main thing that I think will really separate this from the Vuepress approach. I've not seen any documentation that would allow for a similar centralized sitemap, so each page that we don't want to resolve to the default (e.g. this file would be `/faq/README/index.html` normally) we would need to ensure had a `permalink` value set. Lots of growth opportunities from there, however. Things like custom page meta data, dynamic navigation creation by collection tagging, etc.

Westbrook:
https://5ea44e899593300006622960--open-wc-org.netlify.app/index.html

Now with:
- home page
- dynamic side nav (desktop only for this pass)
- syntax highlighting
- all pages static so no SPA
- all pages
- codelabs seem too work...as if magic? (I literally did no things)

Thought for tomorrow:
- pair this version back if we really believe that we'll be working on a new version soon?
  - no edit links
  - no next previous
  - no search (though https://github.com/sw-yx/netlify-plugin-search-index is an interesting option there)
- this is primarily a brute force conversion, in line with the above and the dynamic navigation that Vuepress was making I'd suggest we don't worry too much about making it more dynamic until we decide on a path forward for the IA/content strategy.

bennypowers:
>  And has anyone ever used the edit on github link?
I've very often wished there was a link to source from docs pages. not to the repo readme, to package src

Westbrook:
86 ❌ search 
86 ❌ the `::: tip` tagging
86 ❌ the `script` tags
make the nav work in mobile 📵 
gather any unknown blockers from the team 🚫 

prep to split handbook/tools 👨‍🔬

thepassle:
I just remembered we should also have google analytics

netlify.toml
```diff
@@ -0,0 +1,7 @@
+[[plugins]]
+package = "netlify-plugin-checklinks"
+
+    [plugins.inputs]
+    pretty = true
+
+    skipPatterns = ['open-wc-org.netlify.app']
```
Westbrook, line 7: Links directly to the new preview build will inherently fail because it's not been published yet, so skip them.

bennypowers:
> > And has anyone ever used the edit on github link?
> 
> I've very often wished there was a link to source from docs pages. not to the repo readme, to package src

I think we could do this easily enough if we made it a point to link the package.json `bugs` field to the README for each package?

package.json
```diff
@@ -30,14 +28,31 @@
...
-    "vuepress:copy-static": "cp -f packages/demoing-storybook/demo/custom-elements.json docs/.vuepress/public/demoing/demo/custom-elements.json",
-    "vuepress:start": "vuepress dev docs"
+    "vuepress:build": "cd docs && eleventy",
+    "vuepress:serve": "es-dev-server --root-dir _site",
+    "vuepress:start": "run-p vuepress:watch vuepress:serve",
+    "vuepress:watch": "cd docs && eleventy --watch"
   },
   "dependencies": {
-    "@github/clipboard-copy-element": "^1.1.1"
+    "@fec/eleventy-plugin-remark": "^1.0.2",
```
Westbrook, line 37: The linter needed some of these in `dependencies` and some in `devDependencies` and I'm not fully sure why or if it makes any sense. A later revision where we move the "docs" into a private package would likely benefit some cleanliness here.

Westbrook: This move would also play well with the `@open-wc/11ty` style package that we're leaning towards here...

netlify.toml
```diff
@@ -0,0 +1,7 @@
+[[plugins]]
```
Westbrook, line 1: We might actually want to turn this off now that we've "completed" the cutover as it really blows our build budget at Netlify by about extra 2 or 3 minutes... or maybe we reach out to them and see if they could hook us up with more time for helping them beta test this...

package.json
bennypowers: just don't forget to move them over after wards. cite ☝️ this comment in the issue

netlify.toml
bennypowers: better if we can mooch more time

we can suffice with google search console crawl errors, but that will only be an after-the-fact report, and i'm not sure about notifications

maybe we can run this only on master, or only sporadically?

package.json
Westbrook: Added #1588

LarsDenBakker:
Green light from me! I say let's get this moving, and fix any issues we encounter.

daKmoR:
superseded by https://github.com/open-wc/open-wc/pull/1594

PR #1563, opened 2020-04-25
feat(es-dev-server): add https redirect

ruphin:
Adds an HTTPS redirect when using es-dev-server in https mode.

### DIFF
--- a/packages/es-dev-server/src/create-server.js
+++ b/packages/es-dev-server/src/create-server.js
@@ -23,7 +23,7 @@ function httpsRedirect(req, res) {
  *
  * @param {import('./config').InternalConfig} cfg the server configuration
  * @param {import('chokidar').FSWatcher} fileWatcher
- * @returns {{ app: import('koa'), server: import('http').Server | import('http2').Http2SecureServer }}
+ * @returns {{ app: import('koa'), server: net.Server | httpServer.Server }}
  */
 export function createServer(cfg, fileWatcher = chokidar.watch([])) {
   const middlewares = createMiddlewares(cfg, fileWatcher);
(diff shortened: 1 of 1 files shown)


packages/es-dev-server/src/create-server.js
```diff
@@ -36,7 +47,46 @@ export function createServer(cfg, fileWatcher = chokidar.watch([])) {
       allowHTTP1: true,
```
LarsDenBakker, line 47: I was expecting this one to have to be false to allow connecting to the redirect server

LarsDenBakker:
LGTM, there are some linting errors

packages/es-dev-server/src/create-server.js
ruphin: The application server can keep accepting HTTP1 and HTTP2 connections like it was before.

The new TCP server accepts anything and just forwards requests to either the redirect server or the application server, the protocols that they accept have no impact on this.

ruphin:
I am unable to run linting or tests in this project so I cannot fix any problems. Feel free to add commits to this MR, or just take the changes and apply them locally, and drop this MR.

LarsDenBakker:
It should be a matter of `yarn install` and `yarn format` from the root of the project. Is that not working for you?

ruphin:
`npm install` breaks for me at some point as it tries to run lerna things
`npm run format` breaks as it tries to load `eslint-config-prettier` which doesn't exist

I think it's easier for you to just run the linter and commit whatever changes you need.

LarsDenBakker:
That's because we use yarn workspaces, so you need to use `yarn install` and `yarn format` :)

I can also do it when I get the time.

ruphin:
I don't have yarn, best if you do it :)

ruphin:
Sweet, thanks!

bennypowers:
np
try `npx yarn format` next time

PR #1565, opened 2020-04-25
chore(docs): add eslint-config-prettier to linting docs

jorenbroekema:
I was upgrading my prettier to 2.0 and noticed there are some caveats with eslint that have been discussed on Slack and also seem to be integrated in the generators, but are not yet documented.

### DIFF
--- a/packages/eslint-config/README.md
+++ b/packages/eslint-config/README.md
@@ -55,6 +55,40 @@ Run:
 - `npm run lint:eslint` to check if any file is not correctly formatted
 - `npm run format:eslint` to auto format your files
 
+## Running with Prettier
+
+If you also are using Prettier, in order for your ESLint to play nicely with Prettier, you should also include eslint-config-prettier.
+This way, the Prettier formatting will take into account your linting rules, so you don't need to add your own overrides for Prettier if it clashes with your linting rules.
+
+```sh
+npm add --save-dev eslint-config-prettier
+```
+
+```json
+{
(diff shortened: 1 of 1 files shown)


packages/eslint-config/README.md
```diff
@@ -55,6 +55,43 @@ Run:
 - `npm run lint:eslint` to check if any file is not correctly formatted
 - `npm run format:eslint` to auto format your files
 
+## Running with Prettier
+
+If you also are using Prettier, in order for your ESLint to play nicely with Prettier, you should also include eslint-config-prettier.
+This way, the Prettier formatting will take into account your linting rules, so you don't need to add your own overrides for Prettier if it clashes with your linting rules.
+
```
jorenbroekema, line 62: I would like to add here:

```
If you add Prettier overrides that clash with your linting configuration, the linting will take precedence and the Prettier override will not apply. So take this into account as well.
```
But I'm not sure if that's actually correct, so if someone could verify :) I will add it

LarsDenBakker:
I wonder if we should just include the prettier linting configuration by default. It works fine even when you're not using prettier I think?

jorenbroekema:
@LarsDenBakker  I think it's added like this in the generator, this is more if you're manually updating your repo to Prettier 2. Or do you mean something else?

LarsDenBakker:
I meant to include the plugin by default in our linting config. But we can pick that up later.

PR #1567, opened 2020-04-26
feat(building-rollup): export appendScriptToBody function

abdonrd:
I think this is useful when we want to use the `transform` option for the `html` feature.
For example to custom service worker inject:

```javascript
import { appendScriptToBody } from '@open-wc/building-rollup/src/utils';
// ...

const baseConfig = createSpaConfig({
  outputDir: DIST_PATH,
  legacyBuild: true,
  html: {
    transform: (html) => {
      const code = `
        if ('serviceWorker' in navigator) {
          window.addEventListener('load', function () {
            navigator.serviceWorker
              .register('./service-worker.js')
              .then(function () {
                console.log('Service Worker registered.');
              })
              .catch(function (err) {
                console.log('Service Worker registration failed: ', err);
              });
          });
        }
      `;
      return appendScriptToBody(html, code);
    }
  },
  workbox: workboxConfig
});
```

Also remove unneeded `Terser.minify`.

### DIFF
--- a/packages/building-rollup/src/utils.js
+++ b/packages/building-rollup/src/utils.js
@@ -39,14 +39,24 @@ function pluginWithOptions(plugin, userConfig, defaultConfig, ...otherParams) {
 
 /**
  * @param {string} htmlString
+ * @param {string} code
+ * @returns {string}
+ */
+function appendScriptToBody(htmlString, code) {
+  const documentAst = parse(htmlString);
+  const body = query(documentAst, predicates.hasTagName('body'));
+  const swRegistration = createScript({}, code);
+
+  append(body, swRegistration);
+  return serialize(documentAst);
+}
+
+/**
+ * @param {string} htmlString
  * @returns {string}
  */
 function applyServiceWorkerRegistration(htmlString) {
-  const documentAst = parse(htmlString);
-  const body = query(documentAst, predicates.hasTagName('body'));
-  const swRegistration = createScript(
-    {},
-    Terser.minify(`
+  const { code } = Terser.minify(`
     if ('serviceWorker' in navigator) {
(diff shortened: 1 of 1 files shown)


thepassle:
Looks like there is a failing test:

```
lerna ERR! yarn run test:node exited 1 in '@open-wc/building-rollup'
lerna ERR! yarn run test:node stdout:
yarn run v1.16.0
$ mocha test/**/*.test.js test/*.test.js


  integration tests
    testcase js/rollup.spa.config.js
      ✓ passes the in-browser tests
      ✓ outputs a service worker
    testcase js/rollup.spa-nomodule.config.js
      1) "before all" hook for "passes the in-browser tests"


  2 passing (19s)
  1 failing

  1) integration tests
       testcase js/rollup.spa-nomodule.config.js
         "before all" hook for "passes the in-browser tests":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/repo/packages/building-rollup/test/integration.test.js)
  

```

Otherwise, this looks fine to me, but I remember in the past we had some discussion about depending/publishing code from utils, cc @LarsDenBakker 

Also, would it be easier to just let a user specify their own service worker filename instead, and pass that to the applyServiceWorkerRegistration instead?

abdonrd:
@thepassle I can't reproduce the test error in local. 😕

abdonrd:
And the function could work for add other scripts like analytics?

LarsDenBakker:
I don't think we should expose this from building-rollup. This should be in some kind of dom-5-utils pacakge, but personally I always just do a string replace for these types of things.

abdonrd:
Okey! We can close it then!

PR #1569, opened 2020-04-26
chore: ignore build types files

daKmoR:
What I did:

- generally, ignore `*.d.ts` files (as they get auto generated)
- keep handwritten `*.d.ts` files by adding them to the gitignore ignore 😅
- auto-generate types for lit-helpers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,34 @@ local.log
 
 ## generated codelabs
 docs/.vuepress/public/codelabs
+
+## typings - add manually written *.d.ts files here
+*.d.ts
+!packages/es-dev-server/src/middleware/polyfills-loader-types.d.ts
+!packages/es-dev-server/src/middleware/compatibility-transform-types.d.ts
+!packages/es-dev-server/src/utils/inject-polyfills-loader-types.d.ts
+!packages/mdjs/types/global.d.ts
+!packages/mdjs/index.d.ts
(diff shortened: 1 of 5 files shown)


manolakis:
LGTM :)

daKmoR:
will change to only ignore `*.d.ts` files for "folders/packages" that we actually build types for

bennypowers:
i'd still prefer to write src in ts, 

but perhaps a less fiddly solution would be to require all custom types to live in `@open-wc/types` and have that also be a typeroot

LarsDenBakker:
What does this change in comparison to the other PR? We still need to whitelist all the manually written declaration files?

daKmoR:
you would only need to whitelist for packages if you also generate definition files for it

LarsDenBakker:
Which is all packages :)

LarsDenBakker:
To clarify, for a lot of packages I generate .d.ts from jsdocs but additionally hand write some .d.ts files to express complex types. It's really annoying to write object types in jsdoc. 

Example: https://github.com/open-wc/open-wc/blob/master/packages/polyfills-loader/src/types.d.ts
Imported in: https://github.com/open-wc/open-wc/blob/master/packages/polyfills-loader/src/inject-polyfills-loader.js

So for me personally, this change is annoying and I will forget to include some declaration files. I can live with it if it brings a significant improvement for others though.

PR #1573, opened 2020-04-27
feat(resolve): update to new proposal specs

LarsDenBakker:
This updates our resolve package to the latest specs. It's mostly a copy of the reference code with added types.

I ported the tests to mocha, but didn't change anything to the way the tests are structured even though they're a bit different to what we usually do.


packages/import-maps-resolve/package.json
```diff
@@ -13,28 +13,17 @@
   },
   "author": "open-wc",
   "homepage": "https://github.com/open-wc/open-wc/tree/master/packages/import-maps-resolve",
-  "main": "dist/index.js",
+  "main": "./src/index.js",
```
daKmoR, line 16:
```suggestion
  "main": "./index.js",
```

PR #1578, opened 2020-04-29
Run worbox only on the last output build

abdonrd:
After all, the legacy build browsers doesn’t support SWs.

### DIFF
--- a/packages/building-rollup/src/createSpaConfig.js
+++ b/packages/building-rollup/src/createSpaConfig.js
@@ -38,6 +38,30 @@ function createSpaConfig(options) {
     minify: !userOptions.developmentMode,
   };
 
+  const workboxPlugin = pluginWithOptions(
+    generateSW,
+    userOptions.workbox,
+    {
+      globIgnores: ['polyfills/*.js', 'legacy-*.js', 'nomodule-*.js'],
+      navigateFallback: '/index.html',
+      // where to output the generated sw
+      swDest: path.join(process.cwd(), outputDir, 'sw.js'),
+      // directory to match patterns against to be precached
+      globDirectory: path.join(process.cwd(), outputDir),
+      // cache any html js and css by default
+      globPatterns: ['**/*.{html,js,css,webmanifest}'],
+      skipWaiting: true,
+      clientsClaim: true,
+      runtimeCaching: [
+        {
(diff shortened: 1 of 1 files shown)


packages/building-rollup/src/createSpaConfig.js
```diff
@@ -72,30 +105,7 @@ function createSpaConfig(options) {
...
-            runtimeCaching: [
-              {
-                urlPattern: 'polyfills/*.js',
-                handler: 'CacheFirst',
-              },
-            ],
-          },
-          () => {},
-        ),
+      userOptions.workbox && workboxPlugin,
```
thepassle, line 108: This should not be needed, right? Because you already push the workboxPlugin to the plugins array on line 95?

abdonrd: Ops! I forget to remove, yes!

abdonrd:
And maybe we need a `buildEnd` rollup hook?

thepassle:
> And maybe we need a `buildEnd` rollup hook?

Why?

abdonrd:
I'm asking because I'm not sure if the worbox plugin can run before the `index.html` file is generated.

thepassle:
> I'm asking because I'm not sure if the worbox plugin can run before the `index.html` file is generated.

Does the index.html not get cached with this configuration?

abdonrd:
> > I'm asking because I'm not sure if the worbox plugin can run before the `index.html` file is generated.
> 
> Does the index.html not get cached with this configuration?

Never! haha 😬

Tansito:
Event if you change this [line](https://github.com/open-wc/open-wc/pull/1578/files#diff-8d43bac01c3961a06609fb0eff67e01dR93) to:
```
basicConfig.output[01].plugins.push(workboxPlugin);
```
It works sometimes but others you can finish having a service worker with this configuration:
Screen Shot 2020-04-29 at 1 04 45 PM
With no precacheAndRoute files configuration, everything empty (not just without the `index.html` only).

abdonrd:
@thepassle It works with the current @open-wc starter app!

thepassle:
Thanks for sticking with me here 😄That was a rough one to figure out

abdonrd:
@thepassle @Tansito and thanks to you two too! 👏

PR #1579, opened 2020-04-29
fix(core): use types everywhere, export interfaces

bennypowers:


bennypowers:
@dakmor check out the interface in mdjsStoryParse.js. not sure if that's an exhaustive list of properties that I derived there, and some of the unions might be unnecessarily narrow.

LarsDenBakker:
Looks good to me

LarsDenBakker:
Actually, can we add a `tsconfig` here like this: https://github.com/open-wc/open-wc/blob/master/packages/polyfills-loader/tsconfig.json so that we don't regress on types?

LarsDenBakker:
@daKmoR you wanna check this?

daKmoR:
awesome - thx 🤗

Issue #1580, opened 2020-04-30
[testing][testing-helpers] Breaking changes recently introduced: lit-element 


100stacks:
Hi there! 👋😷 

I really appreciate all the wonderful work you all are doing!  After a recent update, I noticed my Open-WC testing framework is not building.  

## `lit-element` peer dependency was introduced:

https://github.com/open-wc/open-wc/blob/dfe74f307dce5da832ab3ba266a85ddd314792b8/packages/testing-helpers/package.json#L32

```
warning "@open-wc/testing > @open-wc/testing-helpers@1.8.1" has unmet peer dependency "lit-element@^2.2.1".
```
~~Please consider making it a dev dependency so that we don't have to install it at our project root.~~ Just saw this...Issue #175.   

The beauty of using `@open-wc/testing` is that it provides us a web component testing framework that handles all the package dependencies for us! 👏🙌 In this case, since `@open-wc/testing` package.json is pinned to minor and patch level updates, this allowed the breaking change to be introduced.

## Please consider...

Please consider bumping `@open-wc/testing` to a major version release when breaking changes are introduced in one of the package dependencies.  If you do decide to introduce a major release, please consider patching the previous v2.5.x to **not** require the new `lit-element` peer dependency, which looks like it wouldn't include scoped elements support.  And bump `@open-wc/testing-helpers` to a 2.x release with the above update, so current direct consumers of this package don't get caught by this change.

## Also possibly consider...

Please consider forking `@open-wc/testing` to say ~~`@open-wc/testing-lit-elements`~~ (this is probably default), or `@open-wc/testing-agnostic` to allow consumers the choice between using the LitElement framework or not for testing web components.

I think LitElement provides a very useful framework for web components.  It's just that our team has already committed to using the Web Platform as much as possible, and we are trying to limit external dependencies.

Thank you for considering, and again for all the awesome hard work!!! 💪🎉

LarsDenBakker:
It's a real shame we now have a peer dependency on lit-element. It's because of some other important work, but we will think about how we can support this without a dependency on lit-element.

100stacks:
Hi @LarsDenBakker...I can definitely appreciate the important work that it supports. 👍

I'll work around this new requirement until you're able to support another solution.  Thanks!

LarsDenBakker:
We opted for adding it as a real dependency. This does increase the amount of dependencies installed, but only for testing. Lit-element is very small compared to the enormous amount of dev dependencies installed for a typical project.

[...]

PR #1587, opened 2020-04-30
fix(testing-helpers): dependencies & await to the wrong element update

manolakis:
In case scopedElements litFixture was awaiting element update of `node.firstElementChild`, which was `null`, instead of `node`.


bennypowers:
Fixes #1580

Issue #1361, opened 2020-02-22
[es-dev-server] recommend tsc over babel


LarsDenBakker:
The babel typescript plugin seemed like a great idea, but there are quite some areas where it falls short. Most importantly there are issues with file extensions, decorators and class fields. 

For `es-dev-server` we should flip around the recommendation to use `tsc --watch` while developing. Although it scaffolds lots of files, it is a better overall dev experience. Separately we could investigate if using https://github.com/rollup/plugins/tree/master/packages/typescript inside es-dev-server (as a plugin) improves the dev experience.

vikerman:
I am trying to recreate the TypeScript setup manually here with `tsc --watch` integrated into different npm scripts - https://github.com/vikerman/open-wc-typescript-starter/blob/master/tsconfig.json. 

It's low tech but it's better than using rollup in es-dev-server because the turn around time I suspect would be much better since rollup is not involved and stays closer to rest of the JS-only tooling and less maintenance that way.

PR #1591, opened 2020-05-02
Docs/various

LarsDenBakker:
Improves the documentation for some packages. It's all readmes, so it should not interfere with the 11ty migration.

Fixes https://github.com/open-wc/open-wc/issues/1361

### DIFF
--- a/packages/es-dev-server/README.md
+++ b/packages/es-dev-server/README.md
@@ -1,6 +1,6 @@
-# ES dev server
+# es dev server
 
-A web server for development without bundling, utilizing the browser's standard module loader and efficient browser caching for simple and fast web development.
+A web server for development without bundling.
 
 ```bash
 npx es-dev-server --node-resolve --watch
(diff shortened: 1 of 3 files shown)


packages/es-dev-server/README.md
```diff
@@ -218,20 +137,22 @@ If you run the `es-dev-server` command from the root of the project, you can acc
...
+Use the `--open` parameter for when you'd like to keep you index.html in a subfolder.
 
 ```
 node_modules/...
 src/...
 src/index.html
 ```
 
-You can access your app in the browser at `/src/` or `/src/index.html`. You can tell `es-dev-server` to explicitly open at this path:
+You can access your app in the browser at `/src/` or `/src/index.html`. You can tell EDS to explicitly open at this path:
```
daKmoR, line 155: hmmm I'm not sure if we want to start introducing acronyms in documentations? `EDS` is good for internal discussions but can be confusing if it's a first time reader who potentially only skims over the docs to find a specific option

LarsDenBakker: Fair enough, changed it back.

Issue #1213, opened 2020-01-11
Drop support for node v8


LarsDenBakker:
Our tools are updating / want to update and drop support for node v8. Do we see any problems with dropping support?

daKmoR:
v8 had End of support on 2019-12-31
https://en.wikipedia.org/wiki/Node.js#Releases

so imho moving off v8 is a good thing

coreyfarrell:
I'm proceeding, the minimum supported version for the next release of my modules will be node.js 10.13.0 (first LTS).

LarsDenBakker:
Thanks for the update, we will go for v10 too.

PR #1592, opened 2020-05-02
chore: require node v10 or higher

LarsDenBakker:
We implicitly already require node v10 for a lot of packages, this makes that dependency explicit.

Node v8 is no longer supported by node itself, so I think this should not be a breaking change.

What do you guys think?

Fixes https://github.com/open-wc/open-wc/issues/1213

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "@open-wc/root",
   "private": true,
   "license": "MIT",
+  "engines": {
+    "node": ">=10.13.0"
+  },
   "scripts": {
     "build": "lerna run build",
     "build:types": "tsc -p tsconfig.build.types.json",
--- a/packages/building-rollup/package.json
+++ b/packages/building-rollup/package.json
@@ -14,6 +14,9 @@
   "author": "open-wc",
   "homepage": "https://github.com/open-wc/open-wc/tree/master/packages/building-rollup",
   "main": "index.js",
+  "engines": {
+    "node": ">=10.13.0"
+  },
   "scripts": {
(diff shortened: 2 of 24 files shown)


abdonrd:
Why explicit 10.13 instead of 10.0?

Westbrook:
This could potentially be a breaking change for users... I like the idea of it being just a chore, but almost every other project I use that has made this change did it in a breaking version, even if this was the only change in the release.

LarsDenBakker:
@abdonrd no reason, I took this from one of our dependencies. I can just do regular 10 though

@Westbrook implicitly we already don't support node 8 anymore since we're not compiling for it and use features that aren't supported. This is a concern though, I'd rather not release a breaking for everything :(

bennypowers:
do it!

ZempTime:
Suggestion - are there any breaking changes shortly on the horizon this could be grouped with in a release? Seems like a non-urgent but important change, so being able to group it with something else would be convenient & tasteful.

If not, still support moving off due to node no longer maintaining v8 👍 (we're not using it in any of our deps, and haven't been for a while). I see the point about, this is the case already. Maybe this is more a project values kind of question: can either be cautious & unobtrusive, or potentially obtrusive & convenient.

LarsDenBakker:
For some packages there might be breaking changes, for others definitely not. Doing a breaking for all of them is not an option in my opinion. Then I'll rather let things be as they are, and just add this field when we do a breaking (if we remember about this).

LarsDenBakker:
I'll close this for now, we probably need to do this on a per-package basis.

PR #1593, opened 2020-05-02
feat(create): support ts in generator

thepassle:
# How to review:
There's many code duplication because of the way the generator is set up, so it's probably easiest to checkout the branch locally and:

```bash
cd packages/create
npm start
```

and check the output, run npm scripts (linting/building/storybook/testing)

# Todo:

- [x] Storybook
- [x] Fix eslint-config-prettier thing
- [x] Vigorously test
  - [x] WC
    - [x] linting
    - [x] testing
    - [x] storybook
  
  - [x] APP
    - [x] linting
    - [x] testing
    - [x] storybook
    - [x] building

# Notes:

There seems to be a problem with `@types/node` on the latest version https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44019. A fix seems to be to install `npm install @types/node@13.11.1 --save-dev`. But its kind of silly to have to include that in the scaffolded package.json


packages/create/src/generators/wc-lit-element-ts/templates/_tsconfig.json
```diff
@@ -0,0 +1,15 @@
+  {
+  "compilerOptions": {
+    "target": "esnext",
```
LarsDenBakker, line 3: We should pick a target here, `es2018` or `es2019`, need to see what works on latest browsers.

packages/create/src/generators/wc-lit-element-ts/templates/_tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmitOnError": true,
+    "lib": ["es2017", "dom"],
+    "strict": true,
+    "esModuleInterop": true,
```
LarsDenBakker, line 9: This should be false, it's for commonjs modules but we don't support that anyway.

packages/create/src/generators/wc-lit-element-ts/templates/_tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "outDir": "out-tsc"
```
LarsDenBakker, line 12: For the web component this should be `dist` but I guess that creates a problem with the rollup build?

packages/create/src/generators/app-lit-element-ts/templates/_my-app.stories.ts
```diff
@@ -0,0 +1,11 @@
+import { html } from 'lit-html';
+import '../out-tsc/src/<%= tagName %>.js';
```
LarsDenBakker, line 2: Shouldn't this reference the source file?

packages/create/src/generators/demoing-storybook-ts/templates/static-scaffold/stories/index.stories.md
```diff
@@ -0,0 +1,48 @@
+```js script
+import { html } from '@open-wc/demoing-storybook';
+import '../out-tsc/<%= tagName %>.js';
```
LarsDenBakker, line 3: Shouldn't this point to the source file?

LarsDenBakker:
Great job! Quite a big task :) Some general remarks:

When I run the element generator, I still get a `index.js` file.

I'm not a fan of the `tsc-out` name, but I can't really come up with a better name right now.

File extensions should be `.js`, not `.ts`. Even in the source files. That's how TS works 🤷 

I also get the type error after installing. Would be nice if we produce a working project. Is this error only with certain versions of node?

Should we add `out-tsc` to gitignore? Then it won't get formatted/linted.

packages/create/src/generators/demoing-storybook-ts/templates/static-scaffold/stories/index.stories.md
thepassle: For the app generator storybook seems to work fine if I link to `../tagName.js` here, but in the component generator I'm getting a missing import for `../tagName.js` in the browser console, not sure why it works for one and not for the other 😕

thepassle:
> File extensions should be .js, not .ts. Even in the source files. That's how TS works 🤷

fixed

> I also get the type error after installing. Would be nice if we produce a working project. Is this error only with certain versions of node?

it seems to happen on both node 10 and node 13 for me

> Should we add out-tsc to gitignore? Then it won't get formatted/linted.

fixed

thepassle:
PTAL

packages/create/src/generators/demoing-storybook-ts/templates/static-scaffold/stories/index.stories.md
LarsDenBakker: I didn't realize this is in the .md file, then it should be JS.

LarsDenBakker:
Looks good, but what about the `out-tsc` for the components? Then it should be dist, right? And those files should be in the list of files to publish to NPM.

packages/create/src/generators/demoing-storybook-ts/templates/static-scaffold/stories/index.stories.md
thepassle: so `import '../out-tsc/<%= tagName %>.js';` is correct here, right?

thepassle:
> Looks good, but what about the out-tsc for the components? Then it should be dist, right? And those files should be in the list of files to publish to NPM.

Changed `out-tsc` to `dist` for the component only, app still uses `out-tsc`

As long as a `files` property is not specified in `package.json`, it should just take everything, right? We currently dont specifiy a `files` property, so I dont think its needed?

Also, what should we do with the `@types/node` thing? Just add it to the package.json for now and then remove it later?

LarsDenBakker:
Then it takes everything (we should probably change that, but that's a separate thing).

I think we can add `@types/node` but open an issue to remove it later.

thepassle:
created an issue for that: https://github.com/open-wc/open-wc/issues/1604

And cleaned up the branch. This should be good to go now

thepassle:
pipeline seems to fail on `building-rollup` 🤨

```
lerna ERR! yarn run test:node exited 2 in '@open-wc/building-rollup'
lerna ERR! yarn run test:node stdout:
yarn run v1.16.0
$ mocha test/**/*.test.js test/*.test.js


  integration tests
    testcase js/rollup.spa.config.js
      1) "before all" hook for "passes the in-browser tests"
    testcase js/rollup.spa-nomodule.config.js
      2) "before all" hook for "passes the in-browser tests"


  0 passing (22s)
  2 failing

  1) integration tests
       testcase js/rollup.spa.config.js
         "before all" hook for "passes the in-browser tests":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/repo/packages/building-rollup/test/integration.test.js)
      at process._tickCallback (internal/process/next_tick.js:68:7)

  2) integration tests
       testcase js/rollup.spa-nomodule.config.js
         "before all" hook for "passes the in-browser tests":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/repo/packages/building-rollup/test/integration.test.js)
```

LarsDenBakker:
Retries the CI a few times, not sure. They're not failing on master. Did you change anything in those files?

LarsDenBakker:
Actually I don't see `create` tests showing up in the CI logs: https://circleci.com/gh/open-wc/open-wc/8554?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link is it timing out?

thepassle:
Ah shit. Looks like the `owc-app` [folder](https://github.com/open-wc/open-wc/tree/master/packages/create/owc-app) got deleted somehow (guess I deleted it while generating projects/testing), I'm guessing the owc-app is used in a test or something. I'll take a look after work

thepassle:
The problem was with the CLI args (it wasnt taking the typescript option, so it would just hang, causing a failing test). Its fixed now, so this should be good to go 👍

packages/create/src/generators/demoing-storybook-ts/templates/static-scaffold/stories/index.stories.md
LarsDenBakker: yes

thepassle: mosterd na de maaltijd lars 😂

LarsDenBakker: misschien kon je er vannacht niet van slapen

PR #1594, opened 2020-05-02
new site via 11ty + mdjs + building-rollup

daKmoR:
testing it via

```bash
# for dev
yarn docs:start

# for prod
yarn docs:build

# there are intermediate steps
## creates `_site-dev` folder: usable with dev-server
yarn docs:build-dev 

## creates `_site` folder: converts dev version to a production version (via rollup)
yarn docs:build-prod 
```


docs/server.config.js
```diff
@@ -0,0 +1,17 @@
...
+  nodeResolve: true,
+  open: './_site-dev',
+  responseTransformers: [
+    function rewriteBasePath({ contentType, body }) {
+      if (contentType.includes('text/html')) {
+        return {
+          body: body
+            .replace(/href="\//g, 'href="/_site-dev/')
+            .replace(/src="\//g, 'src="/_site-dev/'),
+          // body: body.replace(/<base href=".*">/, '<base href="/_site-dev/">'),
```
Westbrook, line 11: Is this something that should stick around?

docs/rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+    // development mode creates a non-minified build for debugging or development
+    developmentMode: false, // process.env.ROLLUP_WATCH === 'true',
+
+    // set to true to inject the service worker registration into your index.html
+    injectServiceWorker: false,
+    workbox: false,
+  });
+
+  const dest = '_site/';
+  mpaConfig.plugins.push(
```
Westbrook, line 21: Oooh, this is nice. Can't wait to steal this into the similar implementation I've been working no in Spectrum Web Components.

packages/mdjs/src/mdjsProcess.js
```diff
@@ -9,53 +9,65 @@ const raw = require('rehype-raw');
...
     .use(mdjsStoryParse, mdjsStoryParseOptions)
     .use(remark2rehype, { allowDangerousHTML: true })
+    .use(rehypePrism)
     .use(raw)
     .use(htmlSlug)
-    .use(htmlHeading)
+    .use(htmlHeading, htmlHeadingOptions)
     .use(htmlStringify);
 
+  // const defaultPlugins = [{ plugin: markdown, options:  }, remark2rehype]
```
Westbrook, line 32: Do we wanna keep this around? I think we'll need a plugins interface when we go "full public" on this as a plugin...but likely that comes with a more concerted effort later.

LarsDenBakker:
MPA config looks good to me, just need to place it outside of the rollup package for now.

docs/rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+    outputDir: '_site',
+    legacyBuild: true,
+    inputGlob: '_site-dev/**/*.html',
+    rootPath: path.resolve('_site-dev'),
+
+    // development mode creates a non-minified build for debugging or development
+    developmentMode: false, // process.env.ROLLUP_WATCH === 'true',
+
+    // set to true to inject the service worker registration into your index.html
+    injectServiceWorker: false,
```
Westbrook, line 16: Being we've set this up, what's between here and turning the service worker back on?

daKmoR: we need a good (non intrusive) way to notify users about new versions and/or load new version on menu interactions or so...

initially I had this on... but it triggered some frustrating situations where my update was actually correct I could just not see it without manually clearing the "app" cache...
so before we enable this we need to make it bullet prove

thepassle: that shouldnt even be a problem with MPAs, because its no longer a single page app, on every navigation, a new check for a service worker update is done. When a new service worker is found, we can listen for the controllerchange event and reload the page

Westbrook: Makes sense to wait, if there is any question, but I _am_ excited to see coverage for this lingering on the horizon!

thepassle: I can dive into this this weekend, but imo its not blocking to merge this

abdonrd:
With this new setup I see that the dev URL is like: `http://localhost:8000/_site-dev/`

@daKmoR @Westbrook @thepassle There are any option to remove the `/_site-dev/`?
To have a simple `http://localhost:8000` URL.

Thanks in advance!

daKmoR:
@abdonrd yes that is more or less our end goal 🤗 might still be a little away 🙈

on the bright side we already have a working POC 💪

PR #1609, opened 2020-05-07
feat(mdjs): support html preview/-stories

thepassle:


LarsDenBakker:
LGTM, would be nice to be able to configure the name for storybook but we can do that later.

thepassle:
Added a `type` type to the `Story` interface, and improved the check for the type of story. PTAL @daKmoR

PR #1614, opened 2020-05-08
WIP: docs: pwa-ify and update on reload

thepassle:
We need to have a new service worker in the same location and with the same name as our previous service worker for this to take effect, which is why I overrided the default workbox config.

What happens now is that every page will try to register a service worker. If there is a new service worker available, it'll skip waiting, and on controllerchange reload the page.

Would like someone else to confirm to make sure:

Steps:
- Run a build: `yarn docs:build`
- `cd _site && http-server -o` (dont use the Eds configuration for this, because it configures a base path)
- Go the page, confirm service worker is installed (devtools -> application -> Service Workers “#0 activated and is running”)
- Close the tab.
- Make some changes in a readme or anything in the docs
- Run another build: `cd .. && yarn docs:build`
- `cd _site && http-server -o`
- Go to the page, confirm the change is there
- Confirm there is a new service worker is installed (devtools -> application -> Service Workers “#1 activated and is running”)

### DIFF
--- a/docs/rollup.config.js
+++ b/docs/rollup.config.js
@@ -12,9 +12,21 @@ module.exports = async () => {
     // development mode creates a non-minified build for debugging or development
     developmentMode: false, // process.env.ROLLUP_WATCH === 'true',
 
-    // set to true to inject the service worker registration into your index.html
     injectServiceWorker: false,
-    workbox: false,
+    workbox: {
+      globIgnores: ['polyfills/*.js', 'legacy-*.js', 'nomodule-*.js'],
+      swDest: path.join(process.cwd(), '_site', 'service-worker.js'),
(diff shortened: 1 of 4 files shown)


docs/rollup.config.js
```diff
@@ -12,9 +12,21 @@ module.exports = async () => {
     // development mode creates a non-minified build for debugging or development
     developmentMode: false, // process.env.ROLLUP_WATCH === 'true',
 
-    // set to true to inject the service worker registration into your index.html
     injectServiceWorker: false,
-    workbox: false,
+    workbox: {
```
abdonrd, line 16: This config don't override the default, just merge with it, right?

thepassle: oo, good catch, thanks!

PR #1616, opened 2020-05-09
docs: add Publishing Web Components guide

Westbrook:
Super drafty as yet, but wanted to share. Particularly let me know about really bad things or really good things. Much in the middle, I'll hopefully change on my own in the next couple of days, anyways. I definitely got tired in the second half, so be warned.

### DIFF
--- a/docs/publishing/index.md
+++ b/docs/publishing/index.md
@@ -9,7 +9,7 @@ tags:
 
 When you are ready to publish your package to NPM, be sure that you've addressed the recommendations below to ensure the code you publish is as easy for your users to consume as possible. Your package should already have [demos](/demoing/), [documentation](/demoing/storybook-addon-markdown-docs.html), [tests](/testing/), et al, as each of these plays a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time. With this reality in mind, they will not be discussed directly below. If you're looking for information on including these things in your published packages, take a look at our [guides](/guide/index.html) or learn about [getting started](#getting-started) below.
 
-<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically useful for publishing to a CDN or any other context where broader application-specific optimisations may be appropriated.</p></div>
+<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically useful for publishing to a CDN or any other context where broader application-specific optimisations may be appropriate.</p></div>
 
 <style>
 .content__default ul {list-style:none;padding:0;}
--- a/docs/publishing/index.md
+++ b/docs/publishing/index.md
@@ -9,7 +9,7 @@ tags:
 
 When you are ready to publish your package to NPM, be sure that you've addressed the recommendations below to ensure the code you publish is as easy for your users to consume as possible. Your package should already have [demos](/demoing/), [documentation](/demoing/storybook-addon-markdown-docs.html), [tests](/testing/), et al, as each of these plays a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time. With this reality in mind, they will not be discussed directly below. If you're looking for information on including these things in your published packages, take a look at our [guides](/guide/index.html) or learn about [getting started](#getting-started) below.
 
-<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically tuned for publishing to a CDN or other contexts where broader application-specific optimizations will not be applied to the code available therein.</p></div>
+<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically useful for publishing to a CDN or any other context where broader application-specific optimisations may be appropriated.</p></div>
 
 <style>
 .content__default ul {list-style:none;padding:0;}
@@ -80,21 +80,21 @@ When importing something that doesn't point directly to the entry point of a mod
 
 **Yes:** 👍
 
-```
+```js
 import { html } from 'lit-html';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 ```
 
 **No:** 👎
 
-```
+```js
 import { html } from 'lit-html.js';
 import { ifDefined } from 'lit-html/directives/if-defined';
 ```
 
 ## Optional: use `type: "module"`
 
-Using `type: "module"` in your `package.json` will help CDNs and bundlers choose whether to parse your package with module parse goal. However, it will also enforce strict realities on how es modules and common JS modules must be imported into your scripting. If all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your scripting, then using `type: "module"` will allow those CDNs or bundlers to better leverage your package. If scripting in your package has issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
+Using `type: "module"` in your `package.json` will help CDNs and bundlers choose whether to parse your package with a module parse goal. However, it will also enforce that all the code you execute needs to be es modules. This includes all the dev tools (like bundlers, linters, ...) you use as well. CommonJs may still be used but requires a special way of importing. Therefore if all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your tools or scripts, then using `type: "module"` will allow those CDNs or bundlers to better leverage your package. If tools or script in your package have issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
 
 ## Optional: include typings
 
@@ -118,7 +118,7 @@ Minification is an application-level concern. What's more, minifiers get better 
 
 ## Do not use `.mjs` file-extensions
 
-It is still all too common that a `.mjs` file gets served to a browser with an incorrect mime type and causes an application to fail unrecoverably. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please see [Optional: use `type: "module"`](#optional-use-type-module) above as a possible alternative.
+It is still all too common that a `.mjs` file gets served to a browser with an incorrect mime type and causes an application to fail unrecoverably. This happens when the used server does not have `.mjs` configured which could either be because it never became a default in the server software or an older stable version is used. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please see [Optional: use `type: "module"`](#optional-use-type-module) above as a possible alternative.
 
 ## Do not import polyfills
 
--- a/docs/publishing/index.md
+++ b/docs/publishing/index.md
@@ -1,0 +1,135 @@
+---
+title: Publishing Web Components
+section: guides
+tags:
+  - guides
+---
+
+## Publishing Web Components to NPM
+
+When you are ready to publish your package to NPM, be sure that you've addressed the recommendations below to ensure the code you publish is as easy for your users to consume as possible. Your package should already have [demos](/demoing/), [documentation](/demoing/storybook-addon-markdown-docs.html), [tests](/testing/), et al, as each of these plays a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time. With this reality in mind, they will not be discussed directly below. If you're looking for information on including these things in your published packages, take a look at our [guides](/guide/index.html) or learn about [getting started](#getting-started) below.
+
+<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically tuned for publishing to a CDN or other contexts where broader application-specific optimizations will not be applied to the code available therein.</p></div>
+
+<style>
+.content__default ul {list-style:none;padding:0;}
+</style>
+
+### Do: 👍
+
+- ✅ [Do publish latest standard EcmaScript](#do-publish-latest-standard-ecmascript)
+- ✅ [Do publish standard es modules](#do-publish-standard-es-modules)
+- ✅ [Do include `"main": "index.js"` and `"module": "index.js"` in your `package.json`](#do-include-main-indexjs-and-module-indexjs-in-your-packagejson)
+- ✅ [Do export element classes](#do-export-element-classes)
+- ✅ [Do export side effects separately](#do-export-side-effects-separately)
+- ✅ [Do import 3rd party node modules with "bare" import specifiers](#do-import-3rd-party-node-modules-with-bare-import-specifiers)
+- ✅ [Do include file extensions in import specifiers](#do-include-file-extensions-in-import-specifiers)
+- _Optional_: [use `type: "module"` when possible](#optional-use-type-module)
+- _Optional_: [include typings; e.g. `*.d.ts` files](#optional-include-typings)
+- _Optional_: [include source maps](#optional-include-source-maps)
+
+### Don't: 👎
+
+- ❌ [Do not optimize](#do-not-optimize)
+- ❌ [Do not bundle](#do-not-bundle)
+- ❌ [Do not minify](#do-not-minify)
+- ❌ [Do not use `.mjs` file extensions](#do-not-use-mjs-file-extensions)
+- ❌ [Do not import polyfills](#do-not-import-polyfills)
+
+### General concept: Building is an application-level concern
+
+Your users will always have the most intimate knowledge of how and where their code will be delivered to their users. Give them the greatest amount of flexibility in preparing their code for those conditions by publishing your package with the agreement that building is an application-level concern. Any decision that you can make in support of your users having an easier time when and if they decide to bundle is a good decision to make when preparing your package for publication. Every "do" and every "don't" that is included herein will help you in doing just that.
+
+## Do publish latest standard EcmaScript
+
+Write standard EcmaScript from the start, and publishing standard EcmaScrip will come naturally. However, if you choose to write your component in TypeScript or with various emerging specifications or APIs, be sure to convert your code to standard EcmaScript before publication. If you use non-standard syntax; transpile that (and only that) down to a reasonably modern level (e.g. [TC39 Stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md) and/or available cross-browser) to decrease the likelihood that verbose or duplicate code is included in any final production delivery of pages featuring your package.
+
+## Do publish standard es modules
+
+In agreement with the "do" above, es modules are both Stage 4, supported in all modern browsers and tool-chains, ensure your package is published as es modules.
+
+## Do include `"main": "index.js"` and `"module": "index.js"` in your `package.json`
+
+Your `package.json` should have both its `main` entry point and `module` entry point to the same es module file. The `module` entry point informs tooling (bundlers, CDNs, etc.) which file to load by default in a module environment and, because es modules are standard EcmaScript, your `main` entry point should point to the same file. You will find this already configured for you if you've started your project with `npm init @open-wc`, but ensure your `package.json` file includes something similar to the following:
+
+```
+{
+  "main": "index.js",
+  "module": "index.js",
+  // ...
+}
+```
+
+## Do export element classes
+
+Every custom element is a class extension of `HTMLElement`; e.g. `class MyElement extends HTMLElement {}`. Be sure that you export this class from your package. When starting from `npm init @open-wc` you will see this export in `src/MyElement.js` and referenced by the `index.js` entry point. This will make it possible for both customization of your element via extension (e.g. `class NewElement extends MyElement {}`) as well as for advanced usage of the custom element in Scoped Registries, both [native](https://github.com/w3c/webcomponents/issues/716) (when available) or [synthetic](/scoped-elements/).
+
+## Do export side effects separately
+
+A [side effect](<https://en.wikipedia.org/wiki/Side_effect_(computer_science)>) in javascript is created when the script changes state outside of the scope in which it is run. This includes when code in an es module alters values available on the `window` or `document`. A side effect occurs when registering Custom Elements via `customElements.define()`. JS files with side effects should exist separately from your packages entry point. `npm init @open-wc` takes care of this for you by creating a separate `my-element.js` that manages the registration of your custom element.
+
+If you do choose to publish code with side effects as part of your entry point, it should be placed in a separate function that has to explicitly be called by the consumer (e.g. `run()`, `init()`, etc.). This allows your users to determine when these effects are triggered.
+
+## Do import 3rd party node modules with "bare" import specifiers
+
+When referring to 3rd party dependencies in your code, you cannot guarantee that implementing projects rely on the same file structure as you do. To prevent dependency specifiers from causing issues across different environments, be sure to import these dependencies into your project with "bare" import specifiers so that tooling in your users' environment can manage resolving the location of those dependencies; e.g. `import { LitElement } from 'lit-element';`
+
+## Do include file extensions in import specifiers
+
+When importing something that doesn't point directly to the entry point of a module, e.g.: `import {LitElement} from 'lit-element';`, always make sure to include a file extension: `import { ifDefined } from 'lit-html/directives/if-defined.js';`
+
+**Yes:** 👍
+
+```
+import { html } from 'lit-html';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+```
+
+**No:** 👎
+
+```
+import { html } from 'lit-html.js';
+import { ifDefined } from 'lit-html/directives/if-defined';
+```
+
+## Optional: use `type: "module"`
+
+Using `type: "module"` in your `package.json` will help CDNs and bundlers choose whether to parse your package with module parse goal. However, it will also enforce strict realities on how es modules and common JS modules must be imported into your scripting. If all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your scripting, then using `type: "module"` will allow those CDNs or bundlers to better leverage your package. If scripting in your package has issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
+
+## Optional: include typings
+
+`*.d.ts` files are an important bridge between your code and users that choose to develop in TypeScript. If you are already developing in TypeScript, then generating these is as simple as adding the appropriate property to your `tsconfig.json`. If you are developing in JS, it can still be useful to include typings; learn more about [generating type definitions from javascript](https://dev.to/open-wc/generating-typescript-definition-files-from-javascript-5bp2).
+
+## Optional: include source maps
+
+If you choose work in TypeScript, or to use syntax that is newer than the standards discussed above, it can be important to include source maps and the source files (TypeScript or otherwise) that your published code originates from to make understanding that code, as well as possibly reporting/fixing issues in that code, easier for your users.
+
+## Do not optimize
+
+Optimization is an application-level concern. It's not repeating yourself if it is a really important point, it's merely ensuring that it sinks in. Your package will be leveraged by a developer that will know more about their specific delivery targets than you will and you must not attempt to optimize your code beyond delivering the best possible standard EcmaScript that you can.
+
+## Do not bundle
+
+Bundling is an application-level concern. If you bundle, and then another dependency bundles and both dependencies bundle the same transitive dependencies it will become next to impossible for later tooling to deduplicate that code. Be sure NOT to bundle code in your published packages so that your users will have the easiest time possible optimizing their application for the specific use cases they target.
+
+## Do not minify
+
+Minification is an application-level concern. What's more, minifiers get better all the time, and you can ensure that your users can leverage the best in class approach to this space available by skipping the minifier when publishing your package.
+
+## Do not use `.mjs` file-extensions
+
+It is still all too common that a `.mjs` file gets served to a browser with an incorrect mime type and causes an application to fail unrecoverably. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please see [Optional: use `type: "module"`](#optional-use-type-module) above as a possible alternative.
+
+## Do not import polyfills
+
+If you feel you need polyfills, for package demos or the like, feel free to add them as `devDependencies`, but NEVER import them into modules. As has been repeated many times above, only your users will fully understand the use cases they target in such a way to determine whether polyfills should be included in an application, don't force a decision in this area onto them.
+
+### Additional Reading
+
+- [How to Public Web Componoent to NPM by Justing Fagnani](https://justinfagnani.com/2019/11/01/how-to-publish-web-components-to-npm/)
+- [npm-publish](https://docs.npmjs.com/cli/publish)
+- [Contributing Packages to the NPM Registry](https://docs.npmjs.com/packages-and-modules/contributing-packages-to-the-registry)
+
+# Getting started...
+
+There are many things to take into account when publishing your work to be shared publicly, even beyond those listed above. Some start from the very beginning of your development effort; director/file naming & organization, documentation, features, tests, etc. Others can all to easily be seen as an afterthought: bundling, `package.json` entries, registry targets, transpiling, types, etc. All play an important role in making the long term use and maintenance of your work as smooth as possible for you and your consumers. Luckily, beginning your component or application with the `npm init @open-wc` command will get you stared down the path towards making great choices in each of these areas. [Get started today!](/init/)
--- a/docs/_includes/partials/sidebar-guide.njk
+++ b/docs/_includes/partials/sidebar-guide.njk
@@ -1,6 +1,6 @@
 <ul class="sidebar-links">
   <li>
-    <a href="/guide/" class="active sidebar-link">Introduction</a>
+    <a href="/guide/" class="{% if page.url == '/guide/' %}active {% endif %}sidebar-link">Introduction</a>
   </li>
   <li>
     <section class="sidebar-group collapsable depth-0">
@@ -8,27 +8,11 @@
         <span>Guides &amp; Docs</span>
         <span class="arrow right"></span></p>
       <ul class="sidebar-links sidebar-group-items" style="">
-        <li>
-          <a href="/developing/best-practices.html" class="{% if permalink == 'developing/best-practices.html' %}active {% endif %}sidebar-link">Best Practices</a>
-        </li>
-        <li>
-          <a href="/codelabs/" class="{% if permalink == 'codelabs/index.html' %}active {% endif %}sidebar-link">Codelabs</a>
-        </li>
-        <li>
-          <a href="/developing/code-examples.html" class="{% if permalink == 'developing/code-examples.html' %}active {% endif %}sidebar-link">Code examples</a>
-        </li>
-        <li>
-          <a href="/developing/lit-html.html" class="{% if permalink == 'developing/lit-html.html' %}active {% endif %}sidebar-link">lit-html</a>
-        </li>
-        <li>
-          <a href="/developing/types.html" class="{% if permalink == 'developing/types.html' %}active {% endif %}sidebar-link">Types</a>
-        </li>
-        <li>
-          <a href="/developing/ide.html" class="{% if permalink == 'developing/ide.html' %}active {% endif %}sidebar-link">IDE</a>
-        </li>
-        <li>
-          <a href="/guide/component-libraries.html" class="{% if permalink == 'guide/component-libraries.html' %}active {% endif %}sidebar-link">Web Component Libraries</a>
-        </li>
+        {% for item in collections.guide %}
+          <li>
+            <a href="{{ item.url }}" class="{% if page.url == item.url %}active {% endif %}sidebar-link">{{ item.data.title }}</a>
+          </li>
+        {% endfor %}
       </ul>
     </section>
   </li>
@@ -42,7 +26,6 @@
           <section class="sidebar-group is-sub-group depth-1">
             <p class="sidebar-heading open">
               <span>Linting</span>
-              <!---->
             </p>
             <ul class="sidebar-links sidebar-group-items">
               <li>
@@ -61,7 +44,6 @@
           <section class="sidebar-group is-sub-group depth-1">
             <p class="sidebar-heading">
               <span>Developing</span>
-              <!---->
             </p>
             <ul class="sidebar-links sidebar-group-items">
               <li>
@@ -74,7 +56,6 @@
           <section class="sidebar-group is-sub-group depth-1">
             <p class="sidebar-heading">
               <span>Testing</span>
-              <!---->
             </p>
             <ul class="sidebar-links sidebar-group-items">
               <li>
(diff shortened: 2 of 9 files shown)


docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+title: Publishing Web Components
+section: guides
+tags:
+  - guides
+  - guide
+---
+
+## Publishing Web Components
+
+If you have developed a package that you would like to make available to other developers via a public registry, take the following concepts into account to ensure that external users of your code may make the most of it in their projects. If you've made it this far, your package will likely already include things like [demos](/demoing/), [documentation](/demoing/storybook-addon-markdown-docs.html), [tests](/testing/), et al. Each of these play a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time, but are not discussed directly below. If you're looking for information on including these things in your published packages, take a look at our [guides](/guide/index.html) or learn about [getting started](#getting-started) below.
```
thepassle, line 11: - "via a public registry"
maybe just be explicit and say NPM here? A public registry seems needlessly unspecific
- "Each of these play a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time, but are not discussed directly below." 
this sentence is a bit long and by the end of it im slightly confused, maybe chop this up a bit?

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+Keep in mind that these suggestions are not specifically tuned for publishing to a CDN or other contexts where broader application-specific optimizations will not be applied to the code available therein.
```
thepassle, line 13: I feel like there could be more emphasis on this, maybe make it bold, or add an ⚠️ or something?

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+### Do: 👍
+
+- ✅ [Do publish latest standard EcmaScript](#do-publish-latest-standard-ecmascript)
+- ✅ [Do publish standard es modules](#do-publish-standard-es-modules)
+- ✅ [Do include appropriate `package.json` entries](#do-include-appropriate-packagejson-entries)
+- ✅ [Do export element classes](#do-export-element-classes)
+- ✅ [Do export side effects separately](#do-export-side-effects-separately)
+- ✅ [Do keep side effects out of your index.js](#do-keep-side-effects-out-of-your-indexjs)
+- ✅ [Do import 3rd party node modules with "bare" import specifiers](#do-import-3rd-party-node-modules-with-bare-import-specifiers)
+- ✅ [Do include file extensions in import specifiers, always](#do-include-file-extensions-in-import-specifiers-always)
```
thepassle, line 28: ", always" seems unnecessary

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+- ❌ [Do not bundle](#do-not-bundle)
+- ❌ [Do not minify](#do-not-minify)
+- ❌ [Do not optimize](#do-not-optimize)
+- ❌ [Do not use `.mjs` file extensions](#do-not-use-mjs-file-extensions)
+- ❌ [Do not import polyfills](#do-not-import-polyfills)
+
+## Do publish latest standard EcmaScript
+
+Contrary to some long-standing beliefs, an implementing project will always know more about the delivery context of the code it leverages than the developer of any of its dependencies. Taking this fact into account, the act of transpiling your code below a reasonably modern level (e.g. [TC39 Stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md) and/or available cross-browser) is the best way that you can prepare your package to empower downstream projects to make their own decisions about the code they ship to their users.
```
thepassle, line 43: While I'm a big fan of your style of writing, I feel like we need to be extremely specific and clear in these instructions, and we may overall need to "dumb it down" somewhat. 

In this paragraph specifically this sentence could be simplified a lot:
> "an implementing project will always know more about the delivery context of the code it leverages than the developer of any of its dependencies."

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+In some cases, you may choose to use newer specifications that may require special care when preparing for older browsers. In this case, transpiling down to the level you choose from above may be a good idea. If you choose not to do this, be sure to document this reality thoroughly so that your users are not left unawares.
```
thepassle, line 45: Maybe add an example of some kind of syntax/new spec that would need to be transpiled?

> If you choose not to do this, be sure to document this reality thoroughly so that your users are not left unawares.
We should not recommend people to ship newer, non-widely supported syntax in that case, right? Even if well documented?

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do publish standard es modules
+
+Your package (and those packages that it depends on) should be distributed as standard es modules. This supports the "do" above to deliver the latest standard EcmaScript to your users, modules are both Stage 4 and supported in all major browsers, but also prepares your users to have the simplest time building their applications for production as es modules are the only module format natively supported by browsers.
```
thepassle, line 49:
```suggestion
Your package (and the packages that it depends on) should be distributed as standard es modules. This supports the "do" above to deliver the latest standard EcmaScript to your users, modules are both Stage 4 and supported in all major browsers, but also prepares your users to have the simplest time building their applications for production as es modules are the only module format natively supported by browsers.
```

> as es modules are the only module format natively supported by browsers.

as opposed to?

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do include appropriate `package.json` entries
+
+Your `package.json` should both its `main` entrypoint and `module` entrypoint to the same es module file. You will find this already configured for you by `npm init @open-wc`, but ensure your `package.json` file includes something simialar to the following:
```
thepassle, line 53:
```suggestion
Your `package.json` should have both its `main` entrypoint and `module` entrypoint point to the same es module file. You will find this already configured for you if you've started your project with `npm init @open-wc`, but ensure your `package.json` file includes something simialar to the following:
```

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+{
+  "main": "index.js",
+  "module": "index.js",
+  // ...
+}
+```
+
+## Do export element classes
+
+Every custom element is a class extension of `HTMLElement`; e.g. `class MyElement extends HTMLElement {}`. Be sure that you export this class from your package. When starting from `npm init @open-wc` you will see this export in `src/MyElement.js` and reported by the `index.js` entry point. This will make it possible for both customization of your element via extension (e.g. `class NewElement extends MyElement {}`) as well as for advanced usage of the custom element in Scoped Registries, both [native](https://github.com/w3c/webcomponents/issues/716) (when available) or [synthetic](/scoped-elements/).
```
thepassle, line 65:
```suggestion
Every custom element is a class extension of `HTMLElement`; e.g. `class MyElement extends HTMLElement {}`. Be sure that you export this class from your package. When starting from `npm init @open-wc` you will see this export in `src/MyElement.js` and referenced by the `index.js` entry point. This will make it possible for both customization of your element via extension (e.g. `class NewElement extends MyElement {}`) as well as for advanced usage of the custom element in Scoped Registries, both [native](https://github.com/w3c/webcomponents/issues/716) (when available) or [synthetic](/scoped-elements/).
```

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do export side effects separately
+
+A [side effect](<https://en.wikipedia.org/wiki/Side_effect_(computer_science)>) in javascript is created when script effects state outside of the scope in which it is run. This includes when code in an es module effects values available on the `window` or `document` as is inherent to the reality of Custom Elements in that `window.customElements.define()` registers an element in the global registry attached to `window`, whether you explicitly include `window` in your code when doing so or not. JS files with side effects should exist separately from your packages entry point. `npm init @open-wc` exemplifies this through its creation of `my-element.js` which manages the registration of the custom element that this command generates.
```
thepassle, line 69: > This includes when code in an es module effects values

Affects? not sure

> as is inherent to the reality of Custom Elements in that `window.customElements.define()`

I feel like this could be 'dumbed' down a bit

> `npm init @open-wc` exemplifies this through its creation of `my-element.js` which manages the registration of the custom element that this command generates.

not sure if this sentence is needed here.

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do keep side effects out of your index.js
+
+When not able to keep code with side effects out of your entry point, be sure to contain such commands within a method (e.g. `run()`, `init()`, etc.) to prevent those side effects from occurring at run time.
```
thepassle, line 73: ^ This is confusing to me, can we clarify this?

If you're _not_ able to keep side effectful code out of your entrypoint, try to contain it in a method instead? Then its still side effectful, right? It should instead be put in a separate function that has to explicitly be called by the consumer?

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do import 3rd party node modules with "bare" import specifiers
+
+When referring to 3rd party dependencies in your code, you cannot guarantee that implementing projects rely on the same file structure your package as you do. To prevent dependency pointing from causing issue across different environments be sure to import these dependencies into your project with "bare" import specifiers so that tooling in your users' environment can manage resolving the location of those dependencies; e.g. `import { LitElement } from 'lit-element';`
+
+## Do include file extensions in import specifiers, always
+
+When not pointing directly the entry point files referenced by "bar" import specifiers, be sure to reference files (both in your package and your dependencies) with their complete file extension; e.g. `import { ifDefined } from 'lit-html/directives/if-defined.js';`
```
thepassle, line 81: > When not pointing directly the entry point files referenced by "bar" import specifiers,

this sentence is unclear to me

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Optional: use `type: "module"`
+
+If scripting in your packages is resilient to the realities made true in Node@12+ when using `type: "module"` in your `package.json`, you should use it. If all of the JS in your package is es modules, then this is a cheap win for your package and as it can support tooling the references your package in leveraging it more efficiently, however excluding it is a simple way in the near term to not require this of all of the JS your write.
```
thepassle, line 85: we should expand on what 'the realities' are, or link to more information. People reading this may read this, just put type: module, and have their stuff blow up

thepassle: > cheap win for your package and as it can support tooling the references your package in leveraging it more efficiently

this sentence is unclear to me, as well as:

> however excluding it is a simple way in the near term to not require this of all of the JS your write.

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Optional: include source maps
+
+If you choose work in TypeScript, or to use syntax that is newer than the standards discussed above, it can be important to include source maps and the source files (TypeScript or otherwise) that your published code originates from to make understanding that code, as well as possibly reporting/fixing issues in that code, easier for your users.
+
+## Do not bundle
+
+If you bundle, and then another dependency bundles and both dependencies have bundles the same transitive dependencies it will become next to impossible for later tooling to deduplicate that code. Be sure NOT to bundle code in your published packages so that your users will have the easiest time possible optimizing their application for the specific use cases they target.
```
thepassle, line 97:
```suggestion
If you bundle, and then another dependency bundles and both dependencies bundle the same transitive dependencies it will become next to impossible for later tooling to deduplicate that code. Be sure NOT to bundle code in your published packages so that your users will have the easiest time possible optimizing their application for the specific use cases they target.
```

should we add an extra note or TL;DR here that "bundling is an app-level concern"?

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do not minify
+
+Minification is an application-level concern, and much like other recommendations here NOT minifying your published packages will give your users an easier time optimizing their application for the specific use cases they target.
+
+## Do not optimize
+
+At this point, maybe I'll not touch the dead horse anymore.
```
thepassle, line 105: lmao

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Do not use `.mjs` file-extensions
+
+It is still all too common that a `.mjs` file served to a browser is done so with an incorrect mime type causing an application to fail unrecoverably. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please visit [Optional: use `type: "module"`](#optional-use-type-module) above to possible alternative.
```
thepassle, line 109:
```suggestion
It is still all too common that a `.mjs` file gets served to a browser with an incorrect mime type and causes an application to fail unrecoverably. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please see [Optional: use `type: "module"`](#optional-use-type-module) above as a possible alternative.
```

docs/publishing/index.md
Westbrook: Do you think we should name the page "Publishing Web Components to NPM"? I added "via a public registry" because I thought it would allow for the vaguery of publishing to places like GitHub Package Registry, but maybe specificity is better, top to bottom?

Good point on the following sentence!

docs/publishing/index.md
Westbrook: Good call!

docs/publishing/index.md
Westbrook: Yeah, thought it made a good parallel to the "When not pointing directly the entry point files referenced by "bar" import specifiers" below, but probably too _cute_ for real value.

docs/publishing/index.md
Westbrook: Good call, what the newest possible hotness to use as an example?

Good question, I think if I say "this uses X fancy API, be aware" that things would be cool. For instance, I probably wouldn't have transpiled `import.meta` in my code at any point in the past but would have wanted to document that it was used for users to be aware, but maybe that's wrong?

docs/publishing/index.md
Westbrook: CJS, AMD, UMD? Stole that a little from Justin's article. Will rethink.

docs/publishing/index.md
Westbrook: Will check.

Will think.

I want to try to point back to our tools as much as possible. If this wasn't `exemplifies` but `delivers you this`, or similar?

docs/publishing/index.md
Westbrook: This is good clarity:
> It should instead be put in a separate function that has to explicitly be called by the consumer?
Was trying to say that...but, you know, fell short...will acquire!

docs/publishing/index.md
Westbrook: Trying to say something in reference when not requesting the entry point file specifically (via "bare" import specifiers)...would love thoughts on dumbing it down, because it's kind of a confusing concept in general.

docs/publishing/index.md
Westbrook: I was essentially looking to say "if it blows up, don't use it"... @LarsDenBakker you know much more about this than me, any references I could use here, or realities I should address directly?

docs/publishing/index.md
Westbrook: This is a good idea. Would it make sense to include "bundling is an app-level concern" before (or right after) the list of do's/don't's?

docs/publishing/index.md
thepassle: > Good call, what the newest possible hotness to use as an example?

Maybe optional chaining is still a good example of this? 🤔 Or is it 'outdated' already? 😛

>  For instance, I probably wouldn't have transpiled import.meta in my code at any point in the past but would have wanted to document that it was used for users to be aware, but maybe that's wrong?

I think we discussed that stuff like this should be transpiled though, right? But I could be remembering it wrong

docs/publishing/index.md
thepassle: Maybe its fine to just leave it out

docs/publishing/index.md
thepassle: Maybe:

"`npm init @open-wc` takes care of this for you by creating a separate `my-element.js` to manage the actual registration of the custom element."

docs/publishing/index.md
thepassle: When importing something that doesn't point directly to the _entry point_ of a module, e.g.: `import {LitElement} from 'lit-element';`, *always* make sure to include a file extension. `import { ifDefined } from 'lit-html/directives/if-defined.js';`

Maybe another do/dont list?

✅ Do:
`import { html } from 'lit-html';`
`import { ifDefined } from 'lit-html/directives/if-defined.js';`

❌ Dont:
`import { html } from 'lit-html.js';`
`import { ifDefined } from 'lit-html/directives/if-defined';`

docs/publishing/index.md
thepassle: I think thats a good idea yeah.

I also think its not a bad idea to have that line return a couple of times (e.g. in dont minify, dont bundle), it really does drive the whole point home.

docs/publishing/index.md
thepassle: Hm yeah I see your point, but as far as I can tell the github registry hasnt really picked up steam or a lot of usage, especially now that npm is part of github I guess. I also feel like mentioning NPM will probably help with searchability, but thats a gut feeling. Im assuming people might be searching for "how to publish web component npm" (or thats what id google anyway)

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+.content__default ul {list-style:none;padding:0;}
+</style>
+
+### Do: 👍
+
+- ✅ [Do publish latest standard EcmaScript](#do-publish-latest-standard-ecmascript)
+- ✅ [Do publish standard es modules](#do-publish-standard-es-modules)
+- ✅ [Do include appropriate `package.json` entries](#do-include-appropriate-packagejson-entries)
+- ✅ [Do export element classes](#do-export-element-classes)
+- ✅ [Do export side effects separately](#do-export-side-effects-separately)
```
daKmoR, line 25: I think this one is redundant with the next one?
replace it with?

```suggestion
- ✅ [Do have all entry points at the root](#...)
```

e.g. people should not need to know or care about your folder structure... each individual entrypoint you have should be a dedicated file in the root of your package

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+- ✅ [Do include file extensions in import specifiers, always](#do-include-file-extensions-in-import-specifiers-always)
+- _Optional_: [use `type: "module"` when possible](#optional-use-type-module)
+- _Optional_: [include typings; e.g. `*.d.ts` files](#optional-include-typings)
+- _Optional_: [include source maps](#optional-include-source-maps)
+
+### Don't: 👎
+
+- ❌ [Do not bundle](#do-not-bundle)
+- ❌ [Do not minify](#do-not-minify)
+- ❌ [Do not optimize](#do-not-optimize)
```
daKmoR, line 37: "Do not optimize" should be at the top?
sort of like

- Do not optimize
- Explicitly do not minify
- Explicitly do not bundle

then in explaining you could give a why you should not do anything... and in the explicit section you explain specifically why not minify, why not bundle

docs/publishing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+<style>
+.content__default ul {list-style:none;padding:0;}
+</style>
+
+### Do: 👍
+
+- ✅ [Do publish latest standard EcmaScript](#do-publish-latest-standard-ecmascript)
+- ✅ [Do publish standard es modules](#do-publish-standard-es-modules)
+- ✅ [Do include appropriate `package.json` entries](#do-include-appropriate-packagejson-entries)
```
daKmoR, line 23: for lazy people like me this list should have all the info needed to publish something without reading everything in detail... 

e.g. include the appropriate package.json entries in this list

docs/publishing/index.md
Westbrook: Feel like this is pretty arguable...any seconds on "Do have all entry points at the root"? Feel like an implementation detail we could avoid needing to worry about (particularly in that people should be able to choose what API they surface to the world), but I can see how it's _nicer_.

Merging the two side effects points does make sense, however.

docs/publishing/index.md
Westbrook: I have altered this a good bit, maybe it's less troublesome now? Push coming in 20 or 30 minutes.

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+- ❌ [Do not optimize](#do-not-optimize)
+- ❌ [Do not bundle](#do-not-bundle)
+- ❌ [Do not minify](#do-not-minify)
+- ❌ [Do not use `.mjs` file extensions](#do-not-use-mjs-file-extensions)
+- ❌ [Do not import polyfills](#do-not-import-polyfills)
+
+### General concept: Building is an application level concern
+
+Your users will always have the most intimate knowledge of how and where thier code will be used to their users. Give them the greatest amount of flexibility in preparing their code for those conditions by publishing your package with the agreement that building is an application level concern. Any decision that you can make in support of your users having an easier time when and if they decide to bundle is a good decision to make when preparing your package for publication. Every "do" and every "don't" that is included herein will help you in doing just that.
```
thepassle, line 43:
```suggestion
Your users will always have the most intimate knowledge of how and where their code will be used to their users. Give them the greatest amount of flexibility in preparing their code for those conditions by publishing your package with the agreement that bundling is an application level concern. Any decision that you can make in support of your users having an easier time when and if they decide to bundle is a good decision to make when preparing your package for publication. Every "do" and every "don't" that is included herein will help you in doing just that.
```

"to their users" sounds slightly off, is that correct english? Should it be "for" their users?

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Do publish latest standard EcmaScript
+
+Write standard EcmaScript from the start, and publishing standard EcmaScrip will come naturally. However, if you choose to write write your component in TypeScript or with various emerging specifications or APIs, be sure to convert your code to standard EcmaScript before publication. Transpiling your published javascript below a reasonably modern level (e.g. [TC39 Stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md) and/or available cross-browser) will increase the likelihood that verbose or duplicate code is included in any final production delivery of pages featuring your package. Take this into account in cases where you use newer specifications that may require special care when delivering to various browsers. If for any reason you choose to ship non-transpiled code featuring newer specification(s), be sure to document this reality thoroughly so that your users can prepare to leverage your package appropriately.
```
thepassle, line 47:
```suggestion
Write standard EcmaScript from the start, and publishing standard EcmaScrip will come naturally. However, if you choose to write your component in TypeScript or with various emerging specifications or APIs, be sure to convert your code to standard EcmaScript before publication. Transpiling your published javascript below a reasonably modern level (e.g. [TC39 Stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md) and/or available cross-browser) will increase the likelihood that verbose or duplicate code is included in any final production delivery of pages featuring your package. Take this into account in cases where you use newer specifications that may require special care when delivering to various browsers. If for any reason you choose to ship non-transpiled code featuring newer specification(s), be sure to document this reality thoroughly so that your users can prepare to leverage your package appropriately.
```

> be sure to convert your code to standard EcmaScript before publication. Transpiling your published javascript below a reasonably modern level (e.g. [TC39 Stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md) and/or available cross-browser) will increase the likelihood that verbose or duplicate code is included in any final production delivery of pages featuring your package.

I think we can be more clear here; this reads to me as "Totally transpile your code!". I think we should put more emphasis on the stage 4 'baseline' of transpilation, and mention explicitly to _not_ transpile to es5 for example

> If for any reason you choose to ship non-transpiled code

Im still not sure if we should even say this; the recommendation is to transpile those

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+{
+  "main": "index.js",
+  "module": "index.js",
+  // ...
+}
+```
+
+## Do export element classes
+
+Every custom element is a class extension of `HTMLElement`; e.g. `class MyElement extends HTMLElement {}`. Be sure that you export this class from your package. When starting from `npm init @open-wc` you will see this export in `src/MyElement.js` and referenced by the `index.js` entry point. This will make it possible for both customization of your element via extension (e.g. `class NewElement extends MyElement {}`) as well as for advanced usage of the custom element in Scoped Registries, both [native](https://github.com/w3c/webcomponents/issues/716) (when available) or [synthetic](/scoped-elements/).
```
thepassle, line 67: 👌

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Do export side effects separately
+
+A [side effect](<https://en.wikipedia.org/wiki/Side_effect_(computer_science)>) in javascript is created when script changes state outside of the scope in which it is run. This includes when code in an es module alters values available on the `window` or `document`. A side effect occurs when registering Custom Elements via `customElements.define()`. JS files with side effects should exist separately from your packages entry point. `npm init @open-wc` takes care of this for you by creating a separate `my-element.js` that manages the registration of your custom element.
+
+## Do keep side effects out of your index.js
+
+When _not_ able to keep code with side effects out of your entry point, it should instead be put in a separate function that has to explicitly be called by the consumer (e.g. `run()`, `init()`, etc.). This prevents those side effects from occurring at run time.
```
thepassle, line 75: It doesnt really prevent the side effects from occurring at run time, they still occur whenever I opt in to calling `run()` myself, but at least as a user, I've made the conscious decision to call that function and register the side effects, and as author I've been explicit in my communication: "Hey, here's the side effectful stuff". 

I think the last sentence can be dropped

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Do import 3rd party node modules with "bare" import specifiers
+
+When referring to 3rd party dependencies in your code, you cannot guarantee that implementing projects rely on the same file structure as you do. To prevent dependency pointing from causing issue across different environments be sure to import these dependencies into your project with "bare" import specifiers so that tooling in your users' environment can manage resolving the location of those dependencies; e.g. `import { LitElement } from 'lit-element';`
```
thepassle, line 79:
```suggestion
When referring to 3rd party dependencies in your code, you cannot guarantee that implementing projects rely on the same file structure as you do. To prevent dependency specifiers from causing issues across different environments, be sure to import these dependencies into your project with "bare" import specifiers so that tooling in your users' environment can manage resolving the location of those dependencies; e.g. `import { LitElement } from 'lit-element';`
```

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Do include file extensions in import specifiers
+
+When importing something that doesn't point directly to the entry point of a module (e.g.: `import {LitElement} from 'lit-element';`) always make sure to include a file extension (e.g. `import { ifDefined } from 'lit-html/directives/if-defined.js';`).
+
+## Optional: use `type: "module"`
+
+Using `type: "module"` in your `package.json` is used to help CDNs and bundlers choose whether to parse your package with module parse goal. However, it will also enforce strict realities on how es modules and common JS modules must be imported into your scripting. If all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your scripting, then this is a great way to support tooling the references your package in leveraging it more efficiently. If scripting in your package has issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
```
thepassle, line 87:
```suggestion
Using `type: "module"` in your `package.json` is used to help CDNs and bundlers choose whether to parse your package as an actual module. However, it will also enforce strict realities on how es modules and common JS modules must be imported into your scripting or other tooling. If all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your scripting or tooling, then this is a great way to support tooling the references your package in leveraging it more efficiently. If scripting or tooling in your package has issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
```

> tooling the references your package in leveraging it more efficiently

not sure i totally grok what you mean with this, its a bit confusing

docs/publishing/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Optional: include source maps
+
+If you choose work in TypeScript, or to use syntax that is newer than the standards discussed above, it can be important to include source maps and the source files (TypeScript or otherwise) that your published code originates from to make understanding that code, as well as possibly reporting/fixing issues in that code, easier for your users.
+
+## Do not optimize
+
+Building is an application level concern. It's not repeating yourself if it is a really important point, it's merely ensuring that it sinks in. Your package will be leveraged by a developer that will know more about their specific delivery targets that you will and it is important that you don't attempt to optimize your code beyond delivering the best possible standard EcmaScript that you can.
```
thepassle, line 99:
```suggestion
Bundling is an application level concern. It's not repeating yourself if it is a really important point, it's merely ensuring that it sinks in. Your package will be leveraged by a developer that will know more about their specific delivery targets that you will and it is important that you don't attempt to optimize your code beyond delivering the best possible standard EcmaScript that you can.
```

docs/publishing/index.md
Westbrook: `by`, I think I meant `delivered` instead of `used`... 🤔

docs/publishing/index.md
Westbrook: It _may_ be that we want to lean away from the hedge and just say "never transpile your code"? 🦆🦆🦆🦆🦆🦆

I'll take another pass.

docs/publishing/index.md
Westbrook: I've gone back and forth on this word... does "bundling" really include translating, minifying, optimizing, merging modules together, and structuring it for production? Might be thinking too hard, but it seems exclusive only to be inclusive of other things later.

thepassle: > does "bundling" really include translating, minifying, optimizing, merging modules together, and structuring it for production?

IMO; yes. But I see what you're saying. Really what we want to say is "minifying, optimizing, transpiling, etc is an application level concern". 

So... I'm not really sure how to go about this one 😅 Maybe get some feedback from others on this?

docs/publishing/index.md
thepassle: Imo we should be clear about this: 
If you use non-standard syntax; transpile that down. That is the recommendation. If we mention 'if you do not choose to do this', to me it feels like we're giving them an escape hatch: "oh, I guess some people don't transpile those features down? then maybe its fine for me too 🤷 ". I just think we shouldnt mention that at all, because the recommendation is; transpile non-standard syntax.

docs/publishing/index.md
Westbrook: Still having trouble with this one, new version coming to the preview soon.

docs/publishing/index.md
Westbrook: I think I have a good idea here, new version coming soon.

docs/publishing/index.md
thepassle: Can you add this one?

docs/publishing/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Optional: include source maps
+
+If you choose work in TypeScript, or to use syntax that is newer than the standards discussed above, it can be important to include source maps and the source files (TypeScript or otherwise) that your published code originates from to make understanding that code, as well as possibly reporting/fixing issues in that code, easier for your users.
+
+## Do not optimize
+
+Optimization is an application level concern. It's not repeating yourself if it is a really important point, it's merely ensuring that it sinks in. Your package will be leveraged by a developer that will know more about their specific delivery targets that you will and it is important that you don't attempt to optimize your code beyond delivering the best possible standard EcmaScript that you can.
```
thepassle, line 110:
```suggestion
Optimization is an application level concern. It's not repeating yourself if it is a really important point, it's merely ensuring that it sinks in. Your package will be leveraged by a developer that will know more about their specific delivery targets than you will and it is important that you don't attempt to optimize your code beyond delivering the best possible standard EcmaScript that you can.
```

docs/publishing/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+- ❌ [Do not optimize](#do-not-optimize)
+- ❌ [Do not bundle](#do-not-bundle)
+- ❌ [Do not minify](#do-not-minify)
+- ❌ [Do not use `.mjs` file extensions](#do-not-use-mjs-file-extensions)
+- ❌ [Do not import polyfills](#do-not-import-polyfills)
+
+### General concept: Building is an application level concern
+
+Your users will always have the most intimate knowledge of how and where thier code will be delivered to their users. Give them the greatest amount of flexibility in preparing their code for those conditions by publishing your package with the agreement that building is an application level concern. Any decision that you can make in support of your users having an easier time when and if they decide to bundle is a good decision to make when preparing your package for publication. Every "do" and every "don't" that is included herein will help you in doing just that.
```
thepassle, line 42:
```suggestion
Your users will always have the most intimate knowledge of how and where their code will be delivered to their users. Give them the greatest amount of flexibility in preparing their code for those conditions by publishing your package with the agreement that building is an application level concern. Any decision that you can make in support of your users having an easier time when and if they decide to bundle is a good decision to make when preparing your package for publication. Every "do" and every "don't" that is included herein will help you in doing just that.
```

docs/publishing/index.md
```diff
@@ -0,0 +1,135 @@
...
+**Yes:** 👍
+
+```
+import { html } from 'lit-html';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+```
+
+**No:** 👎
+
+```
```
daKmoR, line 90:
```suggestion
```js
```

docs/publishing/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+When referring to 3rd party dependencies in your code, you cannot guarantee that implementing projects rely on the same file structure as you do. To prevent dependency specifiers from causing issues across different environments, be sure to import these dependencies into your project with "bare" import specifiers so that tooling in your users' environment can manage resolving the location of those dependencies; e.g. `import { LitElement } from 'lit-element';`
+
+## Do include file extensions in import specifiers
+
+When importing something that doesn't point directly to the entry point of a module, e.g.: `import {LitElement} from 'lit-element';`, always make sure to include a file extension: `import { ifDefined } from 'lit-html/directives/if-defined.js';`
+
+**Yes:** 👍
+
+```
```
daKmoR, line 83:
```suggestion
```js
```

docs/publishing/index.md
```diff
@@ -0,0 +1,135 @@
...
+section: guides
+tags:
+  - guides
+---
+
+## Publishing Web Components to NPM
+
+When you are ready to publish your package to NPM, be sure that you've addressed the recommendations below to ensure the code you publish is as easy for your users to consume as possible. Your package should already have [demos](/demoing/), [documentation](/demoing/storybook-addon-markdown-docs.html), [tests](/testing/), et al, as each of these plays a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time. With this reality in mind, they will not be discussed directly below. If you're looking for information on including these things in your published packages, take a look at our [guides](/guide/index.html) or learn about [getting started](#getting-started) below.
+
+<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically tuned for publishing to a CDN or other contexts where broader application-specific optimizations will not be applied to the code available therein.</p></div>
```
daKmoR, line 12:
```suggestion
<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically useful for publishing to a CDN or any other context where broader application-specific optimisations may be appropriated.</p></div>
```

thepassle: should be "may be appropriate"

docs/publishing/index.md
```diff
@@ -0,0 +1,135 @@
...
+**No:** 👎
+
+```
+import { html } from 'lit-html.js';
+import { ifDefined } from 'lit-html/directives/if-defined';
+```
+
+## Optional: use `type: "module"`
+
+Using `type: "module"` in your `package.json` will help CDNs and bundlers choose whether to parse your package with module parse goal. However, it will also enforce strict realities on how es modules and common JS modules must be imported into your scripting. If all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your scripting, then using `type: "module"` will allow those CDNs or bundlers to better leverage your package. If scripting in your package has issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
```
daKmoR, line 97:
```suggestion
Using `type: "module"` in your `package.json` will help CDNs and bundlers choose whether to parse your package with a module parse goal. However, it will also enforce that all the code you execute needs to be es modules. This includes all the dev tools (like bundlers, linters, ...) you use as well. CommonJs may still be used but requires a special way of importing. Therefore if all of the JS in your package is es modules, and using this entry in your `package.json` does not cause issues with your tools or scripts, then using `type: "module"` will allow those CDNs or bundlers to better leverage your package. If tools or script in your package have issues running in Node@12+ due to this inclusion, excluding it will ease the local requirements on import statements in your code.
```

docs/publishing/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+Bundling is an application-level concern. If you bundle, and then another dependency bundles and both dependencies bundle the same transitive dependencies it will become next to impossible for later tooling to deduplicate that code. Be sure NOT to bundle code in your published packages so that your users will have the easiest time possible optimizing their application for the specific use cases they target.
+
+## Do not minify
+
+Minification is an application-level concern. What's more, minifiers get better all the time, and you can ensure that your users can leverage the best in class approach to this space available by skipping the minifier when publishing your package.
+
+## Do not use `.mjs` file-extensions
+
+It is still all too common that a `.mjs` file gets served to a browser with an incorrect mime type and causes an application to fail unrecoverably. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please see [Optional: use `type: "module"`](#optional-use-type-module) above as a possible alternative.
```
daKmoR, line 121:
```suggestion
It is still all too common that a `.mjs` file gets served to a browser with an incorrect mime type and causes an application to fail unrecoverably. This happens when the used server does not have `.mjs` configured which could either be because it never became a default in the server software or an older stable version is used. Prevent this issue from ever arising by only publishing `.js` files. In cases where you are working with Node scripting in your package and you find issue with this recommendation, please see [Optional: use `type: "module"`](#optional-use-type-module) above as a possible alternative.
```

docs/publishing/index.md
```diff
@@ -0,0 +1,135 @@
...
+section: guides
+tags:
+  - guides
+---
+
+## Publishing Web Components to NPM
+
+When you are ready to publish your package to NPM, be sure that you've addressed the recommendations below to ensure the code you publish is as easy for your users to consume as possible. Your package should already have [demos](/demoing/), [documentation](/demoing/storybook-addon-markdown-docs.html), [tests](/testing/), et al, as each of these plays a role in making the clearest example of the benefits on your work, outlining the easiest path to integrating that work into other projects, and ensuring that the package as a whole is resilient to change over time. With this reality in mind, they will not be discussed directly below. If you're looking for information on including these things in your published packages, take a look at our [guides](/guide/index.html) or learn about [getting started](#getting-started) below.
+
+<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically useful for publishing to a CDN or any other context where broader application-specific optimisations may be appropriated.</p></div>
```
daKmoR, line 12:
```suggestion
<div class="custom-block warning"><p class="custom-block-title">WARNING</p> <p>These suggestions are not specifically useful for publishing to a CDN or any other context where broader application-specific optimisations may be appropriate.</p></div>
```

PR #1617, opened 2020-05-09
[RFC] docs: use base tag

Westbrook:
@daKmoR didn't catch it in the other PR but wanted to check in on your choice to use `responseTransformers` over a `base` tag. Thoughts?

### DIFF
--- a/docs/_includes/partials/service-worker.njk
+++ b/docs/_includes/partials/service-worker.njk
@@ -1,8 +1,9 @@
+{% if site.url !== '/_site-dev/' %}
 <script>
   if ('serviceWorker' in navigator) {
     window.addEventListener('load', function() {
       navigator.serviceWorker
-        .register(`${window.location.pathname.split('/').length - 1 > 1 ? '../service-worker.js' : './service-worker.js' }`)
+        .register('service-worker.js')
         .then(function() {
           console.log('ServiceWorker registered.');
         })
@@ -21,3 +22,4 @@
     }
   );
 </script>
+{% endif %}
--- a/docs/_data/site.js
+++ b/docs/_data/site.js
@@ -2,5 +2,5 @@ module.exports = {
   name: 'Open Web Components',
   shortDesc:
     'Open Web Components provides a set of defaults, recommendations and tools to help facilitate your web component project. Our recommendations include: developing, linting, testing, building, tooling, demoing, publishing and automating.',
-  url: process.env.DEPLOY_URL ? process.env.DEPLOY_URL : '/',
+  url: process.env.DEPLOY_URL ? process.env.DEPLOY_URL : '/_site-dev/',
 };
--- a/docs/_includes/home.njk
+++ b/docs/_includes/home.njk
@@ -3,7 +3,7 @@
 
   <head>
     {% include 'partials/meta.njk' %}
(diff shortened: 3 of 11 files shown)


docs/_includes/partials/service-worker.njk
```diff
@@ -1,8 +1,9 @@
+{% if site.url !== '/_site-dev/' %}
 <script>
   if ('serviceWorker' in navigator) {
     window.addEventListener('load', function() {
       navigator.serviceWorker
-        .register(`${window.location.pathname.split('/').length - 1 > 1 ? '../service-worker.js' : './service-worker.js' }`)
+        .register('service-worker.js')
```
thepassle, line 6: wont this break?

Westbrook: In what way? The PR build seems to work as expected:
https://5eb7085fc83fc00007895dc0--open-wc-org.netlify.app/index.html

image

You've got a lot more face time with service worker development, so I'll trust you if you think there's something amiss. In theory, the `<base>` tag points this to the right place all the time...

Westbrook:
It _does_ look like the base tag plays games with anchor links: https://5eb7085fc83fc00007895dc0--open-wc-org.netlify.app/faq/index.html Possible that that is a worthwhile annoyance to avoid...

docs/_includes/partials/service-worker.njk
Westbrook: Similarly, even if we didn't use a `<base>` tag, that _could_ be achieved by `/service-worker.js` all the time if we prevent the service worker while in development?

daKmoR:
we don't need a base tag in the live site... so imho it's better to only add it for dev...

I am curious about your reasons to "change" it? what would you gain?

docs/_includes/partials/service-worker.njk
thepassle: o.O if `open-wc.org/linting/linting-eslint.html` registers `service-worker.js` i would expect it to look for a service worker in `/linting/`, which is not there because its at the root `/`, right?

Would be nice if we dont need the window.location hack, im just surprised to see that's working and trying to understand why 😅

daKmoR: yes pls no `<base>` tag and using `/service-worker.js` 👍
we do not have a service worker for dev anyways (as we don't run rollup for this)

Westbrook:
Only in dev makes sense for the OpenWC setup, but I'm also taking a look at the current situation from two other perspectives:

1. Trying to get ramped up on `@mdrn/site` as a new user: what is the `responseTransformers` and why do I need it right out of the box? If we were to include it by default we'd need to really clearly document it. Feels like code smell in my new car, and for my experience `<base>` is what solves the underlying issue you're addressing, this PR is mainly to confirm the alternative approach so we can discuss it and figure out the _right_ pattern for a more public version of the tool.
2. I have to deploy the SWC site with a `<base>` tag anyways (root for local/PR/branch builds and `/spectrum-web-components/${site}` in our public URLs, so, while I likely have reflexive do what I'm already doing-itis, I do think we'll need to address this in some way for the published tool. Whether it's needed by default to address the peculiarities of how the tool works or is addressed as a light configuration, sites often need to be deployed to directories, so I was doing a little thinking ahead.

PR #1628, opened 2020-05-14
Fix/pin ts version

thepassle:
https://polymer.slack.com/archives/C03PF4L4L/p1589449263013300

TIL TS does breaking changes in minor versions


packages/es-dev-server/package.json
```diff
@@ -137,7 +137,7 @@
     "request": "^2.88.0",
     "selfsigned": "^1.10.4",
     "ts-mocha": "^7.0.0",
-    "typescript": "^3.8.3",
+    "typescript": "~3.8.3",
```
daKmoR, line 140: we have typescript as a "root" dependency and it should not be repeated in devDependencies...

```suggestion

```

PR #1631, opened 2020-05-14
docs: a11y updates

Westbrook:
Tweaks colors and `aria-labels` for accessibility.

Uses `#513bfe` which comes from the logo, but is a good bit more purple than what we'd used before @thepassle let me know if it offends your designer's sensibility...


thepassle:
🧐will take a look once preview deploy is done

thepassle:
could we do `#2758ff`? Its a bit less purpl-ey, and still passes color contrast

`#513bfe`:
Screenshot 2020-05-14 at 14 56 17

`#2758ff`:
Screenshot 2020-05-14 at 14 56 13


image

Westbrook:
Updated!

PR #1635, opened 2020-05-14
Fix/storybook resolve

LarsDenBakker:


packages/demoing-storybook/src/start/transformers/mdxToCsfTransformer.js
```diff
@@ -10,7 +10,7 @@ async function mdxToCsfTransformer({ url, body, status }) {
 
     const newBody = (await transformMdxToCsf(body, path)).code;
     if (newBody) {
-      return { body: newBody, contentType: 'text/javascript' };
```
daKmoR, line 13: that change was not "mandatory" right? 🤔

LarsDenBakker: It should have been application/javascript, but then i might as well leave it out since it's redundant

daKmoR: yeah so the default contentType is correct here 👍

PR #1641, opened 2020-05-15
Fix/pin ts version

thepassle:
my crappy git-fu messed up the previous PR somehow 🤷

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "rollup": "^2.7.2",
     "rollup-plugin-copy": "^3.3.0",
     "sinon": "^7.4.1",
-    "typescript": "^3.8.3",
+    "typescript": "~3.8.3",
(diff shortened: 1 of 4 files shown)


packages/es-dev-server/package.json
```diff
@@ -138,7 +138,6 @@
     "request": "^2.88.0",
     "selfsigned": "^1.10.4",
     "ts-mocha": "^7.0.0",
-    "typescript": "^3.8.3",
```
Westbrook, line 141: I thought EDS was written in TS now...is that another line of work?

thepassle: @daKmoR said its not needed because its in the main package.json

LarsDenBakker: 🤷

PR #1647, opened 2020-05-16
[WIP] ci: post npm init @open-wc output to github

Westbrook:
Testing, testing, but...

Let's make more things automatic!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,29 @@ jobs:
       - run:
           name: Publish package
           command: "./node_modules/.bin/lerna publish --message 'chore: release new versions' --yes"
+  generate-repo:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/
+      - add_ssh_keys:
+          fingerprints:
+            - '95:74:63:db:82:da:b3:e6:57:0c:3a:bd:3f:ff:ce:c6'
+      - run:
+          name: Authenticate with registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+      - run: git config --global user.email circleci@circleci
+      - run: git config --global user.name CircleCI
+      - run: 
+          name: Publish init output to GitHub
+          command: |
+            cd packages/create
+            git clone git@github.com:open-wc/init-output-test.git
+            yarn start --type scaffold --scaffoldType wc --features testing demoing building demoing --tagName init-output-test --scaffoldFilesFor testing demoing building --installDependencies false --writeToDisk true --typescript no --overwriteFile always
+            cd init-output-test
+            git add .
+            git commit -am 'updated npm init @open-wc output' && git push -f || exit 0
 
 workflows:
   version: 2
@@ -93,6 +116,12 @@ workflows:
       - test-bs:
           requires:
             - build
+      - generate-repo:
+          requires:
+            - build
+          filters:
+            branches:
+              only: westbrook/generate-repo
       - deploy:
           requires:
             - test-local
--- a/packages/create/src/generators/app/index.js
+++ b/packages/create/src/generators/app/index.js
@@ -14,6 +14,7 @@ import { gatherMixins } from './gatherMixins.js';
  * example:
  * npm init @open-wc --type scaffold --scaffoldType app --tagName foo-bar --installDependencies false
  * npm init @open-wc --type upgrade --features linting demoing --tagName foo-bar --scaffoldFilesFor demoing --installDependencies false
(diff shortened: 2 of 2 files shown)


Westbrook:
Now publishes the JS version to https://github.com/open-wc/init-output-test

Which versions do we want to publish?
- JS WC All
- TS WC All
- JS APP All
- TS APP All
- Others?

Do we have existing repos we should use? New ones?

We should settle on a "machine user" or a more capable SSH key to use so that we're not relying on User Keys for this.

.circleci/config.yml
```diff
@@ -81,6 +81,30 @@ jobs:
...
           command: "./node_modules/.bin/lerna publish --message 'chore: release new versions' --yes"
+  generate-repo:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/
+      - add_ssh_keys:
+          fingerprints:
+            - '95:74:63:db:82:da:b3:e6:57:0c:3a:bd:3f:ff:ce:c6'
```
Westbrook, line 92: This is my user key...we should probably have a Machine Key for the team.

.circleci/config.yml
```diff
@@ -81,6 +81,30 @@ jobs:
...
+          name: Authenticate with registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+      - run: git config --global user.email circleci@circleci
+      - run: git config --global user.name CircleCI
+      - run: 
+          name: Publish init output to GitHub
+          command: |
+            cd packages/create
+            git clone git@github.com:open-wc/init-output-test.git
+            yarn start --type scaffold --scaffoldType wc --features testing demoing building demoing --tagName init-output-test --scaffoldFilesFor testing demoing building --installDependencies false --writeToDisk true --typescript no --overwriteFile always
```
Westbrook, line 103: This ensures we use the _latest_ generator at all times. Which is needed in time like now when we're adding CLI properties like `--overwriteFile`.

.circleci/config.yml
```diff
@@ -93,6 +117,12 @@ workflows:
       - test-bs:
           requires:
             - build
+      - generate-repo:
+          requires:
+            - build
+          filters:
+            branches:
+              only: westbrook/generate-repo
```
Westbrook, line 125: We'd want this just on `master`, of course, but is there any other gating we'd want to use to reduce incidences of this?

.circleci/config.yml
```diff
@@ -81,6 +81,30 @@ jobs:
...
+      - run: 
+          name: Publish init output to GitHub
+          command: |
+            cd packages/create
+            git clone git@github.com:open-wc/init-output-test.git
+            yarn start --type scaffold --scaffoldType wc --features testing demoing building demoing --tagName init-output-test --scaffoldFilesFor testing demoing building --installDependencies false --writeToDisk true --typescript no --overwriteFile always
+            cd init-output-test
+            git add .
+            git commit -am 'updated npm init @open-wc output'
+            git push -f
```
Westbrook, line 107: This might "fail" in some cases when there are no changes... do we want to let it fail or do some more in-depth management here?

LarsDenBakker:
The variants you mention should be fine, it's already quite a bit :)

.circleci/config.yml
LarsDenBakker: Does it mark the CI as failed? That would be a shame.

.circleci/config.yml
LarsDenBakker: Is it possible to do it only when `create` has a version bump?

.circleci/config.yml
LarsDenBakker: Yeah.. I have no idea how this works though :(

Westbrook: IIUC, we just create a new GitHub account `open-wc-machine-user` (or better, damn we've got to name another thing), and then give it specific access/an SSH key and then Bob's you're uncle...

.circleci/config.yml
Westbrook: This is a good question that I'm not 100% sure about. It might be possible to do this with Lerna/Convential Commits... @daKmoR any experience with this? It might be possible to leverage `postpublish`?

.circleci/config.yml
Westbrook: We could make it an "optional" step in GitHub so that it didn't fail the build...but it does go red in Circle... let me see what a `|| exit 0` does!

Westbrook: Oh, that did it!

.circleci/config.yml
Westbrook: Also, if the output is the same, it won't be able to commit anything, so nothing gets pushed: https://app.circleci.com/pipelines/github/open-wc/open-wc/6152/workflows/b7f8bba8-86ec-47d8-bfa7-60cbb469835f/jobs/9037/parallel-runs/0/steps/0-107

PR #1656, opened 2020-05-18
fix(create): generate sourcemap for ts files

vikerman:
Generates a separate .js.map file in out-tsc/ that has the original TS source inlined in it.

This should let us see the original TS source while debugging.

Browser is able to load the sourcemap when used in development mode with es-dev-server.
Currently not being processed by rollup in production mode and the behavior is same as earlier.

### DIFF
--- a/packages/create/src/generators/app-lit-element-ts/templates/_tsconfig.json
+++ b/packages/create/src/generators/app-lit-element-ts/templates/_tsconfig.json
@@ -10,7 +10,9 @@
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
     "importHelpers": true,
-    "outDir": "out-tsc"
+    "outDir": "out-tsc",
+    "sourceMap": true,
+    "inlineSources": true
   },
   "include": ["**/*.ts"]
 }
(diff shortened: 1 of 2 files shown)


thepassle:
Nice!

Could you add this to the web component starter as well? Then we can keep it one PR 🙂https://github.com/open-wc/open-wc/blob/master/packages/create/src/generators/wc-lit-element-ts/templates/_tsconfig.json

vikerman:
Just wanted to check before I did that because that could change the dist/ folder. Do we want people to ship their sourcemaps in npm? (A very quick search on the web didn't yield any reason not to - especially because the sources are inlined in the map and there is not need to ship the TS files)

thepassle:
> Just wanted to check before I did that because that could change the dist/ folder.

Good point. We currently have that as an optional recommendation in https://open-wc.org/publishing/ , so I think that should be fine

Maybe would be good to get a 2nd opinion from @LarsDenBakker here just to be sure

LarsDenBakker:
Yep I think we should. lit-html and lit-element do this too.

For apps we don't recommend it, because it will show potentially unwanted comments to your users.

vikerman:
Made the changes. Please take a look.

thepassle:
Thanks @vikerman !

esnho:
@thepassle is this fix available in a recent release?

thepassle:
This should be available in the current generator, yeah

PR #1657, opened 2020-05-19
11tyDevServer and Rollup Build

daKmoR:
- have a "start cli" which runs es-dev-server and executes 11ty on the demand
- have a "build cli" which runs 11ty and "pipes" the output into rollup

you can test it with
```
yarn docs:start
yarn docs:build
```

Open Todos:
- createMpaConfig is now basically the same as createSpaConfig 🤔  => still needs to move into @open-wc/building-rollup

Optionally for later: (team rocket 😬)
- caching in "start-cli" which get's fully invalidated on any file change
- how to handle configurations (separate config files? one config file? how to merge? how to provide, ...)


Westbrook:
Do we have an issue submitted to 11ty about that patch we're making here? Or is this something they're already planning to do? Want to make sure we've started the conversation of getting the patch here removed from our project at some point.

daKmoR:
jup it's here
https://github.com/11ty/eleventy/issues/789

Issue #1637, opened 2020-05-15
[testing-helpers] "scoped-elements-test-wrapper" is already defined in monorepo


MariusVatasoiu:
I have a monorepo and every package uses @open-wc/testing (implicitly @open-wc/testing-helpers).

When I run karma on all packages, I get the following error:

```
HeadlessChrome 81.0.4044 (Mac OS X 10.14.6) ERROR
  Uncaught 
  at http://localhost:9876/context.html:0:0
  
  Error: Failed to execute 'define' on 'CustomElementRegistry': the name "scoped-elements-test-wrapper" has already been used with this registry
      at packages/ui/node_modules/@open-wc/testing-helpers/src/scopedElementsWrapper.js:65:18
```

The problem is that scopedElementsWrapper.js doesn't check if `scoped-elements-test-wrapper` is already defined.

LarsDenBakker:
You are probably loading multiple instances of the testing package, this can create issues on multiple levels. I recommend looking into making sure there is only one version of the testing package installed, depending on the package manager you may need to configure the symlinks setting to ensure things are loaded just once. Otherwise you can run a separate karma instance per package.

MariusVatasoiu:
I solved this problem by hoisting the packages.

Is there any problem if there would be a simple condition to check if that custom element is already defined?

LarsDenBakker:
There is no way of knowing if the two instances of the library are compatible, so a simple check can create subtle bugs.

@manolakis could we auto generate the name here? or do we need to rely on a static name?

MariusVatasoiu:
I've tested two options.
First, was to simply check is the custom element exists, but you're right, there could be different versions and that's a problem.

Second, was to register a new custom element if `scoped-elements-test-wrapper` already exists. I was checking for custom elements with an index at the end, like scoped-elements-test-wrapper-1, until I find an unregistered one.

LarsDenBakker:
Yep that would probably work in this case

manolakis:
Yes, I think the best option is auto-generate the name. I'm on it!

manolakis:
Although auto-generating the name to avoid naming collision is possible, hoisting the package or using Yarn should fix the problem of many definitions, and IMHO is a better practice to have one version of testing-helpers instead of many, don't you think?

MariusVatasoiu:
I wouldn't consider auto-generating the name as a replacement for better practices, but rather as a safe solution in case you want to do something different.

[...]

PR #1658, opened 2020-05-19
fix(testing-helpers): auto-generate scoped elements test wrapper name

manolakis:
Loading multiple instances of the testing package could cause a naming collision in the ScopedElementsTestWrapper definition. Auto-generating the tag name and checking that it is not used in the global registry before definition solves the problem.

Fix: #1637

### DIFF
--- a/packages/testing-helpers/src/scopedElementsWrapper.js
+++ b/packages/testing-helpers/src/scopedElementsWrapper.js
@@ -48,7 +48,24 @@ class ScopedElementsTestWrapper extends ScopedElementsMixin(LitElement) {
   }
 }
 
-customElements.define('scoped-elements-test-wrapper', ScopedElementsTestWrapper);
+/**
+ * Obtains a unique tag name for the test wrapper
+ * @param {number} [counter=0]
+ * @returns {string}
+ */
+const getWrapperUniqueName = (counter = 0) => {
+  const tag = `scoped-elements-test-wrapper-${counter}`;
+
+  if (customElements.get(tag) !== undefined) {
+    return getWrapperUniqueName(counter + 1);
+  }
+
+  return tag;
+};
+
+const wrapperTagName = getWrapperUniqueName();
+
+customElements.define(wrapperTagName, ScopedElementsTestWrapper);
 
 /**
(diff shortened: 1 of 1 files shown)


packages/testing-helpers/src/scopedElementsWrapper.js
```diff
@@ -58,10 +75,12 @@ customElements.define('scoped-elements-test-wrapper', ScopedElementsTestWrapper)
...
-      .template="${template}"
-    ></scoped-elements-test-wrapper>
-  `;
+  const strings = [
+    `<${wrapperTagName} .scopedElements="`,
+    '" .template="',
+    `"></${wrapperTagName}>`,
+  ];
+
+  // @ts-ignore
```
manolakis, line 84: `@ts-ignore` is required because `html` expects a `TemplateStringsArray` instead of a `string[]`

manolakis:
I'm going to close the PR. Hoisting the package or using Yarn should fix the problem of many definitions. It's better to have one version of testing-helpers instead of many.

LarsDenBakker:
@manolakis I think is important to have. Even though it's better to deduplicate.hoist packages, it may not always be feasible. Since this is the only other singleton part of the package, it would be nice if the package it fully autonomous for maximum flexibility. Of course we should offset that against implementation complexity, but it doesn't seem like a costly implementation here?

manolakis:
It makes sense to me.

@daKmoR ?

daKmoR:
I still think it's only fixing a symptom... 


but I don't mind as long as I don't need to do it 😬

manolakis:
I see both points of view. I agree is better to have only one version of testing-helpers but, if for some reason is not possible, we shouldn't break users test, don't you think? It's not production code...

Should we add some warning if we detect different versions guiding the users to hoist the package or align the package version instead hiding it? 🤷‍♂️

LarsDenBakker:
I don't think it's the role of a testing library to teach people about version management :D I'd say we can merge this as is. What do you think @daKmoR ?

manolakis:
IMHO "scoping" the test wrapper doesn't hurt.

LarsDenBakker:
Indeed!

PR #1663, opened 2020-05-20
feat: allow disabling cache for transform plugin response body

jorenbroekema:
Context:

I moved the response body cache out of middlewares and made it its own function. I abstracted adding to cache and serving from cache to its own functions that the transform plugin can use. I'm not sure if you like the way I approached that but it was a little hard for me to find a better way to move the logic in a way where we can still do something with it inside a transform plugin.


packages/es-dev-server/test/middleware/plugin-transform.test.ts
```diff
@@ -128,12 +128,55 @@ describe('plugin-transform middleware', () => {
...
+    } finally {
+      server.close();
+    }
+  });
+
+  it('allows users to turn off caching of response body', async () => {
+    const { server } = await startServer({
+      transform(ctx) {
+        if (ctx.path === '/app.js') {
+          return { body: `${Date.now()}`, disableCache: true };
```
LarsDenBakker, line 165: Let's do `transformCache: false` to highlight what kind of cache it is.

jorenbroekema: Sounds good :) I changed it

packages/es-dev-server/src/create-middlewares.ts
```diff
@@ -1,28 +1,29 @@
-import koaStatic from 'koa-static';
-import koaEtag from 'koa-etag';
-import koaCompress from 'koa-compress';
 import chokidar from 'chokidar';
```
LarsDenBakker, line 1: What happened to this file? Alphabetically reordered?

packages/es-dev-server/src/middleware/plugin-transform.ts
```diff
@@ -18,12 +52,27 @@ export function createPluginTransformMiddlware(cfg: PluginServeMiddlewareConfig)
...
+    }
+
     await next();
 
     if (context.status < 200 || context.status >= 300) {
       return undefined;
     }
 
+    if (cached) {
+      context.body = cachedBody;
```
LarsDenBakker, line 72: If we're serving from the cache we can avoid running the other middleware. We can just serve it directly, so we can place this above the `await next()`

packages/es-dev-server/src/middleware/plugin-transform.ts
```diff
@@ -53,5 +104,15 @@ export function createPluginTransformMiddlware(cfg: PluginServeMiddlewareConfig)
         }
       }
     }
+
+    if (cfg.fileWatcher && transformCache && cfg.cache && cfg.cacheKeysForFilePaths) {
```
LarsDenBakker, line 108: If these are not optional parameters, we don't need to check if they're there.

packages/es-dev-server/src/middleware/plugin-transform.ts
```diff
@@ -18,12 +52,27 @@ export function createPluginTransformMiddlware(cfg: PluginServeMiddlewareConfig)
...
 
   return async function pluginTransformMiddleware(context, next) {
+    let cached = false;
+    let cacheKey = '';
+    let cachedBody: string | undefined;
+
+    if (cfg.fileWatcher && cfg.cache) {
+      const result = await tryServeFromCache(cfg.cache, context);
+      ({ cached, cacheKey, cachedBody } = result);
+      context = result.context;
```
LarsDenBakker, line 62: What does reassigning the context do? Shouldn't this be like the old logic:

```
        ctx.body = cached.body;
        ctx.response.set(cached.headers);
        ctx.status = 200;
```

packages/es-dev-server/src/create-middlewares.ts
```diff
@@ -129,6 +137,12 @@ export function createMiddlewares(
     );
   }
 
+  let cache: LRUCache<string, CacheEntry> | undefined;
```
LarsDenBakker, line 140: I think we can move these inside `createPluginTransformMiddlware` since that's the only consumer.

packages/es-dev-server/src/create-middlewares.ts
jorenbroekema: Yep, organizeImports VS Code setting. I had it turned on when working on this PR, one day I might be able to make this a build-thing where the CI will include organizing imports in the formatting, but for now it is sadly dependent on .. the user's editor :<

packages/es-dev-server/src/middleware/plugin-transform.ts
jorenbroekema: Yep you're right, and that's what it's doing in tryServeFromCache I believe. I'm not sure why I am doing this here

packages/es-dev-server/src/middleware/plugin-transform.ts
jorenbroekema: Yep you're right, done :)!

packages/es-dev-server/src/middleware/plugin-transform.ts
jorenbroekema: I removed it and it works :)

packages/es-dev-server/src/middleware/plugin-transform.ts
jorenbroekema: I removed the cache and cacheKeysForFilePaths check, I think this was added as a lack of typescript experience, I had some type errors because I defined them as `<thing> | undefined`

packages/es-dev-server/src/create-middlewares.ts
jorenbroekema: Yep :) true, not really sure why I did it but I moved them to `createPluginTransformMiddleware` now and it made the code cleaner for sure

PR #1670, opened 2020-05-23
feat: allow to set plugins for mdjs processing

daKmoR:


packages/demoing-storybook/demo/.storybook/main.js
```diff
@@ -7,6 +7,8 @@ module.exports = {
     'storybook-prebuilt/addon-docs/register.js',
     'storybook-prebuilt/addon-knobs/register.js',
   ],
+  // this would disable the ids of headlines - you can also use it to add your own unified/remark plugins
+  // setupMdjsPlugins: plugins => plugins.filter(plugin => plugin.name !== 'mdSlug'),
```
LarsDenBakker, line 11: Maybe just call this `mdjsPlugins`? Then it follows the convention of rollup and EDS option names.

daKmoR: I had that first as it was only an array when I started... but it needs to be a function (as it get's passed on also the filePath)
so to me `mdjsPlugins` feel like it would accept an array 😅

PR #1671, opened 2020-05-23
feat(rollup-plugin-html): support multiple html inputs

daKmoR:
This now allows configs which can be used for multiple html files.

Also multiple "virtual html files" as strings are possible 👍

example for multiple Html Files:
```js
const html = require('@open-wc/rollup-plugin-html');

module.exports = {
  input: '**/*.html',
  output: {
    dir: './demo/dist',
  },
  plugins: [html({ flatten: false, rootDir: './demo/mpa' })],
};
```

it is a rather big change 😅


packages/rollup-plugin-html/README.md
```diff
@@ -73,7 +74,61 @@ export default {
...
+  input: ['index.html', 'pages/*.html'],
+  output: { dir: 'dist' },
+  plugins: [html({ flatten: false })],
+};
+```
+
+</details>
+
+When keeping the folder structures it may mean that your `rootDir` is not your current working directory.
+Image for example all the html files are generated in to a `_site-in-html` folder.
```
LarsDenBakker, line 101: I think it's better to explain this option in detail at the bottom, and to show just a code example here.

packages/rollup-plugin-html/README.md
```diff
@@ -452,19 +505,31 @@ All configuration options are optional if an option is not set the plugin will f
...
-Path to the HTML file to use as input. Modules in this file are bundled and the HTML is used as the template for the generated HTML file.
+Paths to the HTML file to use as input. Modules in this files are bundled and the HTML is used as the template for the generated HTML file. Does support glob patterns.
 
-### inputHtml
+### rootDir
 
 Type: `string`
 
-Same as `inputPath`, but provides the HTML as a string directly.
+Path where the `files` should start look.
```
LarsDenBakker, line 520: `rootDir` used to be only for HTML files which are provided as strings, in order to indicate relative to where relative imports should be resolved. From what I can tell it's now extended to also be where to start globbing from, and which part to strip when not flattening the filename. We should explain that here.

packages/rollup-plugin-html/README.md
```diff
@@ -452,19 +505,31 @@ All configuration options are optional if an option is not set the plugin will f
...
+
+### flatten
+
+Type: `boolean`
+
+Should folder in filePaths be stripped and all files be places in a single folder (Default to true)
+
+### html
+
+Type: `string|[{ name: string, html: string, rootDir: string}]`
```
LarsDenBakker, line 530: I think we can remove `rootDir` here now, and use the top level option.

packages/rollup-plugin-html/rollup-plugin-html.js
```diff
@@ -31,26 +32,24 @@ const defaultFileName = 'index.html';
...
   /** @type {TransformFunction[]} */
-  let externalTransformFns = [];
+  let thisExternalTransformFns = [];
+  /** @type {HtmlFile[]}  */
+  const thisHtmlFiles = [];
 
   // variables for multi build
   /** @type {string[]} */
-  const multiOutputNames = [];
+  const thisMultiOutputNames = [];
```
LarsDenBakker, line 52: I don't like the `this` but you probably did this on purpose 😆

packages/rollup-plugin-html/README.md
```diff
@@ -452,19 +505,31 @@ All configuration options are optional if an option is not set the plugin will f
...
+Name of the generated HTML file. If `files` is set, defaults to the `files` filename, otherwise defaults to `index.html`.
 
-### inputPath
+### files
 
-Type: `string`
+Type: `string|strings[]`
 
-Path to the HTML file to use as input. Modules in this file are bundled and the HTML is used as the template for the generated HTML file.
+Paths to the HTML file to use as input. Modules in this files are bundled and the HTML is used as the template for the generated HTML file. Does support glob patterns.
```
thepassle, line 514:
```suggestion
Paths to the HTML file to use as input. Modules in this files are bundled and the HTML is used as the template for the generated HTML file. This may also be a glob pattern.
```

packages/rollup-plugin-html/README.md
thepassle, line 520: path where the files should start look? Can you explain this more?

packages/rollup-plugin-html/rollup-plugin-html.js
```diff
@@ -159,23 +191,57 @@ function rollupPluginHtml(pluginOptions) {
...
+          this.emitFile(asset);
+        }
+      } else {
+        const htmlFileName = pluginOptions.name || defaultFileName;
+        this.emitFile(await createHtmlAsset(options.dir, { htmlFileName }));
+      }
+      // this.emitFile(await createHtmlAsset(options.dir));
+
+      // const htmlFileName = pluginOptions.name || defaultFileName;
+      // this.emitFile(await createHtmlAsset(options.dir, { htmlFileName }));
```
LarsDenBakker, line 212: some comments left here

packages/rollup-plugin-html/README.md
```diff
@@ -452,19 +505,31 @@ All configuration options are optional if an option is not set the plugin will f
...
 Type: `string`
 
-Same as `inputPath`, but provides the HTML as a string directly.
+Path where the `files` should start look.
+
+### flatten
+
+Type: `boolean`
+
+Should folder in filePaths be stripped and all files be places in a single folder (Default to true)
```
thepassle, line 526:
```suggestion
Whether or not the folder in `filePaths` should be stripped, and all files should be placed in a single folder. (Defaults to true)
```

packages/rollup-plugin-html/src/getInputHtmlData.js
```diff
@@ -1,39 +1,107 @@
...
+ * Will return all matching file paths fully resolved.
+ * @param {string} fromGlob
+ * @param {string} rootDir
+ * @return {string[]}
+ */
+function listFiles(fromGlob, rootDir) {
+  const files = glob.sync(fromGlob, { cwd: rootDir, absolute: true });
+  return (
+    files
+      // .map(filePath => path.resolve(rootDir, filePath))
```
LarsDenBakker, line 22: Comment left here

packages/rollup-plugin-html/rollup-plugin-html.js
```diff
@@ -129,28 +155,34 @@ function rollupPluginHtml(pluginOptions) {
      * Watches input HTML for file reloads.
      */
     buildStart() {
-      generatedBundles = [];
-      externalTransformFns = [];
+      thisGeneratedBundles = [];
+      thisExternalTransformFns = [];
```
thepassle, line 159: this generated bundles? Do you mean current?

packages/rollup-plugin-html/README.md
```diff
@@ -13,6 +13,7 @@ Plugin for generating HTML files from rollup.
 - Generate one or more HTML pages from a rollup build
 - Inject rollup bundle into an HTML page
 - Optionally use HTML as rollup input, bundling any module scripts inside
+- Optionally use multiple html files (via a glob) or html strings as input
```
thepassle, line 16:
```suggestion
- Optionally use multiple html files via a glob or html strings as input
```

packages/rollup-plugin-html/src/getInputHtmlData.js
```diff
@@ -1,39 +1,107 @@
...
+/**
+ * @param {string} filePath
+ * @param {string} rootDir
+ * @param {boolean} flatten
+ */
+function getName(filePath, rootDir, flatten = true) {
+  if (flatten) {
+    return path.basename(filePath);
+  }
+  return filePath.substring(rootDir.length + 1);
```
LarsDenBakker, line 52: I think path.relative is safer here cross platform

packages/rollup-plugin-html/rollup-plugin-html.js
daKmoR: yeah I had a tough time to understand which variables are local function variabels and which are "plugin" variables...  `this...` seemed better then `global...` 😬

of course open for better suggestions as long as there is distinction

packages/rollup-plugin-html/README.md
daKmoR: this "rootDir" is the one from the html file 😅
and is different then the "global" rootDir... probably a good idea to rename it..

pageRootDir and htmlRootDir or rootDir?

packages/rollup-plugin-html/rollup-plugin-html.js
LarsDenBakker: For variables distinct from function parameter, I don't see the need. 

For similar we can do something like "all" or "gathered". If the plugin is becoming too large, we can also place more logic in separate functions. `createHtmlAsset` is a good candidate for that, and the child plugin logic.

packages/rollup-plugin-html/README.md
LarsDenBakker: We can probably reuse the same variable for both

packages/rollup-plugin-html/README.md
daKmoR: moved it

packages/rollup-plugin-html/README.md
```diff
@@ -444,6 +465,39 @@ export default {
...
+.
+├── _site-in-html
+│   ├── page-a
+│   │   └── index.html
+│   └── index.html
+├── rollup.config.js
+└── package.json
+```
+
+If you now provide `input: '_site-in-html/**/*.html';` then it will keep this folder which will result in this url `https://my-domain.com/_site_in-html/`.
```
thepassle, line 483: this paragraph is not entirely clear to me yet, is there a better way we can describe it? Theres also some typos, maybe we can run it through grammarly?

packages/rollup-plugin-html/README.md
```diff
@@ -73,7 +74,29 @@ export default {
...
+  input: ['index.html', 'pages/*.html'],
+  output: { dir: 'dist' },
+  plugins: [html({ flatten: false, rootDir: '_site' })],
+};
+```
+
+</details>
+
+You can also set the `files` property on the html plugin. This will take precedence over rollups input.
+Additionally it does support an array of files/globs and can be use if you do want to use multiple html plugins with different options:
```
thepassle, line 99:
```suggestion
Additionally it supports an array of files/globs and can be used if you want to use multiple html plugins with different options:
```

packages/rollup-plugin-html/README.md
```diff
@@ -452,19 +506,31 @@ All configuration options are optional if an option is not set the plugin will f
...
-Path to the HTML file to use as input. Modules in this file are bundled and the HTML is used as the template for the generated HTML file.
+Paths to the HTML file to use as input. Modules in this files are bundled and the HTML is used as the template for the generated HTML file. This may also be a glob pattern.
 
-### inputHtml
+### rootDir
 
 Type: `string`
 
-Same as `inputPath`, but provides the HTML as a string directly.
+Path to a directory which will serve as the starting point for all `files`. Also the final file tree will result in relative urls to this rootDir.
```
thepassle, line 521:
```suggestion
Path to a directory which will serve as the starting point for all `files`. The final file tree will result in relative urls to this rootDir.
```

packages/rollup-plugin-html/README.md
```diff
@@ -452,19 +506,31 @@ All configuration options are optional if an option is not set the plugin will f
...
+
+Type: `boolean`
+
+Whether or not the folder in `filePaths` should be stripped, and all files should be placed in a single folder. (Defaults to true)
+
+### html
+
+Type: `string|[{ name: string, html: string, rootDir: string}]`
+
+Provide the HTML directly as string. If multiple files are provided then name, html and rootDir are required.
```
thepassle, line 533:
```suggestion
Provide the HTML directly as string. If multiple files are provided then `name`, `html` and `rootDir` are required.
```

packages/rollup-plugin-html/src/getInputHtmlData.js
daKmoR: 👍

packages/rollup-plugin-html/README.md
daKmoR: ok will keep the same name 👍

packages/rollup-plugin-html/rollup-plugin-html.js
daKmoR: removed

packages/rollup-plugin-html/README.md
daKmoR: grammarly and more infoe

packages/rollup-plugin-html/README.md
```diff
@@ -444,6 +465,39 @@ export default {
 
 </details>
 
+## Details for Plugin rootDir
+
+When keeping the folder structures it may mean that your `rootDir` is not your current working directory.
+Image for example all the html files are generated into a `_site-in-html` folder.
```
thepassle, line 471:
```suggestion
For example, imagine all the html files are generated into a `_site-in-html` folder.
```

packages/rollup-plugin-html/README.md
```diff
@@ -444,6 +465,39 @@ export default {
...
+├── _site-in-html
+│   ├── page-a
+│   │   └── index.html
+│   └── index.html
+├── rollup.config.js
+└── package.json
+```
+
+If you provide `input: '_site-in-html/**/*.html';` it will result in files like `dist/_site-in-html/index.html`. If the dist folder gets automatically uploaded to a static hosting service then it will result in this url `https://my-domain.com/_site_in-html/index.html`.
+By defining `rootDir: './_site-in-html'` & `input: '**/*.html';` we can get do get files like `dist/index.html` and urls like `https://my-domain.com/`.
```
thepassle, line 484:
```suggestion
By defining `rootDir: './_site-in-html'` and `input: '**/*.html';` we can get files like `dist/index.html` and urls like `https://my-domain.com/`.
```

LarsDenBakker:
Tested it locally, looks good! Two open points (also in the comments above): 

Change the `thisVariable`s back to `variable`s, we should just split up the file more to make it easier to follow.

I think we could have only one rootDir on the top level, and imports in virtual files are resolved relative to that and the file name. So in the demo this would be the config:

```
module.exports = {
  output: {
    dir: './demo/dist',
  },
  plugins: [
    html({
      rootDir: __dirname,
      html: [
        // ..
        { name: 'index.html', html: htmlIndex },
        { name: 'pages/page-a.html', html: htmlPageA },
        { name: 'pages/page-b.html', html: htmlPageB },
        { name: 'pages/page-c.html', html: htmlPageC },
      ],
    }),
  ],
};
```

Relative imports would still work, because `./foo.js` in `index.html` would look for __dirname + `./foo.js`, while in `pages/page-a.html` it would look for dirname + `pages/foo.js`.

If you don't have time to do this in this PR now, we could do this as a separate change.

daKmoR:
> Change the thisVariables back to variables, we should just split up the file more to make it easier to follow.

move `createHtmlAssets` functions out and renamed variables back 🤗

> I think we could have only one rootDir on the top level, and imports in virtual files are resolved relative to that and the file name. So in the demo this would be the config:

auto calculate rootDir base on name 👍

abdonrd:
Now you will use it in the docs instead of this, right?

https://github.com/open-wc/open-wc/blob/master/docs/_building-rollup/createMpaConfig.js

daKmoR:
@abdonrd indeed something like this is going to be the next step - we still need to check if an createMpaConfig brings value

Issue #1672, opened 2020-05-23
[@open-wc/rollup-plugin-html] breaks if there is no module in the html file


daKmoR:
Given the following html and rollup config I do get the error

```html
<html>
<body>
  <h1>hey</h1>
</body>
</html>
```

```js
const html = require('@open-wc/rollup-plugin-html');

module.exports = {
  output: {
    input: 'input.html',
    dir: './demo/dist',
  },
  plugins: [
    html(),
  ],
};
```

```
rollup v2.8.0
bundles index.html → demo/dist...
[!] Error: You must supply options.input to rollup
Error: You must supply options.input to rollup
   at Graph.parseModules (open-wc/node_modules/rollup/dist/shared/rollup.js:18426:19)
   at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Expected outcome

- do not fail
- optimize html
- create file in dist

👉  `./demo/dist/input.html` 
```html
<html><body><h1>hey</h1></body></html>
```

thepassle:
What would the expected behavior here be? If there is no module in the index.html, there's no JS to pass to rollup

daKmoR:
it should still optimize the html and create the file in the dist folder...

PR #1673, opened 2020-05-23
fix(rollup-plugin-html): allow pure html input

daKmoR:
closes https://github.com/open-wc/open-wc/issues/1672

Requires https://github.com/open-wc/open-wc/pull/1671 to be merged before 😅  or we merge into it...


LarsDenBakker:
Nice, looks like no resulting module is injected either? 

Could we make the names a bit less "funny" 😆

daKmoR:
open for suggestions 😅

daKmoR:
renamed to `fake-module-for-pure-html-input.js`

PR #1677, opened 2020-05-24
fix(@open-wc/webpack-import-meta-loader): fix a bug in the publicPath…

FredKSchott:
```js
// Currently
console.log(import.meta.url); // -> http://localhost:8080//_dist_/index.js

// Expected
console.log(import.meta.url); // -> http://localhost:8080/_dist_/index.js
```

I had thought that this might be a bug in my usage but based on the logic I'm not sure this ever worked without __webpack_public_path__ being set.

I'd love to use this as a part of Snowpack's Webpack bundler plugin if we can fix.

### DIFF
--- a/packages/webpack-import-meta-loader/webpack-import-meta-loader.js
+++ b/packages/webpack-import-meta-loader/webpack-import-meta-loader.js
@@ -45,7 +45,7 @@ module.exports = function (source) {
         if (publicPath) {
           url += publicPath;
         } else {
-          url += '//';
+          url += '/';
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
👍  Thanks! We probably didn't notice, we've been using https://www.npmjs.com/package/babel-plugin-bundled-import-meta instead of this ourselves.

LarsDenBakker:
Will look into the ci failures, they are not related to the PR.

PR #1686, opened 2020-05-27
docs: update events.md

bschlenk:
Clarify that `.bind` is not necessary to call when attaching an event via `addEventListener` to the element

### DIFF
--- a/docs/faq/events.md
+++ b/docs/faq/events.md
@@ -67,7 +67,7 @@ If you are adding a listener to your element, from inside your element, you shou
 ```js
 constructor() {
     super();
-    this.addEventListener('done', this.handleDone.bind(this));
+    this.addEventListener('done', this.handleDone);
 }
 handleDone(event) {
     this.done = true;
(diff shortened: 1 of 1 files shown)


Westbrook:
Can you confirm that this works with a repro in StackBlitz or the like. Would be happy to update if we are incorrect here, but by my experience the approach you've outlined is not possible.

When using `addEventListener()` if your callback uses `this` the bind is required to prevent `this` from being `undefined`. Alternatively, this could be done via class field, or the callback could choose to reference the target of the event, however these are more advanced techniques that I am not sure belong in this write up, so unless you're binding via the `lit-html` template I don't see how the bind can be skipped.

docs/faq/events.md
```diff
@@ -67,15 +67,15 @@ If you are adding a listener to your element, from inside your element, you shou
...
+    this.addEventListener('done', this.handleDone);
 }
 handleDone(event) {
     this.done = true;
     // Do the work you'd like to complete when "done" here.
 }
 ```
 
-Side note: as you will be adding this listener manually, you _will_ need to ensure that `this` is bound to your liking. See the `.bind()` syntax above.
+Side note: as long as your event handler is a method of your custom element, there is no need to manually `bind` it. When calling `addEventListener`, the value of `this` inside the event handler [is a reference to the element](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#The_value_of_this_within_the_handler).
```
Westbrook, line 78: With this correction, I'm not sure we should side note this...even for me, having written this, it brought confusion in reviewing your update 🤦  In that it doesn't break if you were to bind, I think we should leave the idea of bind out of this sectional altogether. Thoughts?

bschlenk: I can buy that! Although I don't know if it is well known that `this.addEventListener` auto-binds to whatever the event is attached to. Do you think it is better to call it out, given that the section below does call out that bind is required when attaching the event to a different element?

bschlenk:
Here's a stackblitz anyway :) https://stackblitz.com/edit/lit-element-m2t6zq
Clicking on the element is able to alert the current value of `this.name`, even without binding.

PR #1690, opened 2020-05-29
feat(es-dev-server): improve syntax error logging

LarsDenBakker:
This improves the error logging of syntax errors in es-dev-server. It no longer prints a useless stack trace, and it produces a clickable link to the file that has errors.


packages/es-dev-server/demo/syntax-error/syntax-error-resolve-imports.js
```diff
@@ -0,0 +1 @@
+import foo from './bar.js
```
thepassle, line 1: you forgot a quote here 
image

LarsDenBakker: huhuhu

PR #1692, opened 2020-05-29
fix(semantic-dom-diff): shadowDom acceps all nodes with shadowRoot

### LINKED ISSUE #1691: [semantic-dom-diff] assertions should accept all nodes that can take a shadow root
## Description of Bug

Weird case question. I have an element rendering into a non-standard place in order to accomodate a situation:
```typescript
    expect(element).shadowDom.to.equalSnapshot();
```
The element still has a shadow root, but it’s not off the element - it’s off the `renderRoot`. Calling `shadowRoot` on element returns null.
```typescript
    expect(element.renderRoot).shadowDom.to.equalSnapshot();
```
this is blows up right here:
```javascript
  chai.Assertion.addProperty('shadowDom', function shadowDom() {
    new chai.Assertion(this._obj.nodeType).to.equal(1);
    utils.flag(this, 'shadowDom', true);
  });
```
because the nodeType of `renderRoot` is a `DocumentFragment` (11) instead of `Element` (1).

## Resolution

@LarsDenBakker said:
> new chai.Assertion(this._obj.nodeType).to.equal(1);
> should check for all nodes that can take a shadow root
> sure, you can open an issue. or create a PR :smile:
> or both :beer:

Here is the issue for upcoming PR

ZempTime:
Closes #1691

### DIFF
--- a/packages/semantic-dom-diff/chai-dom-diff.js
+++ b/packages/semantic-dom-diff/chai-dom-diff.js
@@ -34,6 +34,7 @@ export const chaiDomDiff = (chai, utils) => {
    */
   chai.Assertion.addProperty('lightDom', function lightDom() {
     new chai.Assertion(this._obj.nodeType).to.equal(1);
+
     utils.flag(this, 'lightDom', true);
   });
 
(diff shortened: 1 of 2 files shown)


ZempTime:
Still unsure if `DocumentFragment` has `innerHTML`. While these tests passed locally, I feel like I'm still missing something. Perhaps it's best to rescind, and if I'm doing weird things, I should simply:
```
    expect(el.renderRoot.innerHTML).to.equalSnapshot();
```

LarsDenBakker:
A document fragment has inner html, so thats fine.

Issue #1351, opened 2020-02-19
[scoped-elements] handle html in attributes


LarsDenBakker:
The scoped elements tagname rewrite currently matches html in attributes like this:

```html
<div foo="<my-element>">
```

LarsDenBakker:
@manolakis did we end up fixing this?

manolakis:
Not yet 🙈

manolakis:
I'm going to start working on it

PR #1693, opened 2020-05-30
fix(scoped-elements): html in attributes

manolakis:
Changed the regexp to a custom algorithm to stop scoping HTML inside attributes.

fixes: https://github.com/open-wc/open-wc/issues/1351


manolakis:
Do you think is ok to merge it or maybe it could be improved?

LarsDenBakker:
Do we still have our benchmarks? Maybe we can check if this hurts performance much?

manolakis:
Good idea. I'll update the performance test

PR #1694, opened 2020-06-03
feat(demoing-storybook): Avoiding absolute imports behind a CLI flag

fcamblor:
In my setup, I needed to rely on `storybook-prebuilt` being served under a sub directory (http://localhost:8080/storybook/).

I tried with `demoing-storybook` but it didn't worked because some absolute paths were generated in storybook-ui.

This PR aims at providing a `--relative-imports` CLI flag allowing to generate relative imports instead of absolute imports, when generating storyook-ui imports.


LarsDenBakker:
I'm not a fan of adding flags like this. I think we can just always use relative imports, but we need to make sure it covers all the use cases.

Issue #1681, opened 2020-05-25
The future of open-wc


LarsDenBakker:
Open Web Components started two years ago with a simple goal: provide helpful tools and sensible recommendations to make it easier to write web components. Having recently received open-wc's 1000th GitHub star, we are pleased to see our efforts are paying off for many. 

Our goal has always been to reuse as much of the existing tooling as possible, but we ended up developing some of our own custom tooling as well. Either to ensure web components are properly supported, or to reduce the tooling complexity when writing web components. 

This growth is painfully obvious on our website (https://open-wc.org/), which is becoming harder to navigate as we create more products. This issue was also highlighted by some of our community members, and after a lot of discussion we decided that we want to make some changes to open-wc. 

We're still going through all the details of what it would look like exactly, but our primary goal is to ensure that open-wc can return to its core purpose of helping people develop web components. For our more generic web development tools we will create a new home, where we can bring the simpler and lightweight development experience to more users.

This will not be a big bang transition. We will do this gradually, being careful to not create too many changes for you to keep up with. We will share more details when they become available.

LarsDenBakker:
Also see: https://gist.github.com/web-padawan/970b47b51bc9e4f774c28e14574cad51

LarsDenBakker:
We have started work on a new repository at: https://github.com/modernweb-dev/web

LarsDenBakker:
Hi all 👋 An update from our side. We are now live with "Modern Web", that will be the place where we will create generic tooling related to buildless development. open-wc will be a child project, with a focus on web components.

Check out the blog post here: https://modern-web.dev/blog/introducing-modern-web/

The transition will be gradual, we will slowly move packages one by one.

tvvignesh:
@LarsDenBakker This is awesome. Really appreciate the efforts from you all. One suggestion. It would be better to link this issue in the related repo's README (eg. es-dev-server) so that people would migrate from `es-dev-server` to `@web/dev-server`

Thanks.

PR #1702, opened 2020-06-09
feat: update rollup dependencies

LarsDenBakker:


moderndeveloperllc:
@LarsDenBakker Any blocker on this? It would be really nice to get in.

LarsDenBakker:
Im working on some other improvements which will include this. Do you need it for something specifically?

moderndeveloperllc:
@LarsDenBakker Nothing really. Just about to start a new project, but I have node 14, so I shouldn't have the Mac issue with older versions. I'm guessing is the changes talked about in #1681 happen, this will all change anyways!

PR #1708, opened 2020-06-11
perf: (DedupeMixin) reduce loop, memory usage and function call

drix:
Reduce number of loop interactions, memory usage (no need generate array for the chain), remove unnecessary private function and reduce code size.


bennypowers:
Apologies for letting this go stale

TL;DR: we like it, and so far LGTM, but we'd like to see perhaps some regression tests, and to get an idea of if this will affect any edge cases

drix:
Hi @bennypowers, I hope that you like, I do enjoy to make code simples, smaller and performant, I work on this file since was the first I opened with the intent of collaborate, I will be glad to continue improve performance in other files if we succeed here.
 I'll add some test, regression should be inexistent since it remove a private function, I did run a full scan to see if the function was used anywhere and find no match.

bennypowers:
@bashmish Would you mind taking a :eyes:, please?

manolakis:
The change looks good to me.

IMO the added unit test is covered with the previous tests, so I would remove it.

drix:
Unnecessary test removed

LarsDenBakker:
Sorry this one took so long to get back to. I merged it now, thanks a lot 👍

PR #1711, opened 2020-06-12
fix(testing-karma): disable restartOnFileChange option

LarsDenBakker:
After a lot of debugging, I found out that there is a race condition inside the karma browser client where sometimes it cancels a test run right after a reload when the `restartOnFileChange` option is set. 

Disabling this makes it more stable, but you have to wait for a test run to finish before rerunning tests again.


daKmoR:
nice 👌

PR #1721, opened 2020-06-20
fix(es-dev-server): only cache HTML when there is lastModified

LarsDenBakker:


packages/es-dev-server/demo/plugin-serve/README.md
```diff
@@ -0,0 +1 @@
+erewrewrweewrw
```
daKmoR, line 1: you wanna keep this file? maybe add some context what it's for? 😅

LarsDenBakker: uhhh nope we don't :D

PR #1726, opened 2020-06-23
feat(es-dev-server): add cors option

### LINKED ISSUE #1696: [es-dev-server] add cors option
I have cors issues when serving es-dev-server and loading the files served by this server from another one.

`--cors` flag could be added for activating it like it's done for other servers (see https://github.com/vercel/serve/blob/6213de8946122c68878a175d620fa3ffb58f75ab/bin/serve.js#L187)

MathieuPuech:
Closes #1696


packages/es-dev-server/src/create-middlewares.ts
```diff
@@ -197,6 +199,10 @@ export function createMiddlewares(config: ParsedConfig, fileWatcher = chokidar.w
   middlewares.push(createPluginMimeTypeMiddleware({ plugins }));
   middlewares.push(createPluginServeMiddlware({ plugins }));
 
+  if (cors) {
```
LarsDenBakker, line 202: I think we should move this above the `  middlewares.push(createPluginServeMiddlware({ plugins }));` so that we also allow CORS for files served from plugins.

MathieuPuech: I've added an entry in README.md and moved the code one line before

LarsDenBakker:
👍

PR #1727, opened 2020-06-25
feat(create): add web app manifest

aress31:
- Addition of a `manifest.json` file
- Addition of `logo`, `favicon` and `robots.txt` files
- Addition of a few extra tags in `index.html` -> `<head>`


bennypowers:
You'll have to change the commit message
```
git commit --amend
```
change message to "feat(create): add web app manifest"

packages/create/src/generators/app-lit-element-ts/templates/static/index.html
```diff
@@ -7,6 +7,26 @@
...
+    It will be replaced with the URL of the `public` folder during the build.
+    Only files inside the `public` folder can be referenced from the HTML.
+
+    Unlike '/favicon.ico' or 'favicon.ico', '%PUBLIC_URL%/favicon.ico' will
+    work correctly both with client-side routing and a non-root public URL.
+    Learn how to configure a non-root public URL by running `npm run build`.
+  -->
+  <link
+    rel="apple-touch-icon"
+    href="%PUBLIC_URL%/logo192.png"
```
thepassle, line 21: how will this get replaced by the build?

I think it would be nice to have a solution here that is not tied to a buildstep (e.g. works locally/buildless as well)

thepassle:
This will require a change to `building-rollup` as well to copy over `manifest.json` and favicons/imges and stuff as well. We'll have to use `rollup-plugin-copy` for that, and make sure it runs during the `buildStart` hook. E.g. something like:

```js
    copy({
        hook: 'buildStart',
        targets: [
          { src: 'path/to/manifest.json', dest: 'dist/' },
        ]
      }),
```

Im not sure if that technically would be a breaking change 🤔

aress31:
@thepassle, yes that was my worry, but this could be actually a good thing to add `rollup-plugin-copy`. Indeed the directory tree of a newly created app could contain a `public` and an `assets` folder, these folder could automatically be copied to the `dist` final folder on build time so that every `images, fonts and assets` would be copied without any extra tweaking from devs.

What do you think of this?

bennypowers:
@thepassle if we don't already copy, how is that a breaking change?

thepassle:
> @thepassle if we don't already copy, how is that a breaking change?

I wonder what happens if a user uses our config, extended it to use the copy plugin, and then _we_ add the copy plugin

bennypowers:
hmmm yeah I'm not sure what you'd get with multiple copy instances

aress31:
How about adding a an array/JSON object bar to the base config so that if a user want to add/override the rollup-copy-plugin it can be done.

aress31:
@thepassle and @bennypowers what did you decided regarding this pending PR?

PR #1734, opened 2020-06-28
fix(create): TypeScript support

aress31:
- Add default returned type to `render()` to avoid `Missing return type on function.eslint@typescript-eslint/explicit-module-boundary-types)` errors.
- Apply recommendations from https://github.com/benmosher/eslint-plugin-import for TypeScript support.

### DIFF
--- a/packages/create/src/generators/app-lit-element-ts/templates/_MyApp.ts
+++ b/packages/create/src/generators/app-lit-element-ts/templates/_MyApp.ts
@@ -1,4 +1,4 @@
-import { LitElement, html, css, property } from 'lit-element';
+import { css, html, LitElement, property, TemplateResult } from 'lit-element';
 import { openWcLogo } from './open-wc-logo.js';
 
 export class <%= className %> extends LitElement {
@@ -49,7 +49,7 @@ export class <%= className %> extends LitElement {
     }
   `;
 
-  render() {
+  render(): TemplateResult {
     return html`
       <main>
         <div class="logo">${openWcLogo}</div>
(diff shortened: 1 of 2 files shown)


packages/create/src/generators/linting-eslint-ts/templates/_.eslintrc.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:import/errors',
     'plugin:import/warnings',
+    'plugin:import/typescript',
```
thepassle, line 11: does adding this create any additional linting errors?

packages/create/src/generators/app-lit-element-ts/templates/_MyApp.ts
```diff
@@ -49,7 +49,7 @@ export class <%= className %> extends LitElement {
     }
   `;
 
-  render() {
+  render(): TemplateResult {
```
thepassle, line 52: we should also add this in the component starter, too

packages/create/src/generators/linting-eslint-ts/templates/_.eslintrc.js
aress31: Not entirely sure but I think it was what triggered the `Missing return type on function.eslint@typescript-eslint/explicit-module-boundary-types` linting error on the `render()` method.

packages/create/src/generators/app-lit-element-ts/templates/_MyApp.ts
aress31: Could you point me to this file so that I can add the relevant code?

Issue #1739, opened 2020-07-02
[eslint-config] Improve ESLint v7 support by adding plugins to peerDeps


busterc:
When we extend `open-wc/eslint-config` and install it on a project, the various plugins found in `dependencies` are not resolved by ESLint v7:
```json
"dependencies": {
    "eslint-plugin-babel": "^5.3.0",
    "eslint-plugin-html": "^6.0.0",
    "eslint-plugin-import": "^2.18.2",
    "eslint-plugin-lit": "^1.2.0",
    "eslint-plugin-no-only-tests": "^2.4.0",
    "eslint-plugin-wc": "^1.2.0"
  }
```
Therefore, we have to manually install each plugin individually as a `devDependencies` for each project that extends `open-wc/eslint-config`.

You'll notice that the `airbnb-base` config includes `plugin-import` in `peerDependencies` and they refer to installing peerDeps in their [README](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base#eslint-config-airbnb-base-1) and recommending `install-peerdeps` as a shortcut.

Like airbnb-base, it would excellent to get to use such a shortcut, but we need those plugins added as `peerDependencies`:
```json
"peerDependencies": {
    "eslint-plugin-babel": "^5.3.0",
    "eslint-plugin-html": "^6.0.0",
    "eslint-plugin-import": "^2.18.2",
    "eslint-plugin-lit": "^1.2.0",
    "eslint-plugin-no-only-tests": "^2.4.0",
    "eslint-plugin-wc": "^1.2.0"
  }
```
Please.

PR #1740, opened 2020-07-02
feat(eslint-config): Improve ESLint v7 support by adding plugins to peerDeps

busterc:
To better support ESLint v7 peerDep resolution.
See: https://github.com/open-wc/open-wc/issues/1739
==
When we extend `open-wc/eslint-config` and install it on a project, the various plugins found in `dependencies` are not resolved by ESLint v7:
```json
"dependencies": {
    "eslint-plugin-babel": "^5.3.0",
    "eslint-plugin-html": "^6.0.0",
    "eslint-plugin-import": "^2.18.2",
    "eslint-plugin-lit": "^1.2.0",
    "eslint-plugin-no-only-tests": "^2.4.0",
    "eslint-plugin-wc": "^1.2.0"
  }
```
Therefore, we have to manually install each plugin individually as a `devDependencies` for each project that extends `open-wc/eslint-config`.

You'll notice that the `airbnb-base` config includes `plugin-import` in `peerDependencies` and they refer to installing peerDeps in their [README](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base#eslint-config-airbnb-base-1) and recommending `install-peerdeps` as a shortcut.

Like `airbnb-base`, it would excellent to get to use such a shortcut, but we need those plugins added as `peerDependencies`:

```json
"peerDependencies": {
    "eslint-plugin-babel": "^5.3.0",
    "eslint-plugin-html": "^6.0.0",
    "eslint-plugin-import": "^2.18.2",
    "eslint-plugin-lit": "^1.2.0",
    "eslint-plugin-no-only-tests": "^2.4.0",
    "eslint-plugin-wc": "^1.2.0"
  }
```

### DIFF
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -22,6 +22,14 @@
     "eslint",
     "config"
   ],
+  "peerDependencies": {
+    "eslint-plugin-babel": "^5.3.0",
+    "eslint-plugin-html": "^6.0.0",
+    "eslint-plugin-import": "^2.18.2",
+    "eslint-plugin-lit": "^1.2.0",
+    "eslint-plugin-no-only-tests": "^2.4.0",
+    "eslint-plugin-wc": "^1.2.0"
+  },
   "dependencies": {
     "babel-eslint": "^10.0.3",
     "eslint": "^6.7.2",


LarsDenBakker:
@busterc Sorry for the late response here. I get the general idea, but I think it's weird to have it both as dev and peer dependencies. Ideally we should be able to resolve the path before passing it to eslint, but I don't know if eslint supports that yet.

busterc:
> @busterc Sorry for the late response here. I get the general idea, but I think it's weird to have it both as dev and peer dependencies. Ideally we should be able to resolve the path before passing it to eslint, but I don't know if eslint supports that yet.

@LarsDenBakker Thanks for looking into this!

It does seem weird to have plugins as both `dependencies` and `peerDependencies`; however, for shareable configs that are extended, ESLint simply does not install plugins regardless of where they're defined. Over the years there have been many issues posted about this inconvenience and ESLint developers continue to say ['add plugins as peerDependencies'](https://github.com/eslint/eslint/issues/2518) Eventually the [WIP to re-implement eslint configs](https://github.com/eslint/rfcs/pull/9#issuecomment-653219866) should resolve those complaints.

In the meantime, you'll notice that the `airbnb-base` config includes `eslint-plugin-import` in both `dependencies` and `peerDependencies` to minimize the pain when extending their configs: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/package.json#L62-L69

[One alternative](https://github.com/eslint/eslint/issues/2518#issuecomment-102433136) is to include the plugins _**only in**_ `peerDependencies`; however that would require installing those peerDeps even when not extending the config. As Airbnb didn't opt for that, I guessed open-wc would prefer not too either. Otherwise, I can certainly update the PR to remove the plugin references from `dependencies`?

LarsDenBakker:
Thanks for those references. Seems like it's the best option we have right now then.

PR #1752, opened 2020-07-10
fix(es-dev-server): fix messagechannel to be ES5 compatible

vikerman:
The inlined script on message-channel.ts is not transpiled. This code was using
the arrow function which was failing in ES5 only browsers like IE11.

The fix is to use non-arrow anonymous function similar to rest of the script.

Tested manually with IE11.


LarsDenBakker:
Thanks

PR #1776, opened 2020-07-29
fix(building-rollup): allow to configure absolute service worker urls

daKmoR:


LarsDenBakker:
@daKmoR I think this one is obsolete now?

daKmoR:
indeed

PR #1778, opened 2020-08-01
feat: support building of multi page applications (mpa)

daKmoR:
Solves:
- Allows to provide a `{ rootDir }` to all configs (Basic, SPA, MPA)
- Allows to use absolute URLs `/` in the html source when using rollup-plugin-html
- Enables correct import.meta processing using `{ rootDir }`
- Creates preload links relative to the currently processing html file
- Creates service worker urls relative to the currently processing html file


packages/building-rollup/src/utils.js
```diff
@@ -39,11 +40,30 @@ function pluginWithOptions(plugin, userConfig, defaultConfig, ...otherParams) {
   return plugin(config, ...otherParams);
 }
 
+/**
+ *
+ * @param {SpaOptions} userOptions
+ * @param {string} outputDir
+ * @param {string} htmlFileName
+ */
+function createSwPath(userOptions, outputDir, htmlFileName = '') {
```
LarsDenBakker, line 49: Does `htmlFileName` need to have a default value? Isn't this always a known value?

packages/building-rollup/src/utils.js
```diff
@@ -39,11 +40,30 @@ function pluginWithOptions(plugin, userConfig, defaultConfig, ...otherParams) {
...
+  let swPath;
+  if (typeof userOptions.workbox === 'object' && userOptions.workbox.swDest) {
+    swPath = userOptions.workbox.swDest.replace(`${outputDir}/`, '');
+  } else {
+    swPath = './sw.js';
+  }
+
+  swPath = path.relative(path.dirname(htmlFileName), swPath);
```
LarsDenBakker, line 57: I don't think this will work for windows. We need to check which variables are always POSIX, or if some use native file system separators.

packages/rollup-plugin-html/rollup-plugin-html.js
```diff
@@ -24,6 +24,14 @@ const { createHtmlAsset, createHtmlAssets } = require('./src/createHtmlAssets');
 const watchMode = process.env.ROLLUP_WATCH === 'true';
 const defaultFileName = 'index.html';
 
+/**
+ * @param {string} id
+ * @param {string | undefined} rootDir
+ */
+function isAbsoluteUrl(id, rootDir) {
+  return rootDir && id.startsWith('/') && !id.startsWith(rootDir);
```
LarsDenBakker, line 32: You already check for the presence of `rootDir` below, so you can make it a required parameter here.

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
```diff
@@ -72,7 +73,12 @@ function rollupPluginPolyfillsLoader(pluginOptions = {}) {
         let preloaded = [];
         for (const entrypoint of entrypoints) {
           preloaded.push(entrypoint.importPath);
-          preloaded.push(...entrypoint.chunk.imports);
+
+          // js files (incl. chunks) will always be in the root directory
+          const pathToRoot = path.relative('./', path.dirname(entrypoint.importPath));
+          for (const chunkPath of entrypoint.chunk.imports) {
+            preloaded.push(path.join(pathToRoot, chunkPath));
```
LarsDenBakker, line 80:
```suggestion
            preloaded.push(path.posix.join(pathToRoot, chunkPath));
```

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
```diff
@@ -72,7 +73,12 @@ function rollupPluginPolyfillsLoader(pluginOptions = {}) {
         let preloaded = [];
         for (const entrypoint of entrypoints) {
           preloaded.push(entrypoint.importPath);
-          preloaded.push(...entrypoint.chunk.imports);
+
+          // js files (incl. chunks) will always be in the root directory
+          const pathToRoot = path.relative('./', path.dirname(entrypoint.importPath));
```
LarsDenBakker, line 78:
```suggestion
          const pathToRoot = path.posix.relative('./', path.posix.dirname(entrypoint.importPath));
```

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
```diff
@@ -72,7 +73,12 @@ function rollupPluginPolyfillsLoader(pluginOptions = {}) {
         let preloaded = [];
         for (const entrypoint of entrypoints) {
           preloaded.push(entrypoint.importPath);
-          preloaded.push(...entrypoint.chunk.imports);
+
+          // js files (incl. chunks) will always be in the root directory
```
LarsDenBakker, line 77: rollup outputs a flat directory by default, but you can actually output into directories

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
LarsDenBakker: I don't understand what this does. It create a path relative to the same directory?

LarsDenBakker:
Looks good 💪

packages/building-rollup/src/utils.js
daKmoR: indeed - changed

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
daKmoR: yeah... ok to fix/improve once we have a need for it?

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
daKmoR: jup you get `foo/index.html` and it will give you `../`  if you have `index.html` it gives you ``

LarsDenBakker: This is what I get in the node terminal>

```js
const path = require('path');
> path.relative('./', 'foo/bar')
'foo/bar'
> path.relative('./', 'foo')
'foo'
```

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
LarsDenBakker: it would be a breaking change so preferably not

packages/rollup-plugin-polyfills-loader/rollup-plugin-polyfills-loader.js
daKmoR:
```
{
  importPath: './inline-module-index-1.js',
  pathToRoot: '',
  preload: 'navigation.js'
}
{
  importPath: '../inline-module-index-12.js',
  pathToRoot: '..',
  preload: '../navigation.js'
}
```

so we "extract" the leading ".." so we can add them to our chunk path which will then be the preload path

LarsDenBakker: Ah now I understand, makes sense. That would automatically work when files are output into directories as well. So we can remove the comment and the point below is solved as well.

PR #1779, opened 2020-08-03
chore(docs): Fix JavaScript arrow notation style

pndewit:
The arrows are not getting any space, meaning anything that is printed right after the arrow is rendered on top of it.

I changed the following rule:
```css
.arrow{display:inline-block;width:0;height:0}
```

To:
```css
.arrow{display:inline-block}
```


daKmoR:
uh nice it fixed the following 💪

## Current
Screenshot 2020-08-04 at 18 16 43

## With PR

Screenshot 2020-08-04 at 18 16 48

daKmoR:
awesome thank you for this fix 💪

pndewit:
Glad I could help 👍

PR #1784, opened 2020-08-04
fix: support rollup watch mode in rollup-plugin-html

### LINKED ISSUE #1256: Rollup config does not pick up changes to index.html in watch mode
I am using the rollup config `createDefaultConfig` from `@open-wc/building-rollup` as supplied by `npm init @open-wc`. I am not using esdevserver during development as suggested, instead I need to use my own server, so I run `rollup -c -w` during development and my dev server serves the build that is generated. Rollup successfully picks up changes to js and rebuilds, but does not pick up on changes to `index.html`, so I have to restart `rollup -c -w`. It would be great if the rollup config could be updated to handle this.

SamiHiltunen:
This commit fixes two issues that prevent rollup-plugin-html from
being used with Rollup's watch mode:

1. Each run adds the entrypoint from the input HTML file to the
   input modules. This commit clears the whole build state each run
   so the inputs from earlier runs are not included in the later
   builds.
2. Changes to the input HTML files are not detected. This is fixed
   by adding Rollup's input file as a watch file if it is a valid
   input for the plugin.

Closes #1256


packages/rollup-plugin-html/rollup-plugin-html.js
```diff
@@ -87,6 +106,8 @@ function rollupPluginHtml(pluginOptions) {
           return null;
         }
       }
+
+      htmlFiles = [];
```
LarsDenBakker, line 110: Isnt this already reset by the reset function?

SamiHiltunen:
@LarsDenBakker This should be ready for review. I don't really have any experience with JS nor Rollup, so please let me know if there's anything that should be changed or kept in mind.

LarsDenBakker:
Looks good to me, thanks. I didn't expect options to be called in watch mode.

PR #1789, opened 2020-08-05
docs(component-libraries): remove extra tab

EisenbergEffect:
An extra tab in the source file causes the FAST library list item to be indented and rendered as a sub-bullet.


LarsDenBakker:
👍 Thanks. The CI failure is because browserstack isn't always stable.

Issue #1517, opened 2020-04-15
[testing-karma] Upgrade karma to 5.0.1


jarrodek:
New Karma version was released that has fixed a security vulnerability related to `minimist` library.
https://github.com/karma-runner/karma/releases/tag/v5.0.0

Please, update tooling to use new Karma version.

LarsDenBakker:
Would be good to update, we first need to see if there will be any breaking changes for us.

PR #1803, opened 2020-08-10
chore: upgrade to karma 5.1.1

marcoswebdev:
Fixes https://github.com/open-wc/open-wc/issues/1517

### DIFF
--- a/packages/karma-esm/package.json
+++ b/packages/karma-esm/package.json
@@ -48,7 +48,7 @@
     "@babel/plugin-proposal-decorators": "^7.8.3",
     "@babel/preset-typescript": "^7.8.3",
     "@types/lodash-es": "^4.17.3",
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
Unfortunately this might be a breaking change. We need to look into that.

marcoswebdev:
Yeah, all tests are green, what else can we check?

The [breaking changes](https://github.com/karma-runner/karma/blob/v5.0.0/CHANGELOG.md#breaking-changes) are:

- Karma plugins which rely on the fact that Karma uses Bluebird promises may break as Bluebird-specific API is no longer available on Promises returned by the Karma core
- server: Deprecated createPreprocessor removed, karma-browserify < 7 version doesn't work
- no more testing on node 8.

LarsDenBakker:
Seems good indeed, thanks.

PR #1833, opened 2020-09-10
docs: testing karma coverage check

mario-aleo:
After #1556, as stated by @Westbrook on #1164, the `coverageReporter` `global` setting is now defined inside the `check` attribute instead of the `thresholds`  :)


LarsDenBakker:
Thanks!

PR #1834, opened 2020-09-10
feat(semantic-dom-diff): diff tags from Scoped Elements with its original name

jdvivar:
I think the readme update speaks for itself :)


LarsDenBakker:
Thanks for the PR! @manolakis recently added the original tag name as an attribute on the element. I think we could leverage that to avoid having to pass the scoped element names. Perhaps we could even turn it on my default?

packages/semantic-dom-diff/get-diffable-html.js
```diff
@@ -177,9 +179,18 @@ export function getDiffableHTML(html, options = {}) {
     return attrStr;
   }
 
+  /** @param {Element} el */
+  function getLocalName(el) {
+    const { unscopedTag } = /^(?<unscopedTag>.*?)(-\d+)?$/.exec(el.localName).groups;
```
manolakis, line 184: I like how you create this function to return the localName 🤗. However, I think you could improve the DX if you use this function to check if `data-tag-name` exists returning its content or the `localName` if not. Then you don't need to pass the `scopedTags` collection to the `equalSnapshot` method. What do you think?

Maybe something like this could work
```js
function getLocalName(el) {
  return el.getAttribute('data-tag-name') || el.localName;
}
```

jdvivar: I wasn't aware of the `data-tag-name` which was added recently, I should have reviewed it first, damn! This is better in all the ways.

Thanks @LarsDenBakker and @manolakis for the review comments. Let me update the request.

manolakis:
Looks good to me. Would you mind to squash your commits into a single one? 🤗

Thanks!

manolakis (approved):

PR #1832, opened 2020-09-09
Update terser to v5

abdonrd:
Now `terser.minify` is async.

https://github.com/terser/terser/blob/master/CHANGELOG.md#v500-beta0

~~Depends on #1836~~

PR #1836, opened 2020-09-11
chore: update html-minifier to html-minifier-terser

abdonrd:
Motivation: https://github.com/open-wc/open-wc/pull/1832#pullrequestreview-486640682

In the snapshoots just remove a few extra `;`.


Westbrook (approved):
LGTM.

LarsDenBakker:
LGTM as well. However since we still use lerna in this repository, we need to create a commit for each package separate. I can't do that from the git GUI unfortunately.

abdonrd:
Why create a commit for each package separate? 🤔

LarsDenBakker:
That's how lerna works, the commits are a trigger for cutting a release and creating a changelog.

abdonrd:
@LarsDenBakker PR updated. Now it's right?

LarsDenBakker:
Yep, thanks!

PR #1839, opened 2020-09-12
feat(semantic-dom-diff): diff options now process scoped elements

jdvivar:
Hey guys!

It's me again, trying to improve our `semantic-dom-diff` package to better process scoped elements.

I realised... why put the burden on devs to get the real names of scoped elements' tags when using diff options? It's important they know, but snapshots shouldn't care that much. I myself ended up with a test which looked like this:

```js
const getScopedTag = (el, tag) => el.constructor.getScopedTagName(tag, el.constructor.scopedElements)
const domDiffConfig = (el) => {
  return {
    ignoreChildren: [
      getScopedTag(el,'ing-input-amount'),
      getScopedTag(el,'ing-select'),
      getScopedTag(el,'ing-labs-spinner')
    ],
    ignoreAttributes: [
      getScopedTag(el,'ing-labs-spinner')
    ]
  }
}
```

I improved my previous contribution, from `getLocalName` to `getTagName` which will work both for  a `Node` or `Element` input type, and use it in every instance of the code needing a tag name.

Now you can ignore scoped elements' tags like a breeze! 💨

### DIFF
--- a/packages/semantic-dom-diff/README.md
+++ b/packages/semantic-dom-diff/README.md
@@ -288,9 +288,11 @@ it('renders correctly', async () => {
 
 **[Scoped elements](https://open-wc.org/scoped-elements/)**
 
-You might want to ignore scoped elements when working with _Shadow DOM snapshots_, like so:
+Let's take this example going forward:
 
 ```js
+import { ScopedElementsMixin } from '@open-wc/scoped-elements';
+import { LitElement } from 'lit-element';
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
Looks like a good change to me, I'm glad to see the tag name attribute is really useful in this way. What do you think @manolakis ?

manolakis (approved):
Change looks good to me.

Thanks for your contribution! 🤗

PR #1851, opened 2020-09-23
chore: test packages with "next-major" release from lit-html/element

Westbrook:
This positions a branch for us to test the upcoming major releases of `lit-html` and `lit-element` in our tools and likely won't be merged until those dependencies get moved to their stable channel.

Notes:
- there were a number of versions of both packages and I'm not sure if there was a reason for it
  - I flattened ALL `lit-element` references (including one reference to a pre-2.0 release
  - I flattened `lit-element` references to the `1.*` release as it seemed that some of the `0.*` references were meant to be

In theory, if we run `yarn publish --canary --dist-tag next` we'll be able to make successive releases on the `next` channel for these packages and keep up with the ongoing development of `lit-html/element` as we go.

Thoughts?

### DIFF
--- a/packages/testing-helpers/src/lib.js
+++ b/packages/testing-helpers/src/lib.js
@@ -1,10 +1,8 @@
-import { TemplateResult } from 'lit-html';
-
 export const isIterable = object => object != null && typeof object[Symbol.iterator] === 'function';
 
 function isValidNonIterableRenderArg(x) {
   return (
-    x instanceof TemplateResult ||
+    typeof x._$litType$ !== 'undefined' ||
     x instanceof Node ||
     typeof x === 'number' ||
     typeof x === 'boolean' ||
--- a/packages/testing-helpers/src/litFixture.js
+++ b/packages/testing-helpers/src/litFixture.js
@@ -1,9 +1,8 @@
-import { TemplateResult } from 'lit-html';
 import { fixtureWrapper } from './fixtureWrapper.js';
 import { render } from './lit-html.js';
 import { elementUpdated } from './elementUpdated.js';
 import { NODE_TYPES } from './lib.js';
-import { getScopedElementsTemplate } from './scopedElementsWrapper.js';
+// import { getScopedElementsTemplate } from './scopedElementsWrapper.js';
 
 /**
  * @typedef {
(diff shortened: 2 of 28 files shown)


tsconfig.build.types.json
```diff
@@ -20,7 +20,6 @@
     "packages/semantic-dom-diff/get-diffable-html.js",
     "packages/semantic-dom-diff/src/utils.js",
     "packages/polyfills-loader/*.js",
-    "packages/import-maps-resolve/*.js",
-    "packages/scoped-elements/*.js"
+    "packages/import-maps-resolve/*.js"
```
Westbrook, line 23: Skip typing this for now as it's likely not compatible with `lit-next`.

packages/testing-helpers/src/litFixture.js
```diff
@@ -43,7 +42,13 @@ export function litFixtureSync(template, options = {}) {
     wrapper,
   );
 
-  if (template instanceof TemplateResult) {
+  if (
```
Westbrook, line 45: Reverse the check as `TemplateResult` is no longer a vale.

packages/testing-helpers/src/lib.js
```diff
@@ -1,10 +1,8 @@
-import { TemplateResult } from 'lit-html';
-
 export const isIterable = object => object != null && typeof object[Symbol.iterator] === 'function';
 
 function isValidNonIterableRenderArg(x) {
   return (
-    x instanceof TemplateResult ||
+    typeof x._$litType$ !== 'undefined' ||
```
Westbrook, line 5: Iffy. Happy to hear better ideas about this one.

justinfagnani: This is ok for now. We'll have to land on a final solution and document it on our end.

kevinpschaaf: We now have `isTemplateResult()` in module `lit-html/directive-helpers.js`.

https://github.com/Polymer/lit-html/blob/lit-next/packages/lit-html/CHANGELOG.md#added

justinfagnani:
@Westbrook should this be re-opened now? We have an API-stable prerelease coming out now.

PR #1856, opened 2020-10-01
docs: remove linting types

### LINKED ISSUE #1056: [doc] generator with linting types does not work
Accoding to [documentation](https://open-wc.org/linting/linting-types.html#linting-types), there is an option to add `linting-types-js` for enabling types linting. It's not working and I suppose it's unupdated documentation about that. What is a correct option for generator to enable that? If it still supposed to work of course.

trystan2k:
Fix #1056 

Is it necessary to do any config to do a redirect as the page will not exist anymore ? I didn't found any config on this matter, so not sure.

This is what I have understood from the issue (just need to remove the file - and there is no other mention to it in code). Please let me know if need any other change or feel free to decline the PR if this is not what needs to be done.

Thank you !


LarsDenBakker:
Thanks!

Issue #1826, opened 2020-09-03
Unused @ts-expect-error directive in @open-wc/testing-helpers


buzzub:
A change introduced in @open-wc/testing-helpers 1.8.9 (https://github.com/open-wc/open-wc/commit/88ffd995efeb94d79ee686f665d18e8ca405e3bc) is causing an error when running all our tests via jest. The addition of `// @ts-expect-error` produces this error:

```
Test suite failed to run

node_modules/@open-wc/testing-helpers/src/scopedElementsWrapper.js:32:1 - error TS2578: Unused '@ts-expect-error' directive.

32 // @ts-expect-error
 ~~~~~~~~~~~~~~~~~~~
```

The only way we can get around this error is by limiting the version of @open-wc/testing-helpers to 1.8.8

What is the purpose of the @ts-expect error ? Could @ts-ignore be used here instead?

Struggling to understand how to prevent this error as there seems to be no way to bypass it

daKmoR:
@buzzub would a `// @ts-ignore` work?

@bennypowers why did we go for expect error again? 🤔

buzzub:
@ts-ignore would work fine. @ts-expect-error doesn’t get an error / isn’t used and that causes the problem.

bennypowers:
`@ts-expect-error` is preferable to `@ts-ignore` in most cases, as `@ts-ignore` will simply suppress all errors, whereas `@ts-expect-error` tells the compiler that there should be an error on that line, for example, when unit testing bad input to a function that throws on bad input.

If the compiler finds a `@ts-expect-error` but _doesn't_ find an error afterwards, that itself becomes an error.

So it looks like according to @buzzub that in fact the original code now works properly. Perhaps @buzzub has some tsconfig which allows for the usage on https://github.com/open-wc/open-wc/commit/88ffd995efeb94d79ee686f665d18e8ca405e3bc or is using TS 4?

@buzzub  when you remove the `@ts-expect-error` comment altogether (and don't replace it with a `@ts-ignore`), do your tests compile?

buzzub:
> @buzzub when you remove the @ts-expect-error comment altogether (and don't replace it with a @ts-ignore), do your tests compile?

Yes, just removed `@ts-expect-error` line and the tests run just fine. I also switched it to `@ts-ignore` and the tests still compile fine.

We are using Typescritpt 4.0.2

FYI compiler options from tsconfig below, but I can't find any relevant config that would prevent this error:
```
"compilerOptions": {
[...]

PR #1857, opened 2020-10-01
fix(testing-helpers): replace ts-expect-error with ignore

jorenbroekema:
This is needed because different typescript versions seem to either throw or not throw, inconsistently. This makes expect-error problematic for users with typescript versions that don't throw on this. Review when the TS issue is closed by typescript.

fixes https://github.com/open-wc/open-wc/issues/1826


fernandopasik (approved):
Thanks for fixing this one

Westbrook (approved):
Yay, TS!

PR #1860, opened 2020-10-03
docs(es-dev-server): add deprecation notice

LarsDenBakker:


packages/es-dev-server/README.md
```diff
@@ -6,6 +6,12 @@ tags:
   - guides
 ---
 
+> **Notice**
+>
+> Development of es-dev-server continues under a new name: [web dev server](https://modern-web.dev/docs/dev-server/overview/). We recommend using it for new projects, and upgrading for existing projects.
```
justinfagnani, line 11: This link doesn't work, at least against the live site.

LarsDenBakker: Hmmm, it seems to work fine for me?

PR #1861, opened 2020-10-03
feat(create): update to use @web/test-runner

### LINKED ISSUE #1842: Update the project generator to web test runner
We should start using `@web/test-runner` for the [project generator](https://github.com/open-wc/open-wc/tree/master/packages/create)

More info on @web/test-runner here: https://modern-web.dev/docs/test-runner/overview/

Need more guidance, help, or instructions to get you started? Reach out to us on the Polymer slack 🙂

trystan2k:
Hi,

I have updated the generators to use WTR instead of Karma.

After all, I have added the WTR config file `web-test-runner.config.mjs` as for the typescript app generator has distinct output folder (out-tsc) as of the webcomponent one, so if I put the config in package.json directly, one or the other would fail (as the WC one uses `dist` folder). Let me know if there would be another solution to this, if you want to avoid having the config file.

Also, I have moved the `deepmerge` dependency from the testing generator to the building-rollup as this is the only place using it now.

Anything that needs to be change, please let me know !

Thanks!

Closes #1842


packages/create/src/generators/app-lit-element-ts/templates/static-testing/web-test-runner.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
+    files: 'out-tsc/test/**/*.test.js',
+    concurrency: 10,
```
LarsDenBakker, line 3: Default concurrency is based on CPU cores, we should stick to that

packages/create/src/generators/app-lit-element/templates/static-testing/web-test-runner.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
+    files: 'test/**/*.test.js',
+    concurrency: 10,
```
LarsDenBakker, line 3: same as above

packages/create/src/generators/testing-wtr-ts/templates/static/web-test-runner.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
+    files: 'dist/**/test/**/*.test.js',
+    concurrency: 10,
```
LarsDenBakker, line 3: same as above

packages/create/src/generators/testing-wtr/templates/static/web-test-runner.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
+    files: 'test/**/*.test.js',
+    concurrency: 10,
```
LarsDenBakker, line 3: same as above

packages/create/test/snapshots/fully-loaded-app/web-test-runner.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
+    files: 'test/**/*.test.js',
+    concurrency: 10,
```
LarsDenBakker, line 3: same as above

packages/create/src/generators/app-lit-element-ts/templates/static-testing/web-test-runner.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
+    files: 'out-tsc/test/**/*.test.js',
```
LarsDenBakker, line 2: let's stick with an indent of 2 spaces :)

LarsDenBakker:
I think the config file would be fine, it's a good way to teach users where they can go to modify the configuration.

trystan2k:
Hi Lars,

Thanks for the comments. I have just pushed them.

Let me know if there is anything else !

LarsDenBakker:
Thanks!

PR #1781, opened 2020-08-03
feat: add eslint-plugin-lit-a11y

thepassle:
# eslint-plugin-lit-a11y

Initial setup for `eslint-plugin-lit-a11y`

Just putting this up here to allow for contributions to happen. If you're interested in contributing, please create PRs to `feat/eslint-plugin-lit-a11y` (this branch) so we can gather everything together. 🙂

### Rules

<details>
<summary>You can find a list of to-be-implemented rules here:
</summary>

You can take the implementations of [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) as an example.

- [x] alt-text
- [x] accessible-emoji
- [x] anchor-has-content
- [x] anchor-is-valid
- [x] aria-activedescendant-has-tabindex
- [x] aria-props
- [x] ~~aria-proptypes~~ -> aria-attr-types
- [x] aria-role
- [x] aria-unsupported-elements
- [x] autocomplete-valid
- [x] click-events-have-key-events
- [ ] control-has-associated-label
- [x] heading-has-content
- [x] iframe-has-title
- [x] img-redundant-alt
- [ ] interactive-supports-focus
- [ ] label-has-associated-control
- [ ] label-has-for
- [ ] media-has-caption
- [x] mouse-events-have-key-events
- [x] no-access-key
- [x] no-autofocus
- [x] no-distracting-elements
- [ ] no-interactive-element-to-noninteractive-role
- [ ] no-noninteractive-element-interactions
- [ ] no-noninteractive-element-to-interactive-role
- [ ] no-noninteractive-tabindex
- [x] no-onchange
- [x] no-redundant-roles
- [ ] no-static-element-interactions
- [x] role-has-required-aria-props
- [x] role-supports-aria-props
- [x] scope
- [x] tabindex-no-positive
</details>

### Adding a new rule

Add a new rule by running:
```bash
git checkout feat/eslint-plugin-lit-a11y
cd packages/eslint-plugin-lit-a11y
npm i -g yo generator-eslint
yo eslint:rule
```

### Implementing a new rule

Implementing rules is made possible by the TemplateAnalyzer from [eslint-plugin-lit](https://www.npmjs.com/package/eslint-plugin-lit). You can take the implementations of [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) as an example.

Here's an example of a simple rule:

```js
return {
  TaggedTemplateExpression: (node) => {
[...]

PR #1869, opened 2020-10-10
feat(eslint-plugin-lit-a11y): interactive-supports-focus

marcoswebdev:
Initial structure for https://github.com/open-wc/open-wc/pull/1781 interactive-supports-focus rule

Example: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/interactive-supports-focus.md

### DIFF
--- a/packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
+++ b/packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
@@ -1,0 +1,108 @@
+/**
+ * @fileoverview
+ * @author open-wc
+ */
+
+const includes = require('array-includes');
+const { TemplateAnalyzer } = require('../../template-analyzer/template-analyzer.js');
+const { eventHandlersByType } = require('../utils/eventsHandlers.js');
+const { getAttrVal } = require('../utils/getAttrVal.js');
+const { getTabIndex } = require('../utils/getTabIndex.js');
+const { isDisabledElement } = require('../utils/isDisabledElement.js');
+const { isHiddenFromScreenReader } = require('../utils/isHiddenFromScreenReader.js');
+const { isInteractiveElement } = require('../utils/isInteractiveElement.js');
+const { isPresentationRole } = require('../utils/isPresentationRole.js');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Elements with an interactive role and interaction handlers (mouse or key press) must be focusable.',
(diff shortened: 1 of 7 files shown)


thepassle:
Nice, this is great!

Would it be possible to add a few more test cases, and remove the unused/commented testcases? 🙏

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,108 @@
+/**
+ * @fileoverview
+ * @author open-wc
+ */
+
+const includes = require('array-includes');
```
bennypowers, line 6: We don't need this shim, 
https://node.green/#ES2016-features-Array-prototype-includes

so let's just use `Array#includes` instead, like we're doing here on L#63

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,108 @@
...
+
+    const interactiveProps = [...eventHandlersByType.mouse, ...eventHandlersByType.keyboard];
+
+    //----------------------------------------------------------------------
+    // Public
+    //----------------------------------------------------------------------
+
+    return {
+      TaggedTemplateExpression: node => {
+        if (
```
bennypowers, line 51: We have a helper in utils, so you can
```js
const { isHtmlTaggedTemplate } = require('../utils/isLitHtmlTemplate.js');
```
and here:
```suggestion
        if (isHtmlTaggedTemplate(node)) {
```

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,108 @@
...
+                // && !isNonInteractiveElement(element.name, attributes)
+                // && !isNonInteractiveRole(element.name, attributes)
+                !hasTabindex
+              ) {
+                const role = getAttrVal(element.attribs.role);
+                if (includes(tabbable, role)) {
+                  // Always tabbable, tabIndex = 0
+                  context.report({
+                    node,
+                    message: `Elements with the '${role}' interactive role must be tabbable.`,
```
bennypowers, line 92: eslint has it's own interpolation feature which makes it easier to reuse messages.

this file is a good use case for messageId as well

take a look at the eslint docs:

https://eslint.org/docs/developer-guide/working-with-rules#using-message-placeholders

bennypowers (changes requested):

marcoswebdev:
I have created a new PR where I fix your suggestions @bennypowers!

PR #1871, opened 2020-10-11
refactor/eslint-plugin-lit-a11y/types

bennypowers:
adds 'suggestion' type to all rules
adds 'Accessibility' category to all rules
defines 'Visitor' type
defines 'is*TaggedTemplate' helpers
defines 'isAria*' helpers
updates rule template
removes instances of for...in
adds oxford comma
removes use of global isNaN
collapses nested if statements
removes eslint-disable comments in most cases


packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
```diff
@@ -40,19 +46,15 @@ module.exports = {
...
+        if (isHtmlTaggedTemplate(node)) {
           const analyzer = TemplateAnalyzer.create(node);
 
           analyzer.traverse({
             enterElement(element) {
               // todo: figure out what is meant with literalChildValue, try to see from tests
               const literalChildValue = element.children.find(child => child.type === 'text');
 
-              if (literalChildValue && emojiRegex().test(literalChildValue.data)) {
+              if (literalChildValue && EMOJI_REGEXP.test(literalChildValue.data)) {
```
bennypowers, line 57: does this need to get an instance at runtime, or is once at module init enough?

packages/eslint-plugin-lit-a11y/lib/rules/alt-text.js
```diff
@@ -4,16 +4,19 @@
...
 // Rule Definition
 //------------------------------------------------------------------------------
 
-module.exports = {
+/** @type{import('eslint').Rule.RuleModule} */
+const AltTextRule = {
   meta: {
+    type: 'suggestion',
     docs: {
       description: 'Images require alt text',
```
bennypowers, line 18: unrelated to this PR, should this rule allow ``?

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
```diff
@@ -59,8 +77,7 @@ const validityCheck = (value, expectedType, permittedValues) => {
     case 'integer':
     case 'number':
       // Booleans resolve to 0/1 values so hard check that it's not first.
-      // eslint-disable-next-line no-restricted-globals
-      return typeof value !== 'boolean' && isNaN(Number(value)) === false;
+      return typeof value !== 'boolean' && Number.isNaN(Number(value)) === false;
```
bennypowers, line 80: Was previously globalThis.isNaN, but if we're explicitly casting with `Number()`, then `Number.isNaN()` should be fine, right?

packages/eslint-plugin-lit-a11y/lib/rules/aria-attrs.js
```diff
@@ -4,55 +4,20 @@
...
 
 const { TemplateAnalyzer } = require('../../template-analyzer/template-analyzer.js');
+const { isAriaPropertyName } = require('../utils/aria.js');
+const { isHtmlTaggedTemplate } = require('../utils/isLitHtmlTemplate.js');
 
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
 
-const validAriaProps = [
```
bennypowers, line 12: This change actually adds 12 new valid props:

```js
const newKeys = [...ariaPropsMap.keys()];
[
  "aria-activedescendant",
  "aria-atomic",
  "aria-autocomplete",
  "aria-busy",
  "aria-checked",
  "aria-colcount",
  "aria-colindex",
  "aria-colspan",
  "aria-controls",
  "aria-current",
  "aria-describedby",
  "aria-details",
  "aria-disabled",
  "aria-dropeffect",
  "aria-errormessage",
  "aria-expanded",
  "aria-flowto",
  "aria-grabbed",
  "aria-haspopup",
  "aria-hidden",
  "aria-invalid",
  "aria-keyshortcuts",
  "aria-label",
  "aria-labelledby",
  "aria-level",
  "aria-live",
  "aria-modal",
  "aria-multiline",
  "aria-multiselectable",
  "aria-orientation",
  "aria-owns",
  "aria-placeholder",
  "aria-posinset",
  "aria-pressed",
  "aria-readonly",
  "aria-relevant",
  "aria-required",
  "aria-roledescription",
  "aria-rowcount",
  "aria-rowindex",
  "aria-rowspan",
  "aria-selected",
  "aria-setsize",
  "aria-sort",
  "aria-valuemax",
  "aria-valuemin",
  "aria-valuenow",
  "aria-valuetext"
]
```

packages/eslint-plugin-lit-a11y/lib/rules/aria-role.js
```diff
@@ -76,20 +79,15 @@ module.exports = {
...
-          node.tag.name === 'html'
-        ) {
+        if (isHtmlTaggedTemplate(node)) {
           const analyzer = TemplateAnalyzer.create(node);
 
           analyzer.traverse({
             enterElement: element => {
-              // eslint-disable-next-line
-              for (const attr in element.attribs) {
+              for (const [attr, value] of Object.entries(element.attribs)) {
```
bennypowers, line 87: did we need to traverse the entire prototype chain in these cases?

packages/eslint-plugin-lit-a11y/lib/rules/aria-unsupported-elements.js
```diff
@@ -40,24 +43,19 @@ module.exports = {
...
-        ) {
+        if (isHtmlTaggedTemplate(node)) {
           const analyzer = TemplateAnalyzer.create(node);
 
           analyzer.traverse({
             enterElement: element => {
               if (unsupportedElements.includes(element.name)) {
-                // eslint-disable-next-line
-                for (const attr in element.attribs) {
+                for (const attr of Object.keys(element.attribs)) {
```
bennypowers, line 52: see previous comment

packages/eslint-plugin-lit-a11y/lib/utils/isLitHtmlTemplate.js
```diff
@@ -0,0 +1,49 @@
...
+ * @template {'html'|'svg'|'css'} Name
+ * @typedef {import('estree').TaggedTemplateExpression & { tag: { type: 'Identifier'; name: Name } }} LitTaggedExpression
+ */
+
+/**
+ * Whether a node is a lit-element css-tagged template expression
+ * @param {import('eslint').Rule.Node} node
+ * @returns {node is LitTaggedExpression<'css'>}
+ */
+function isCssTaggedTemplate(node) {
```
thepassle, line 11: is this used anywhere?

packages/eslint-plugin-lit-a11y/lib/utils/isLitHtmlTemplate.js
```diff
@@ -0,0 +1,49 @@
...
+    node.tag.name === 'html'
+  );
+}
+
+/**
+ * Whether a node is a lit-html svg-tagged template expression
+ * @param {import('eslint').Rule.Node} node
+ * @returns {node is LitTaggedExpression<'svg'>}
+ */
+function isSvgTaggedTemplate(node) {
```
thepassle, line 37: same for this one

packages/eslint-plugin-lit-a11y/lib/rules/tabindex-no-positive.js
```diff
@@ -53,8 +52,7 @@ module.exports = {
 
                 const value = Number(val);
 
-                // eslint-disable-next-line
-                if (isNaN(value)) {
+                if (Number.isNaN(value)) {
```
bennypowers, line 55: see notes on globalThis.isNaN, above

packages/eslint-plugin-lit-a11y/lib/utils/aria.js
```diff
@@ -0,0 +1,24 @@
+const { aria, roles } = require('aria-query');
+
+/**
+ * @param {string} string
+ * @return {string is import("aria-query").ARIARoleDefintionKey}
+ */
+function isAriaRole(string) {
+  // @ts-expect-error: we need to disambiguate from the string type here.
+  return [...roles.keys()].includes(string);
```
bennypowers, line 9: should this maybe `.get`, like below, for perf?

thepassle:
looks like theres some type issues in the pipeline:
image

packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
thepassle: at module init should be good enough yeah

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
thepassle: should be fine I think 👍

packages/eslint-plugin-lit-a11y/lib/utils/isLitHtmlTemplate.js
bennypowers: not yet, but potentially could be

packages/eslint-plugin-lit-a11y/lib/utils/isLitHtmlTemplate.js
bennypowers: same answer

packages/eslint-plugin-lit-a11y/lib/rules/aria-attrs.js
thepassle: `validAriaProps` is just deleted? nothing replaced it? 🤔

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
```diff
@@ -122,7 +137,7 @@ module.exports = {
                 ariaAttributes.forEach(attr => {
                   const normalizedName = attr.toLowerCase();
                   // Not a valid aria-* state or property.
-                  if (aria.get(normalizedName) === undefined) {
+                  if (!isAriaPropertyName(normalizedName)) {
```
bennypowers, line 140: shouldn't we report an error here? or just leave it to `aria-attr`

packages/eslint-plugin-lit-a11y/lib/utils/aria.js
thepassle: sounds good 👍

packages/eslint-plugin-lit-a11y/lib/rules/aria-role.js
thepassle: hmm not sure, can you think of a better way to iterate in this case?

packages/eslint-plugin-lit-a11y/lib/rules/no-onchange.js
```diff
@@ -4,18 +4,21 @@
...
 
-module.exports = {
+/** @type {import("eslint").Rule.RuleModule} */
+const NoOnchangeRule = {
   meta: {
+    type: 'suggestion',
     docs: {
-      description: 'Enforce usage of onBlur over onChange for accessibility.',
-      category: 'Fill me in',
+      description: 'Enforce usage of @blur over @change for accessibility.',
```
bennypowers, line 20: Unrelated to this PR, should the error message suggest using `@input` instead of `@change` if user wants to track every user input? That's probably a common mistake, but swapping `@input` for `@blur` won't make sense to users who mistook `@change` for `@input`

bennypowers:
> looks like theres some type issues in the pipeline:
> image

yeouch, passes on my local machine 🙁  
maybe it's a problem with the yarn lock

packages/eslint-plugin-lit-a11y/lib/rules/aria-attrs.js
bennypowers: `isAriaProperty` tests against the above list, see `util/aria.js`

packages/eslint-plugin-lit-a11y/lib/rules/aria-role.js
bennypowers: usually when you want to iterate over an object's properties, `.keys` or `.entries` is where it's at.

typical use of `for..in` requires `if(Object.hasOwnProperty(key))` to prevent executing on prototypical properties.

so in most cases, my change is preferred. case for allowing `for ... in` with `// eslint-disable` is when you specifically want to iterate over every single property up the entire prototype chain

bennypowers:
https://github.com/A11yance/aria-query/pull/74

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
bennypowers: this is covered by `aria-attr` rule, so yeah

packages/eslint-plugin-lit-a11y/lib/rules/alt-text.js
thepassle: yep, would be a good addition 👍 maybe we can create a new hacktoberfest issue for it? Or we can also do it in this PR if you prefer

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
thepassle: so this is resolved right?

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
```diff
@@ -77,77 +93,58 @@ const validityCheck = (value, expectedType, permittedValues) => {
...
+        if (isHtmlTaggedTemplate(node)) {
           const analyzer = TemplateAnalyzer.create(node);
 
           analyzer.traverse({
             enterElement(element) {
-              const ariaAttributes = Object.keys(element.attribs).filter(attr =>
-                attr.startsWith('aria-'),
-              );
+              const ariaAttributes = Object.keys(element.attribs)
+                .map(attr => attr.toLowerCase())
```
thepassle, line 118: nitpick, but maybe the lowercasing could be done in `isAriaPropertyName` as well?

Or maybe we can even have a general `extractAriaAttributes(element.attribs)` helper function?

packages/eslint-plugin-lit-a11y/lib/rules/aria-attrs.js
```diff
@@ -4,97 +4,44 @@
...
-                      loc,
-                      message: 'Elements cannot use an invalid ARIA attribute.',
-                    });
-                  }
+              for (const attr of Object.keys(element.attribs)) {
+                if (isInvalidAriaAttribute(attr)) {
+                  const loc = analyzer.getLocationForAttribute(element, attr);
+                  context.report({
+                    loc,
+                    message: 'Invalid ARIA attribute "{{attr}}".',
```
thepassle, line 40: TIL, nice

packages/eslint-plugin-lit-a11y/lib/rules/img-redundant-alt.js
```diff
@@ -35,44 +48,42 @@ module.exports = {
...
+                      .includes(keyword),
+                  );
+
+                  if (contraband.length > 0) {
+                    const formatted = formatter.format(contraband);
+                    const loc = analyzer.getLocationForAttribute(element, 'alt');
+                    context.report({
+                      loc,
+                      message:
+                        '<img> alt attribute must be descriptive; it cannot contain the banned {{plural}} {{formatted}}.',
```
thepassle, line 80: what is `{{plural}}` here?

packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-types.js
bennypowers: I agree, I'll define that helper in `util/aria.js`

packages/eslint-plugin-lit-a11y/lib/rules/aria-attrs.js
bennypowers: yup, makes it easier to refactor to use `messageId` down the line

packages/eslint-plugin-lit-a11y/lib/rules/img-redundant-alt.js
bennypowers: "word"|"words"

packages/eslint-plugin-lit-a11y/lib/rules/aria-role.js
thepassle: ah yes yes, good change 👍

packages/eslint-plugin-lit-a11y/lib/rules/img-redundant-alt.js
thepassle: Its in `data`, gotcha 👍 ty

packages/eslint-plugin-lit-a11y/lib/rules/alt-text.js
bennypowers: spun out #1872

bennypowers:
NB: in order to avoid type errors (which would cause runtime errors, I'm sure), we skip analyzing any spread elements in `static get properties() {...}`.

This PR touches some stuff outside of the plugin, because of problems with `@types/eslint` and `@types/estree`:
- bump `create` with a typeerror fix, now that we're using `@types/eslint`
- add `resolutions` field to the root package.json, since `@rollup/pluginutils` depends on an outdated version of `@types/estree`

thepassle (approved):

PR #1881, opened 2020-10-15
feat(eslint-plugin-lit-a11y): add click-events-have-key-events whitelist

### LINKED ISSUE #1878: [eslint-plugin-lit-a11y] add click-events-have-key-events whitelist
The following code is valid:
`<button @click=${foo}>click</button>`

But the following is not:
`<ing-button @click=${foo}>click</ing-button>`

We should add a allowList to allow button-like custom-elements via options. We can just add the check here:

https://github.com/open-wc/open-wc/blob/34ef9e392b8c3a21dc75c809d8bcdafab7aa885b/packages/eslint-plugin-lit-a11y/lib/rules/click-events-have-key-events.js#L61

Something like:

```js
allowList.includes(element.name)
```

You can find an example of how to add configurable options for a rule here:

https://github.com/open-wc/open-wc/blob/34ef9e392b8c3a21dc75c809d8bcdafab7aa885b/packages/eslint-plugin-lit-a11y/lib/rules/img-redundant-alt.js#L38

```js
schema: [
      {
        keywords: {
          type: 'array',
          items: {
            type: 'string',
          },
          uniqueItems: true,
          additionalItems: false,
        },
      },
    ],
```

trystan2k:
Hi,

I believe this fix #1878 

Please let me know if you see anything missing or incorrect !

Thank you !

### DIFF
--- a/packages/eslint-plugin-lit-a11y/lib/rules/click-events-have-key-events.js
+++ b/packages/eslint-plugin-lit-a11y/lib/rules/click-events-have-key-events.js
@@ -24,7 +24,18 @@ const ClickEventsHaveKeyEventsRule = {
       recommended: false,
     },
     fixable: null,
-    schema: [],
+    schema: [
+      {
+        allowList: {
+          type: 'array',
+          items: {
+            type: 'string',
+          },
+          uniqueItems: true,
+          additionalItems: false,
+        },
+      },
+    ],
   },
 
   create(context) {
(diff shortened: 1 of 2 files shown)


thepassle (approved):

thepassle:
You're insane 😄 That was so fast, thanks!

trystan2k:
You put it in a way that was very easy, 90% was already done by your description 😁

PR #1883, opened 2020-10-15
feat(eslint-plugin-lit-a11y): add heading-has-content function calls

### LINKED ISSUE #1879: [eslint-plugin-lit-a11y] heading-has-content support function calls
The following code is valid code, but currently fails the linting rule:

```js
html`<h1>${foo()}</h1>`
```

This, however, does pass:
```js
html`<h1>${foo}</h1>`
```

Gotta figure out why the former fails, and fix it

You can add a failing test here:

https://github.com/open-wc/open-wc/blob/34ef9e392b8c3a21dc75c809d8bcdafab7aa885b/packages/eslint-plugin-lit-a11y/tests/lib/rules/heading-has-content.js#L24

By adding the following line:

```js
    { code: "html`<h2>${foo()}</h2>`" },
```

Then for a smooth debug experience you can run: 
```bash
mocha tests/lib/rules/heading-has-content.js --watch
```

### LINKED ISSUE #1880: [eslint-plugin-lit-a11y] aria-attr-valid-value support function calls
Seems to be a similar problem as https://github.com/open-wc/open-wc/issues/1879, see that issue for more info

Failing test:
```js
    { code: 'html`<div aria-label="${foo("foo")}"></div>`' },
```

trystan2k:
Hi again :D

This one should fix #1879 

Please let me know if you see any issue !

Thanks !


trystan2k:
I just realize that this also closes #1880 as it share the same `utils.js` file. 
I added a new commit with the test from the issue.

packages/eslint-plugin-lit-a11y/tests/lib/rules/aria-attr-valid-value.js
```diff
@@ -109,6 +109,9 @@ ruleTester.run('aria-attr-valid-value', rule, {
     { code: 'html`<div aria-labelledby="additions additions" />`' },
     { code: 'html`<div aria-labelledby="additions removals text" />`' },
     { code: 'html`<div aria-labelledby="additions removals text all" />`' },
+
+    // FUNCTION CALL
+    { code: 'html`<div aria-label="${foo("foo")}"></div>`' },
```
thepassle, line 114: Could you add a test for:
```js
    { code: 'html`<div aria-label="${this.foo("foo")}"></div>`' },
```

as well? Just to make sure 🙂

thepassle:
I think this should also fix https://github.com/open-wc/open-wc/issues/1880

Could you add the (currently) failing test from that issue and see if its solved as well?

packages/eslint-plugin-lit-a11y/tests/lib/rules/aria-attr-valid-value.js
trystan2k: Sure. 
Just added and test passed as well :D

trystan2k:
> I think this should also fix #1880
> 
> Could you add the (currently) failing test from that issue and see if its solved as well?

I think it was the 2 tests I added for `aria-attr-valid-value.js`, right ? I tried to link the PR to 2 issues (not sure if it is doable) and could not. But yes, in theory this PR fix both issues.

packages/eslint-plugin-lit-a11y/tests/lib/rules/heading-has-content.js
```diff
@@ -36,6 +36,12 @@ ruleTester.run('heading-has-content', rule, {
     { code: 'html`<h5>${foo}</h5>`' },
     { code: 'html`<h6>${foo}</h6>`' },
     { code: "html`<h1><div aria-hidden='true'>foo</div> foo</h1>`" },
+    { code: 'html`<h1>${foo()}</h1>`' },
+    { code: 'html`<h1>${foo("hello")}</h1>`' },
+    { code: 'html`<h1>${foo(1)}</h1>`' },
+    { code: 'html`<h1>${foo(true)}</h1>`' },
+    { code: 'html`<h1>${foo(bar)}</h1>`' },
+    { code: 'html`<h1>${foo(bar, "hello", 1, true)}</h1>`' },
```
thepassle, line 44: could you add one for `this.foo()` here as well?

thepassle:
> I think it was the 2 tests I added for `aria-attr-valid-value.js`, right ? I tried to link the PR to 2 issues (not sure if it is doable) and could not. But yes, in theory this PR fix both issues.

Ah yes, my bad, looked too fast 🙂 Added one more comment 👍

packages/eslint-plugin-lit-a11y/tests/lib/rules/heading-has-content.js
trystan2k: Done !

trystan2k:
> > I think it was the 2 tests I added for `aria-attr-valid-value.js`, right ? I tried to link the PR to 2 issues (not sure if it is doable) and could not. But yes, in theory this PR fix both issues.
> 
> Ah yes, my bad, looked too fast 🙂 Added one more comment 👍

No problem !

packages/eslint-plugin-lit-a11y/tests/lib/rules/heading-has-content.js
trystan2k: Sorry, just saw I did not done correctly. Just pushed it correctly now

thepassle (approved):

thepassle:
fixes https://github.com/open-wc/open-wc/issues/1880
fixes https://github.com/open-wc/open-wc/issues/1879

thepassle:
Thanks again for the great work 🙂

PR #1890, opened 2020-10-16
fix(eslint-plugin-lit-a11y): support member and conditional expressions

### LINKED ISSUE #1889: [eslint-plugin-lit-a11y] aria-attr-valid-value fails on this.foo
This passes:
```js
    { code: 'html`<div aria-valuemax=${foo} />`' },
```

This fails, but should pass:
```js
    { code: 'html`<input aria-valuemax=${this.max} />`' },
```

I think something is not going right with the placeholder somehow, because the expression `${this.max}` should be substituted with a placeholder starting with `{{`, and be caught here:

https://github.com/open-wc/open-wc/blob/7ac9b273f6b97ac2cdf8f12972c08a38ec767808/packages/eslint-plugin-lit-a11y/lib/rules/aria-attr-valid-value.js#L125

I _think_ it might be a missing case that we dont cover here, but im not sure:
https://github.com/open-wc/open-wc/blob/7ac9b273f6b97ac2cdf8f12972c08a38ec767808/packages/eslint-plugin-lit-a11y/template-analyzer/util.js#L219

ASTExplorer is your friend 🙂 
https://astexplorer.net/#/gist/e7e7299d93a75e395402cdb42d40b97c/2e6ec797100db941ad68a04c59a342b676b7b150

**EDIT:** The same seems to be the case for ternaries in expressions, e.g.: `<input aria-disabled=${this.disabled ? 'true' : 'false'} />` I'm assuming the same problem is the case there.

Westbrook:
fixes #1889

Maybe just "refs" #1889, but this allows for the following syntaxes to lint:
```
<input aria-valuemax=${this.max} />
<input aria-disabeld=${this.disabled ? 'true' : 'false'} />
```
Currently is replaces it with a placeholder, but I can see room (and reason) to want to follow these tests/properties down their path to ensure that the actual typings of the binding that they represent are achieved.

### DIFF
--- a/packages/eslint-plugin-lit-a11y/template-analyzer/util.js
+++ b/packages/eslint-plugin-lit-a11y/template-analyzer/util.js
@@ -241,13 +241,9 @@ function templateExpressionToHtml(node) {
     } else if (isCallExpression(expr)) {
       html += `{{{${getIdentifierName(expr.callee)}(${getCallExprValue(expr)})}}}`;
     } else if (isMemberExpressions(expr)) {
-      html += `{{{${expr.property.name}}}}`;
+      html += `{{}}`;
     } else if (isConditionalExpression(expr)) {
-      if (isMemberExpressions(expr.test.left)) {
-        html += `{{{${expr.test.left.property.name}}}}`;
-      } else {
-        html += `{{}}`;
-      }
+      html += `{{}}`;
     }
   }
   return html;
--- a/packages/eslint-plugin-lit-a11y/template-analyzer/util.js
+++ b/packages/eslint-plugin-lit-a11y/template-analyzer/util.js
@@ -241,9 +241,9 @@ function templateExpressionToHtml(node) {
     } else if (isCallExpression(expr)) {
       html += `{{{${getIdentifierName(expr.callee)}(${getCallExprValue(expr)})}}}`;
     } else if (isMemberExpressions(expr)) {
-      html += `{{}}`;
+      html += `{{{}}}`;
     } else if (isConditionalExpression(expr)) {
-      html += `{{}}`;
+      html += `{{{}}}`;
     }
   }
   return html;
--- a/packages/eslint-plugin-lit-a11y/template-analyzer/util.js
+++ b/packages/eslint-plugin-lit-a11y/template-analyzer/util.js
@@ -187,6 +187,22 @@ function isLiteral(expr) {
 
 /**
  * @param {import("estree").Expression} expr
+ * @return {expr is import("estree").MemberExpression}
+ */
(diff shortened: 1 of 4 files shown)


thepassle:
Looks good 🙂 Would be great (as discussed on slack) if you could add some tests for these 👍

Westbrook:
I'm gonna place a few more updates in line of the PR to keep my review today clean and make keeping track of the versions a little easier on my end.

thepassle:
> I'm gonna place a few more updates in line of the PR to keep my review today clean and make keeping track of the versions a little easier on my end.

Sure 👍 let me know when you're ready for a review 🙂

Westbrook:
OK, these two are the clear fixes that I need. Most others look like settings on my end...happy to get this moving forward and possibly included in a release if that were on the list for today.

thepassle:
Looks like some type errors are blocking the pipeline ☹️

image

packages/eslint-plugin-lit-a11y/template-analyzer/util.js
```diff
@@ -224,6 +240,10 @@ function templateExpressionToHtml(node) {
       html += `{{${expr.value}}}`;
     } else if (isCallExpression(expr)) {
       html += `{{{${getIdentifierName(expr.callee)}(${getCallExprValue(expr)})}}}`;
+    } else if (isMemberExpressions(expr)) {
+      html += `{{}}`;
+    } else if (isConditionalExpression(expr)) {
+      html += `{{}}`;
```
Westbrook, line 246: These two clear the type issues, but I'm not sure if there _should_ be something in the placeholders?

thepassle:
Looks like the test is still failing, I'll take a look at this tomorrow and try to fix it 🙂

packages/eslint-plugin-lit-a11y/template-analyzer/util.js
```diff
@@ -224,6 +240,10 @@ function templateExpressionToHtml(node) {
       html += `{{${expr.value}}}`;
     } else if (isCallExpression(expr)) {
       html += `{{{${getIdentifierName(expr.callee)}(${getCallExprValue(expr)})}}}`;
+    } else if (isMemberExpressions(expr)) {
+      html += `{{}}`;
```
thepassle, line 244:
```suggestion
      html += `{{{}}}`;
```

packages/eslint-plugin-lit-a11y/template-analyzer/util.js
thepassle, line 246:
```suggestion
      html += `{{{}}}`;
```

thepassle (approved):

PR #1893, opened 2020-10-16
feat(eslint-plugin-lit-a11y): interactive-supports-focus

marcoswebdev:
Second strike after https://github.com/open-wc/open-wc/pull/1869

interactive-supports-focus rule
#1781 

Example: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/interactive-supports-focus.md

### DIFF
--- a/packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
+++ b/packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
@@ -1,0 +1,142 @@
+/**
+ * @fileoverview
+ * @author open-wc
+ */
+
+const { TemplateAnalyzer } = require('../../template-analyzer/template-analyzer.js');
+const { eventHandlersByType } = require('../utils/eventsHandlers.js');
+const { getAttrVal } = require('../utils/getAttrVal.js');
+const { getTabIndex } = require('../utils/getTabIndex.js');
+const { isDisabledElement } = require('../utils/isDisabledElement.js');
+const { isHiddenFromScreenReader } = require('../utils/isHiddenFromScreenReader.js');
+const {
+  isInteractiveElement,
+  isNonInteractiveElement,
+} = require('../utils/isInteractiveElement.js');
+const { isInteractiveRole, isNonInteractiveRole } = require('../utils/isInteractiveRole.js');
+const { isHtmlTaggedTemplate } = require('../utils/isLitHtmlTemplate.js');
+const { isPresentationRole } = require('../utils/isPresentationRole.js');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Elements with an interactive role and interaction handlers (mouse or key press) must be focusable.',
+      category: 'Fill me in',
+      recommended: false,
+    },
+    messages: {
+      elementWithRoleMustBe: "Elements with the '{{role}}' interactive role must be {{type}}.",
+    },
+    fixable: null, // or "code" or "whitespace"
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          tabbable: {
+            type: 'array',
+            items: {
+              type: 'string',
+            },
+            uniqueItems: true,
+            additionalItems: false,
+          },
+        },
+        required: ['tabbable'],
+        additionalProperties: false,
+      },
+    ],
+  },
+
+  create(context) {
+    // variables should be defined here
+
+    //----------------------------------------------------------------------
+    // Helpers
+    //----------------------------------------------------------------------
+
+    const tabbable = (context.options && context.options[0] && context.options[0].tabbable) || [];
+    const interactiveProps = [...eventHandlersByType.mouse, ...eventHandlersByType.keyboard];
+
+    //----------------------------------------------------------------------
+    // Public
+    //----------------------------------------------------------------------
+
+    return {
+      TaggedTemplateExpression(node) {
+        // 1. only target html`` tagged template literals
+        if (isHtmlTaggedTemplate(node)) {
+          const analyzer = TemplateAnalyzer.create(node);
+
+          analyzer.traverse({
+            enterElement(element) {
+              // TODO: review this statement
+              // if (!includes(domElements, type)) {
+              //   // Do not test higher level JSX components, as we do not know what
+              //   // low-level DOM element this maps to.
+              //   return;
+              // }
+
+              const { attribs, name } = element;
+
+              const hasInteractiveProps = Object.keys(attribs).some(attr =>
+                interactiveProps.includes(attr),
+              );
+
+              if (
+                !hasInteractiveProps ||
+                isDisabledElement(attribs) ||
+                isHiddenFromScreenReader(name, attribs) ||
+                isPresentationRole(attribs)
+              ) {
+                // Presentation is an intentional signal from the author that this
+                // element is not meant to be perceivable. For example, a click screen
+                // to close a dialog.
+                return;
+              }
+
+              const hasTabindex = getTabIndex(getAttrVal(attribs.tabindex)) !== undefined;
+
+              if (
+                hasInteractiveProps &&
+                isInteractiveRole(attribs) &&
+                !isInteractiveElement(name, attribs) &&
+                !isNonInteractiveElement(name, attribs) &&
+                !isNonInteractiveRole(name, attribs) &&
+                !hasTabindex
+              ) {
+                const role = getAttrVal(attribs.role);
+
+                if (tabbable.includes(role)) {
+                  // Always tabbable, tabIndex = 0
+                  context.report({
+                    node,
+                    messageId: 'elementWithRoleMustBe',
+                    data: {
+                      role,
+                      type: 'tabbable',
+                    },
+                  });
+                } else {
+                  // Focusable, tabIndex = -1 or 0
+                  context.report({
+                    node,
+                    messageId: 'elementWithRoleMustBe',
+                    data: {
+                      role,
+                      type: 'focusable',
+                    },
+                  });
+                }
+              }
+            },
+          });
+        }
+      },
+    };
+  },
+};
--- a/packages/eslint-plugin-lit-a11y/lib/utils/getTabIndex.js
+++ b/packages/eslint-plugin-lit-a11y/lib/utils/getTabIndex.js
@@ -1,0 +1,27 @@
+const getTabIndex = tabIndexAttrValue => {
+  // String and number values.
+  if (['string', 'number'].includes(tabIndexAttrValue)) {
+    // Empty string will convert to zero, so check for it explicity.
+    if (typeof tabIndexAttrValue === 'string' && tabIndexAttrValue.length === 0) {
+      return undefined;
+    }
+
+    const value = Number(tabIndexAttrValue);
+    if (Number.isNaN(value)) {
+      return undefined;
+    }
+
+    return Number.isInteger(value) ? value : undefined;
+  }
+
+  // Booleans are not valid values, return undefined.
+  if (tabIndexAttrValue === true || tabIndexAttrValue === false) {
+    return undefined;
+  }
+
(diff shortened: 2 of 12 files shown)


marcoswebdev:
I added basic test and I don't know if it makes sense to add more like in https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/__tests__/src/rules/interactive-supports-focus-test.js#L199, where they are using a reducer to test more scenarios:

```js
ruleTester.run(`${ruleName}:recommended`, rule, {
  valid: [
    ...alwaysValid,
    ...passReducer(
      interactiveRoles,
      eventHandlers.filter((handler) => !includes(triggeringHandlers, handler)),
      codeTemplate,
    ),
    ...passReducer(
      interactiveRoles.filter((role) => !includes(recommendedRoles, role)),
      eventHandlers.filter((handler) => includes(triggeringHandlers, handler)),
      tabindexTemplate,
    ),
  ]
    .map(ruleOptionsMapperFactory(recommendedOptions))
    .map(parserOptionsMapper),
  invalid: [
    ...failReducer(recommendedRoles, triggeringHandlers, tabbableTemplate),
    ...failReducer(
      interactiveRoles.filter((role) => !includes(recommendedRoles, role)),
      triggeringHandlers,
      focusableTemplate,
    ),
  ]
    .map(ruleOptionsMapperFactory(recommendedOptions))
    .map(parserOptionsMapper),
});
```

What I couldn't replicate was the template generation using this function:

```js
function template(strings, ...keys) {
  return (...values) => keys.reduce(
    (acc, k, i) => acc + (values[k] || '') + strings[i + 1],
    strings[0],
  );
}

const codeTemplate = template`<${0} role="${1}" ${2}={() => void 0} />`;
```

Because with the litHtml template it enters in conflict, something like this doesn't work:

```js
 const codeTemplate = template`html`<${0} role="${1}" ${2}={() => void 0} />``;
````
or
```js
 const codeTemplate = template`'html`<${0} role="${1}" ${2}={() => void 0} />`'`;
```

what do you think?

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,129 @@
...
+  { code: 'html`<select @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<area href="#" @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<area @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<summary @click=${() => void 0} />`' },
+  { code: 'html`<textarea @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<a @click="${showNextPage()}">Next page</a>`' },
+  { code: 'html`<a @click="${showNextPage()}" tabIndex=${undefined}>Next page</a>`' },
+  { code: 'html`<a @click="${showNextPage()}" tabIndex="bad">Next page</a>`' },
+  { code: 'html`<a @click=${() => void 0} />`' },
+  { code: 'html`<a tabIndex="0" @click=${() => void 0} />`' },
```
thepassle, line 46: `tabIndex` should be `tabindex`

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,128 @@
...
+        'Elements with an interactive role and interaction handlers (mouse or key press) must be focusable.',
+      category: 'Fill me in',
+      recommended: false,
+    },
+    messages: {
+      elementWithRoleMustBe: "Elements with the '{{role}}' interactive role must be {{type}}.",
+    },
+    fixable: null, // or "code" or "whitespace"
+    schema: [
+      // fill in your schema
```
thepassle, line 37: the schema for `tabbable` array should be in here, otherwise line#48 won't do anything

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,128 @@
...
+                !isNonInteractiveRole(name, attribs) &&
+                !hasTabindex
+              ) {
+                const role = getAttrVal(attribs.role);
+
+                if (tabbable.includes(role)) {
+                  // Always tabbable, tabIndex = 0
+                  context.report({
+                    node,
+                    messageId: 'elementWithRoleMustBe',
```
thepassle, line 104: How will this report in the IDE? what does the error message look like? 🤔

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,128 @@
...
+
+    return {
+      TaggedTemplateExpression(node) {
+        // 1. only target html`` tagged template literals
+        if (isHtmlTaggedTemplate(node)) {
+          const analyzer = TemplateAnalyzer.create(node);
+
+          analyzer.traverse({
+            enterElement(element) {
+              // TODO: review this statement
```
thepassle, line 63: unused comment

packages/eslint-plugin-lit-a11y/lib/utils/isInteractiveRole.js
```diff
@@ -0,0 +1,94 @@
...
+ * role. It will return false for elements that do not have a role. So whereas
+ * a `div` might be considered non-interactive, for the purpose of this utility,
+ * it is considered neither interactive nor non-interactive -- a determination
+ * cannot be made in this case and false is returned.
+ */
+
+const isNonInteractiveRole = (tagName, attributes) => {
+  // Do not test higher level JSX components, as we do not know what
+  // low-level DOM element this maps to.
+  if (!dom.has(tagName)) {
```
thepassle, line 69: we shouldnt need this check for custom-elements right? How would this rule test against custom elements? 🤔

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,129 @@
...
+  { code: 'html`<input @click=${() => void 0} />`' },
+  { code: 'html`<input @click=${() => void 0} role="combobox" />`' },
+  { code: 'html`<button @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<option @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<select @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<area href="#" @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<area @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<summary @click=${() => void 0} />`' },
+  { code: 'html`<textarea @click=${() => void 0} className="foo" />`' },
+  { code: 'html`<a @click="${showNextPage()}">Next page</a>`' },
```
thepassle, line 42: Im confused about some of these 'valid' cases, for example an anchor without an href is not focusable, so this one should fail, right?

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
```diff
@@ -0,0 +1,129 @@
...
+  { code: 'html`<div role="option" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="radio" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="spinbutton" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="switch" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="tablist" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="tab" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="textbox" tabIndex="0" @click=${() => void 0} />`' },
+  { code: 'html`<div role="textbox" aria-disabled="true" @click=${() => void 0} />`' },
+  { code: 'html`<foo-bar @click=${() => void 0} aria-hidden=${false}><foo-bar/>`' },
+  { code: 'html`<Input @click=${() => void 0} type="hidden" />`' },
```
thepassle, line 85: this `Input` looks like a react component 🙂

packages/eslint-plugin-lit-a11y/lib/utils/isInteractiveElement.js
```diff
@@ -124,6 +124,75 @@ const isInteractiveElement = (tagName, attributes) => {
...
+ * is an element with an inherently non-interactive role, then this utility
+ * returns true. Elements that lack either an explicitly assigned role or
+ * an inherent role are not considered. For those, this utility returns false
+ * because a positive determination of interactiveness cannot be determined.
+ */
+const isNonInteractiveElement = (tagName, attributes) => {
+  // Do not test higher level JSX components, as we do not know what
+  // low-level DOM element this maps to.
+  if (!domKeys.includes(tagName)) {
+    return false;
```
thepassle, line 182: is this check needed for custom-elements?

thepassle:
Also need to add the rule in the packages main `README.md`

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
bennypowers: are we also checking `.tabIndex`?

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
bennypowers: see line 33
```js
    messages: {
      elementWithRoleMustBe: "Elements with the '{{role}}' interactive role must be {{type}}.",
    },
```

packages/eslint-plugin-lit-a11y/lib/utils/getTabIndex.js
```diff
@@ -0,0 +1,26 @@
+const getTabIndex = tabIndexAttrValue => {
+  // String and number values.
+  if (['string', 'number'].indexOf(typeof tabIndexAttrValue) > -1) {
```
bennypowers, line 3: nitpick: prefer Array#includes

marcoswebdev: Done! I like that nitpick, thanks!

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
marcoswebdev: 🙈

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
marcoswebdev: Me too, I need to review these scenarios!

packages/eslint-plugin-lit-a11y/lib/rules/interactive-supports-focus.js
marcoswebdev: Added! I don't know if it needs to be required or not?

packages/eslint-plugin-lit-a11y/tests/lib/rules/interactive-supports-focus.js
thepassle: I dont think we are but we probably should... we might need to check this in the other rules, too

bennypowers:
@marcoswebdev we made some progress in polishing the code base recently, try rebasing against master and check out some of the existing rules for patterns you can employ here.

marcoswebdev:
I'm a little bit stuck here. Using the plugin in some projects I saw that it already lint errors related to bad usages of interactive elements. So maybe this rule is covered with the ones in place. What do you think @thepassle and @bennypowers?

bennypowers:
@marcoswebdev it is similar to other rules, but covers different things, so it's still worth implementing

PR #1899, opened 2020-10-17
fix(eslint-plugin-lit-a11y): aria-label img role

bennypowers:
allows
```js
html`
  <div role="img" aria-label="foo"></div>
  <div role="img" aria-labelledby="label"></div>
`;
```
also allows
```js
html`
  <span role="img" alt="Panda">🐼</span>
`
```

### DIFF
--- a/packages/eslint-plugin-lit-a11y/docs/rules/accessible-emoji.md
+++ b/packages/eslint-plugin-lit-a11y/docs/rules/accessible-emoji.md
@@ -1,8 +1,10 @@
-# Enforce emojis are wrapped in <span> and provide screenreader access. (accessible-emoji)
+# Enforce emojis are wrapped in <span> and provide screenreader access. (accessible-emoji) (DEPRECATED)
 
 > Emoji help us communicate complex ideas very easily. When used in native apps and applications, emoji are reasonably accessible to screen readers, but on the web we need to do a little more to make sure everyone can understand emoji.
 
 - [Léonie Watson](https://tink.uk/accessible-emoji/)
+
+While many modern user agents are able to announce emoji to screen reader users, this rule is still useful for apps targetting older user agents.
 
 ## Rule Details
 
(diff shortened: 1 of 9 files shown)


bennypowers:
i'm not sure about whether `<span role="img" alt="panda">` is ok or not

erikkroes:
>I don't think an alt-attribute on a span does anything at all.

> (And I'm pretty sure this isn't relevant, but actua11y: an alt-text for emoji's should not be needed. )
https://twitter.com/erikkroes/status/1317535014783995907

bennypowers:
Maybe we should remove accessible-emoji from recommended rules

alt-text should maybe also check the element tag name

thepassle:
> Maybe we should remove accessible-emoji from recommended rules

yep, sounds fine 👍 

> alt-text should maybe also check the element tag name

it already does?

bennypowers:
`alt-text` checks for img tag as well as img role

-----

Lots of useful discussion here

https://twitter.com/PowersBenny/status/1317522603834331136?s=20

bennypowers:
Updates:

- refactored `accessible-emoji` and `alt-text`, and updated docs and test for both
- disabled `accessible-emoji` in `plugin:lit-a11y/recommended`. I think it's worth keeping around as an opt-in
- permit `` and `<div role="img" aria-hidden="true"></div>` in `alt-text`

packages/eslint-plugin-lit-a11y/docs/rules/accessible-emoji.md
```diff
@@ -4,6 +4,8 @@
 
 - [Léonie Watson](https://tink.uk/accessible-emoji/)
 
+While many modern user agents are able to announce emoji to screen reader users, this rule is still useful for apps targetting older user agents.
```
thepassle, line 7: I think we should at least also add the Deprecated warning in the top heading if we're gonna keep it around, although Im still in favor of just getting rid of it :p

bennypowers: added deprecated flag and header text

thepassle (approved):

Issue #1904, opened 2020-10-17
[eslint-plugin-lit-a11y] detect lit-html import


thepassle:
We should add a check to each rule to see if `html` is indeed imported from lit-html or lit-element. Different rendering libraries may use a `html` tagged template literal as well, and may have a differing syntax than lit-html, we dont want to clash with that.

We can look at `ImportDeclaration`s in the visitor (similar to how we declare the `TaggedTemplateExpression` function), and then check if one of the `ImportSpecifiers` is `html`, and if the `source` is lit-html or lit-element

We can keep track of a `isLitHtml` variable in the scope of the `create` function, and then do something like:

```js
  create(context) {
    let isLitHtml = false;
    return {
      ImportDeclaration(node) {/* check if import is lit-html, set isLitHtml to true */}
      TaggedTemplateExpression(node) {
        if (isHtmlTaggedTemplate(node) && isLitHtml) {/* implement rule */}
      }
    }
  }
```

We should also make the source configurable; packages may re-export `html`, e.g.:
`import {html} from 'ing-web';`. We should be able to use this for that: https://eslint.org/docs/user-guide/configuring#adding-shared-settings and reference it from the `context` object passed down by the `create` function

This check should be added to _all_ rules.

AST explorer is your friend
https://astexplorer.net/#/gist/12d236f9f40aee114413e87e2a564b6b/ff4653d065bfd3f51e3e9bafd0254494e38349b8

One potential problem here could be the tests. We'd need to add the lit-html import to every single test fixture... maybe we can find a better way to mock it somehow. But its something to consider

bennypowers:
What about reexports? `@open-wc/testing` does that, so does `@apollo-elements/lit-apollo`

Better to trust the user, I think. They can use `overrides` of there using two versions of `html`

thepassle:
Reexports are also accounted for in the issue

PR #1908, opened 2020-10-18
feat: detect lit-html

thepassle:
Only run rules if lit-html is imported from a valid lit-html source. Default valid sources are: `['lit-html', 'lit-element']`, and additional valid lit-html sources (in case a package re-exports lit-html) can be configured via

```js
settings: {
  litHtmlSources: ['ing-web']
}
```

closes https://github.com/open-wc/open-wc/issues/1904

### DIFF
--- a/packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
+++ b/packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
@@ -54,7 +54,7 @@ const AccessibleEmojiRule = {
 
     return {
       ImportDeclaration(node) {
-        if(hasLitHtmlImport(node, validLitHtmlSources)) {
+        if (hasLitHtmlImport(node, validLitHtmlSources)) {
           isLitHtml = true;
         }
       },
--- a/packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
+++ b/packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
@@ -8,6 +8,7 @@ const { TemplateAnalyzer } = require('../../template-analyzer/template-analyzer.
 const { isTextNode } = require('../../template-analyzer/util.js');
 const { isHiddenFromScreenReader } = require('../utils/isHiddenFromScreenReader.js');
 const { isHtmlTaggedTemplate } = require('../utils/isLitHtmlTemplate.js');
+const { hasLitHtmlImport, createValidLitHtmlSources } = require('../utils/utils.js');
 
 //------------------------------------------------------------------------------
 // Rule Definition
@@ -27,6 +28,9 @@ const AccessibleEmojiRule = {
   },
 
   create(context) {
+    let isLitHtml = false;
+    const validLitHtmlSources = createValidLitHtmlSources(context);
+
     /**
      * @param {import('parse5-htmlparser2-tree-adapter').Element} element
      * @return {boolean}
@@ -44,8 +48,13 @@ const AccessibleEmojiRule = {
     }
 
     return {
+      ImportDeclaration(node) {
+        if(hasLitHtmlImport(node, validLitHtmlSources)) {
+          isLitHtml = true;
+        }
+      },
       TaggedTemplateExpression(node) {
-        if (isHtmlTaggedTemplate(node)) {
+        if (isHtmlTaggedTemplate(node) && isLitHtml) {
           const analyzer = TemplateAnalyzer.create(node);
 
           analyzer.traverse({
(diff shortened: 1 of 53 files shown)


packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
```diff
@@ -49,8 +53,13 @@ const AccessibleEmojiRule = {
     }
 
     return {
+      ImportDeclaration(node) {
```
bennypowers, line 56: can we implement this once and extend for rules, instead of copying?

bennypowers: maybe on `context`? https://eslint.org/docs/developer-guide/working-with-rules#the-context-object

packages/eslint-plugin-lit-a11y/tests/lib/rules/accessible-emoji.js
```diff
@@ -68,5 +69,5 @@ ruleTester.run('accessible-emoji', rule, {
       code: 'html`<span role="img" alt="Panda face">🐼</span>`',
       errors: [{ messageId: 'wrapEmoji' }],
     },
-  ],
+  ].map(prependLitHtmlImport),
```
bennypowers, line 72: can we apply settings instead of mapping over snippets?
not sure this is possible: https://eslint.org/docs/developer-guide/nodejs-api#ruletester

bennypowers (changes requested):
general comment: i'm not sure the rule name explains itself well enough.

like, we're not adding an a11y rule that "you should import html", right?
instead, it's a setting on all rules that checks whether or not the rule applies.

In that vein, I wonder if we can just make this a feature of each rule (see above re: extending) rather than its own rule

thepassle:
> general comment: i'm not sure the rule name explains itself well enough.

Thats because its not a rule. It just detects if lit-html is imported from a valid lit-html source, to make sure we only test rules on valid lit-html templates

packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
thepassle: We cant, because for each file we need to find the lit-html ImportDeclaration

packages/eslint-plugin-lit-a11y/tests/lib/rules/accessible-emoji.js
thepassle: I tried to find something like that, but I couldnt find it, so mapping seemed like the most DRY solution here (even though its repeated in all the tests :p )

packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
bennypowers: but surely we can just define this code elsewhere and import it

at the very least
```js
create(context) {
  return {
    ...checkLitHtmlImport,
   TaggedTemplateExpression(...)
  }
}
```

packages/eslint-plugin-lit-a11y/tests/lib/rules/accessible-emoji.js
bennypowers: it's a maintenance nightmare tho

bennypowers:
> > general comment: i'm not sure the rule name explains itself well enough.
> 
> Thats because its not a rule. It just detects if lit-html is imported from a valid lit-html source, to make sure we only test rules on actual or 'valid' lit-html templates

right, that's exactly why I think it _shouldn't_ be a rule, if we can avoid it

thepassle:
> right, that's exactly why I think it shouldn't be a rule, if we can avoid it

Its not a rule

packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
thepassle: I wanted to do something like that, but it gets messy because we need the `isLitHtml` variable from the scope of `create`

bennypowers:
> > right, that's exactly why I think it shouldn't be a rule, if we can avoid it
> 
> Its not a rule

🤦‍♂️ my bad

packages/eslint-plugin-lit-a11y/tests/lib/rules/accessible-emoji.js
thepassle: Just went through the docs again, and through some of the jsx plugins code as well, but I really dont see a way to simplify this further

bennypowers (approved):

Issue #1913, opened 2020-10-19
[eslint-plugin-lit-a11y] support checking Literal values inside expressions


thepassle:
Related to the discussion [here](https://github.com/open-wc/open-wc/issues/1912#issuecomment-711837381)

It would be nice to be able to check `Literal` values (only `string`s, `number`s, `boolean`s) in expressions, so we can lint templates like: 
```js
html`<div .tabIndex=${1}></div>`
``` 

Currently, all expressions in a template are replaced by a placeholder that looks something like `{{__Q:0__}}`, and as such are not linted.

Replacement of the expression with a placeholder happens here:
https://github.com/open-wc/open-wc/blob/73c500197e1403ac7c4ec78412f922e52a58c696/packages/eslint-plugin-lit-a11y/template-analyzer/util.js#L168

And here:
https://github.com/open-wc/open-wc/blob/73c500197e1403ac7c4ec78412f922e52a58c696/packages/eslint-plugin-lit-a11y/template-analyzer/util.js#L131

The problem with this approach is that using something else than the `{{__Q:0__}}` placeholder is that it throws off the reporting locations in your IDE. 
![image](https://user-images.githubusercontent.com/17054057/96427610-10231780-11ff-11eb-98cb-fca83ca0ed10.png)

Also adding values to the placeholder, and extracting them again from the placeholder when we want to lint them. We had something similar to this before, but it has been reverted. Here is the reverting PR for reference:  https://github.com/open-wc/open-wc/pull/1907/files

<hr/>

Example of what should be supported here:
https://astexplorer.net/#/gist/63d7bd2b15b549da721fac2afc2357d2/f3f5c1ea4768b65d2e818ed0e7910d4fbb7a801f

If we get something succesfully working, we should also see if we can upstream and contribute this to [eslint-plugin-lit](https://github.com/43081j/eslint-plugin-lit/)

thepassle:
I wonder if it would be possible to keep an internal map of placeholders and their values in the `TemplateAnalyzer` when replacing the expression 🤔 That way we dont have to change the placeholder value itself, also the placeholder should be unique to a node.

something like:
`util.js`:

```diff
+/**
+ * @param {import("estree").Expression} expr
+ * @return {expr is import("estree").SimpleLiteral}
+ */
+function isLiteral(expr) {
+  return expr && expr.type === 'Literal';
+}

function templateExpressionToHtml(node) {
  let html = '';

[...]

PR #1915, opened 2020-10-19
Feat/check literal values

thepassle:
fixes https://github.com/open-wc/open-wc/issues/1913


thepassle:
@bennypowers any idea how to fix this? 
image

packages/eslint-plugin-lit-a11y/lib/rules/aria-role.js
```diff
@@ -50,21 +51,23 @@ const AriaRoleRule = {
                   return;
                 }
 
-                if (rawValue.startsWith('{{')) {
+                const val = getExpressionValue(analyzer, rawValue);
+
+                if (!val && rawValue.startsWith('{{')) {
```
bennypowers, line 56: maybe a helper `isInterpolation`?

packages/eslint-plugin-lit-a11y/lib/rules/img-redundant-alt.js
```diff
@@ -39,7 +39,7 @@ const ImgRedundantAltRule = {
     },
     messages: {
       imgRedundantAlt:
-        '<img> alt attribute must be descriptive; it cannot contain the banned {{plural}} {{formatted}}.',
+        '<img> alt attribute must be descriptive; it cannot contain the banned {{plural}} {{banned}}.',
```
bennypowers, line 42: Good catch!

bennypowers (approved):

packages/eslint-plugin-lit-a11y/lib/rules/aria-role.js
thepassle: yeah we can do, id prefer the name `isPlaceholder` though

PR #1921, opened 2020-10-21
chore(eslint-plugin-lit-a11y): remove dupe block

bennypowers:
somehow we ended up importing this polyfill twice


thepassle:
good catch!

thepassle (approved):

Issue #1940, opened 2020-10-23
[eslint-plugin-lit-a11y] role-supports-aria error


dlockhart:
When linting a component, I'm getting the following error:

> TypeError: Cannot destructure property 'props' of 'roles.get(...)' as it is undefined.
Occurred while linting /Users/dlockhart/source/brightspace-ui/core/components/calendar/calendar.js:496
    at Object.enterElement (/Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/lib/rules/role-supports-aria-attr.js:45:32)
    at visit (/Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/template-analyzer/template-analyzer.js:320:17)
    at /Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/template-analyzer/template-analyzer.js:327:13
    at Array.forEach (<anonymous>)
    at visit (/Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/template-analyzer/template-analyzer.js:326:20)
    at /Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/template-analyzer/template-analyzer.js:327:13
    at Array.forEach (<anonymous>)
    at visit (/Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/template-analyzer/template-analyzer.js:326:20)
    at TemplateAnalyzer.traverse (/Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/template-analyzer/template-analyzer.js:335:5)
    at TaggedTemplateExpression (/Users/dlockhart/source/brightspace-ui/core/node_modules/eslint-plugin-lit-a11y/lib/rules/role-supports-aria-attr.js:39:20)

The line it's pointing to is this one:
https://github.com/BrightspaceUI/core/blob/1fab3c130cd6849e95ae578a5f87526dd431eaa8/components/calendar/calendar.js#L496

There are a few `role` usages in that template, but if I remove the `role="${ifDefined(role)}"` one -- where the local variable `role` can be `'dialog'` or `undefined` -- it fixes things.

thepassle:
I'll take a look, thanks for reporting!

thepassle:
[...]

PR #1942, opened 2020-10-23
chore: add some more testcases

thepassle:
Adds some more testcases to make sure we dont run into https://github.com/open-wc/open-wc/issues/1940 anywhere else


bennypowers (approved):
should maybe implement that `TemplateAnalyzer.isPlaceholder`

PR #1944, opened 2020-10-23
WIP: fix(eslint-plugin-lit-a11y): fix required aria-controls

thepassle:


packages/eslint-plugin-lit-a11y/tests/lib/rules/role-has-required-aria-attrs.js
```diff
@@ -30,8 +30,9 @@ ruleTester.run('role-has-required-aria-attrs', rule, {
         'html`<span role="checkbox" aria-checked="false" aria-labelledby="foo" tabindex="0"></span>`',
     },
     { code: 'html`<span role="row"></span>`' },
+    { code: 'html`<span role="${foo}"></span>`' },
+    { code: 'html`<span role="checkbox" aria-checked=${true}></span>`' },
     { code: 'html`<input type="checkbox" role="switch" aria-checked="true" />`' },
-    { code: 'html`<div role="combobox" aria-controls="foo"  aria-expanded="foo"></div>`' },
```
Westbrook, line 34: Why remove what looks like a valid combo box usage? Seems like we’d want to test this...

Westbrook: Can we add this as `<div role="combobox"  aria-expanded="foo"></div>` instead of removing?

PR #1945, opened 2020-10-24
feat(eslint-config): add lit a11y plugin to openwc eslint config

thepassle:


LarsDenBakker:
Should we make this a breaking change?

thepassle:
yeah this should be breaking, but husky stopped me from doing it directly in the commit message:

```
⧗   input: feat(eslint-config): BREAKING CHANGE: add lit-a11y to eslint config
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
```

so I figured i'd do it on merge

LarsDenBakker (approved):
The breaking change goes on the 2nd line in the commit body.

PR #1961, opened 2020-10-29
Update baseLibraries.js

graynorton:
## What I did

1. Replaced the incorrect `stencil` package reference with what I believe is correct: `@stencil/core`


graynorton:
Looks like this has already been addressed!

PR #1965, opened 2020-11-02
Update to going-bundless.md documentation

aam198:
## What I did

1. I have added "will need" to the sentence as the sentence did not make sense but please revise as needed to complete the sentence about generator project.

Thanks!

### DIFF
--- a/docs/guides/developing-components/going-buildless.md
+++ b/docs/guides/developing-components/going-buildless.md
@@ -32,7 +32,7 @@ For example:
 import foo from 'foo';
 ```
 
-Imports such as this are useful when dependencies in your `node_modules` folder need to reference other packages. We this in the generator project as well.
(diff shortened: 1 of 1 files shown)


docs/guides/developing-components/going-buildless.md
```diff
@@ -32,7 +32,7 @@ For example:
 import foo from 'foo';
 ```
 
-Imports such as this are useful when dependencies in your `node_modules` folder need to reference other packages. We this in the generator project as well.
+Imports such as this are useful when dependencies in your `node_modules` folder need to reference other packages. We will need this in the generator project as well.
```
LarsDenBakker, line 35: Thanks. It should actually be like this :)

```suggestion
Imports such as this are useful when dependencies in your `node_modules` folder need to reference other packages. We use this in the generator project as well.
```

LarsDenBakker:
Sorry for the delay here. We need to get the CI passing, would you be able to do a rebase? Or perhaps re-apply the changes?

PR #1966, opened 2020-11-02
feat: define lazy components by element instance instead of by class

manolakis:
## What I did

1. Allows lazy components to be defined by scoped element instances instead of by scoped elements classes.

this change makes scoped elements to be more aligned with the scoped registries proposal.

### DIFF
--- a/packages/scoped-elements/src/transform.js
+++ b/packages/scoped-elements/src/transform.js
@@ -1,4 +1,5 @@
 import { registerElement } from './registerElement.js';
+import { Cache } from './Cache.js';
 
 /**
  * @typedef {import('./types').ScopedElementsMap} ScopedElementsMap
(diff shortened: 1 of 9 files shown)


packages/scoped-elements/src/transform.js
```diff
@@ -21,9 +22,9 @@ const re = new RegExp(`<\\/?([a-z](${chars})*-(${chars})*)`, 'g');
 /**
  * The global cache of processed string arrays
  *
- * @type {Map<TemplateStringsArray, TemplateStringsArray>}
+ * @type {Cache<TemplateStringsArray, TemplateStringsArray>}
  */
-const globalCache = new Map();
+const globalCache = new Cache();
```
LarsDenBakker, line 27: since this is a global, should the cache use a weakmap?

manolakis: Agree, I'm doing the change

manolakis: oh, WeakMap can't be used as it doesn't allow strings as keys and requires objects.

PR #1967, opened 2020-11-02
Adding Application publishing link to developing-components/publishin…

sashafirsov:
…g.md

Component publishing is surfaced before application and when looking for "publishing" or "build" info. Hence I assumed that build is not really applicable in open-wc overall. Which is obviously not true as it is in scope of application publishing. The cross-reference on different approaches to publishing in Component and in Application docs would clear the difference. 
Unfortunately for application I have not found similar page, so used rollup build page for this reference.

## What I did

1. Added link to App publish docs

### DIFF
--- a/docs/guides/developing-components/publishing.md
+++ b/docs/guides/developing-components/publishing.md
@@ -1,5 +1,7 @@
 # Developing Components >> Publishing ||60
 
+Refer for [Application publishing](../../docs/building/rollup.md) guide as it is different from Component. 
+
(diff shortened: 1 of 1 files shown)


docs/guides/developing-components/publishing.md
```diff
@@ -1,5 +1,7 @@
 # Developing Components >> Publishing ||60
 
+Refer for [Application publishing](../../docs/building/rollup.md) guide as it is different from Component.
```
thepassle, line 3:
```suggestion
> Note: This guide is intended for publishing components, not applications. You can find more information about bundling applications [here](../../docs/building/rollup.md).
```

sashafirsov: Good wording, applying

PR #1973, opened 2020-11-07
docs: move storybook docs to site

Westbrook:
## What I did

1. Move the docs to one place so no one makes the mistake I did of updating them in the wrong place.


LarsDenBakker:
Thanks for your contribution

Issue #1979, opened 2020-11-10
Optional chaining in scoped-elements Cache.js breaks webpack 4


michaelwarren1106:
## Expected behavior

Recently a minor update (1.2.4 -> 1.3.0) of @open-wc/scoped-elements introduced optional chaining implementation that breaks all webpack 4 builds.

Imo, this kind of thing should have been a major bump because lots of builds are going to pull in 1.3.0 updates that are still on webpack 4 (vue cli, create-react-app, storybookjs all are still on webpack 4) automatically and then have broken builds.

Webpack 4's version of acorn isn't updated to the latest and won't be, but in lots of places upgrading to webpack 5 is a non-starter. Therefore, webpack cannot support optional chaining, so the current solution has been to back down the version of scoped-elements to 1.2.4, the last working version

IMO, this line:
https://github.com/open-wc/open-wc/blob/master/packages/scoped-elements/src/Cache.js#L22

should be reverted back to non-optional-chained impl and a major bump release instead to re-include it.

LarsDenBakker:
It's strange that webpack doesn't support standard javascript syntax, but we had this issue with import.meta for a long time as well. In this case I think it's fine to rewrite the code to remove the usage, the optional chaining doesn't add that much here.

michaelwarren1106:
Is the version of acorn that webpack uses. Is an older one, webpack 5 I think has an updated version, but it’s acorn that didn’t like the optionals I believe

MathieuPuech:
Webpack issue about this is here: https://github.com/webpack/webpack/issues/10227

The issue is closed because they have fixed it in Webpack 5, they will not backport it to webpack 4.

Webpack 5 support in progress:
- React https://github.com/facebook/create-react-app/pull/7929
- Vue https://github.com/vuejs/vue-cli/pull/5611
- Angular https://github.com/angular/angular-cli/pull/18873
- Storybook https://github.com/storybookjs/storybook/issues/9216

For storybook, here is what you can do, `.storybook/main.js`
```javascript
module.exports = {
  // ...
  webpackFinal: async config => {
    return {
      ...config,
      module: {
        ...config.module,
        rules: [
          ...config.module.rules,
          {
            test: [
              new RegExp(`node_modules(\\/|\\\\)@open-wc(.*)\\.js$`),
            ],
            use: {
              loader: require.resolve('babel-loader'),
              options: {
                plugins: [
                  require.resolve('@babel/plugin-proposal-optional-chaining'),
                ],
[...]

PR #1981, opened 2020-11-11
fix(scoped-elements): refactor to remove optional chain in Cache

vidarc:
## What I did

1. Refactored code to improve tooling support. Optional chaining isn't fully supported yet (namely webpack 4). Though it does look like most tools are "close" to it.
2. Addresses https://github.com/open-wc/open-wc/issues/1979


LarsDenBakker:
Thanks!

PR #1984, opened 2020-11-14
feat(dev-server-hmr): add web component HMR plugin

LarsDenBakker:
## What I did

This adds a plugin for doing hot module replacement with web components. 

It works by detecting classes in your code which implement a specific base class defined in the config. For these base classes when you edit the source code we re-import the module. We get a reference to the updated class and call a static`hotReplaceCallback` on the class if it exists. 

If the base class doesn't implement a `hotReplaceCallback` we default to copying over the properties the new class over to the existing class. 

We also track all the rendered components on the page, and call a `hotReplaceCallback` of the component instance. This should take care of re-rendering or updating the live component.

Base classes should implement these callbacks to do the actual re-rendering, or these callbacks should be patched in externally. For LitElement I added a patch, I'm looking into how we can approach this for other base classes as well.

Right now it only works with class based components. We can look into supporting function and object based components too.

TODO:

- [x] Look into more class based component libs (skate, LWC, FAST)
- [x] Add more babel syntax plugins
- [x] Improve syntax error handling
- [x] Add docs


packages/dev-server-hmr/src/babel/babelPluginWcHmr.js
```diff
@@ -0,0 +1,241 @@
...
+        let injectedRegister = false;
+
+        // collect all base classes we should match without an import
+        for (const b of options.baseClasses) {
+          if (!b.import) {
+            baseClassNames.push(b.name);
+          }
+        }
+
+        program.traverse({
```
43081j, line 162: aren't you triggering a double-traversal here? presumably whatever calls this file as a whole is already traversing the AST (hence the visitor you're returning). so you should instead keep your state outside this `Program` function so you can move your `ImportDeclaration` etc top level.

LarsDenBakker: We have to do two traversals because we first collect all import statements, and then inspect the base classes. Import statements aren't guaranteed to always be on top.

43081j: no but you can visit imports and class declarations/expressions, then have an exit visitor for `Program`:

```ts
Program: {
  exit(node) {
    // by this point, your import and class visitors already ran
  }
}
```

LarsDenBakker: When I'm encountering a class, I already need to know which imports there are in this file.

PR #1989, opened 2020-11-22
docs: ensure that landing on the docs page does not strand mobile users

Westbrook:
## What I did

1. Added a list of top level pages in the Documentation section to prevent mobile entry to the section to look empty.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -7,4 +7,14 @@ eleventyNavigation:
 
 Our documentation is hand crafted and optimized to serve as a dictionary to look up details once the need arises.
 
+Go deeper in the following areas:
+
+- [Development](./development/generator.md)
+- [Testing](./testing/testing-package.md)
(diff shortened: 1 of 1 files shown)


docs/docs/index.md
```diff
@@ -7,4 +7,14 @@ eleventyNavigation:
 
 Our documentation is hand crafted and optimized to serve as a dictionary to look up details once the need arises.
 
+Go deeper in the following areas:
+
+- [Development](./development/generator/)
```
daKmoR, line 12: links should always be to `*.md` files so that they work on github and on the website

Westbrook: Oh, cool. Maybe an even better reason to have it be generated in some way...maybe not at the 11ty level, but pre-commit style so it can be in GitHub, too? 

BUT, updated, thanks!

daKmoR: jup some (pre)processing on files... not sure yet about a good API for it 😅

PR #2006, opened 2020-12-04
chore: add ING to supported by

thepassle:
## What I did

1. Added ING to "Supported By" section

We should also show it on open-wc.org, where should we add it? 🤔 @daKmoR @Westbrook


daKmoR (approved):

daKmoR:
sweet 🤗

yeah probably add to the front page 🤔 
after "Why Open Web Components?" 🤔

or a dedicated page? 🤔

PR #2018, opened 2020-12-19
simplify render method in first example, fix a couple typos

dennishall:
## What I did

1. For the render method illustrating that the click handler won't `exit`, I think the word was meant to be won't `exist`.
2. For the same render method, there seemed to be unnecessary nesting of `html``` calls, so I simplified it.
3. Near the end of the file, there is a typo I fixed: `didispatchEventspatch` -> `dispatchEvent`.


LarsDenBakker:
Thanks for the fixes! Looks like the linter isnt happy about it though.

Issue #2017, opened 2020-12-18
[semantic-dom-diff] Option to normalize adjacent textNodes when separated by comment


kevinpschaaf:
## Expected behavior
Two adjacent textNodes separated only by a comment can be (optionally, if needed) treated as non-semantic.

## Actual Behavior
Currently comments are excluded from the treewalk that generates the diffable HTML, however if two adjacent textNodes are separated by a comment, this is treated as a semantic difference and are printed on separate lines.

Example:
```js
const div = document.createElement('div');
div.innerHTML = `<span>Hello, <!--ignore-me-->world!</span>`;
assert.lightDom.equal(div, `<span>Hello, world!</span>`);
```
```
      + expected - actual
      
       <span>
      -  Hello,
      -  World!
      +  Hello, World!
       </span>
```

## Additional context

Package: `@open-wc/semantic-dom-diff`

kevinpschaaf:
There are cases in `lit-html` where we (need to) leave comment markers around `textNodes` to identify part boundaries, and it can get tedious to treat those as semantic.

LarsDenBakker:
I think we should always merge adjacent text nodes. Since we're filtering out the comments, I don't think you could practically even assert that they are two different text nodes right now.

kevinpschaaf:
Awesome, thank you gentlemen!

PR #2023, opened 2020-12-26
fix(semantic-dom-diff): normalize adjacent textNodes

manolakis:
## What I did

1. Concatenates the sibling text nodes.

Fixes: https://github.com/open-wc/open-wc/issues/2017

### DIFF
--- a/packages/semantic-dom-diff/get-diffable-html.js
+++ b/packages/semantic-dom-diff/get-diffable-html.js
@@ -92,9 +92,25 @@ export function getDiffableHTML(html, options = {}) {
     return '  '.repeat(depth);
   }
 
-  /** @param {Text} textNode */
-  function printText(textNode) {
-    const value = textNode.nodeValue.trim();
+  /**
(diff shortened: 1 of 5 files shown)


packages/semantic-dom-diff/get-diffable-html.js
```diff
@@ -236,11 +236,35 @@ export function getDiffableHTML(html, options = {}) {
...
+
+    return outputText;
+  }
+
   let container;
 
   if (typeof html === 'string') {
     container = document.createElement('diff-container');
-    container.innerHTML = html;
+    container.innerHTML = removeComments(html);
```
LarsDenBakker, line 267: doesn't this only handle the case of input as text? what about input as a node?  perhaps we can handle this while doing the node walk and we encounter a text node?

manolakis: You are right. I changed the implementation concatenating the sibling text nodes.

PR #2027, opened 2020-12-28
docs: blog about upcoming lit-html APIs

Westbrook:
## What I did

1. Blogged about using the upcoming class based directive APIs in lit-html@2.0


Westbrook:
My and the linter have come to terms with each other...for now.

daKmoR (approved):

PR #2034, opened 2021-01-04
fix(lit-helpers): peer dependency versions

bennypowers:
## What I did

1. widened the peerDependencies version constraints for lit-element and lit-html in the `lit-helpers` package

### DIFF
--- a/packages/lit-helpers/package.json
+++ b/packages/lit-helpers/package.json
@@ -30,8 +30,8 @@
     "helpers"
   ],
   "peerDependencies": {
-    "lit-element": "^2",
-    "lit-html": ">= 1 < 2"
+    "lit-element": "^2.0.0",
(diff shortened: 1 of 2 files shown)


packages/lit-helpers/package.json
```diff
@@ -30,8 +30,8 @@
     "helpers"
   ],
   "peerDependencies": {
-    "lit-element": "^1.0.0",
-    "lit-html": "^1.0.0"
+    "lit-element": ">= 1 < 3",
```
daKmoR, line 33: why not just?
```suggestion
    "lit-element": "^2.0.0",
```

as lit-element 1.x is something completely different (pre polymer team - not using lit-html) 🤔

bennypowers: updated

daKmoR (approved):

PR #2041, opened 2021-01-12
chore: generate social media images for every page

daKmoR:
## What I did

1. Generate social media images for every page

This will increase CI time for netlify 🤔 
So maybe we wanna wait until we have this as a github action? 🤔


daKmoR:
superseeded by https://github.com/open-wc/open-wc/pull/2052

PR #2045, opened 2021-01-13
fix(scoped-elements): add constructor to type definition

jorenbroekema:
In short: In order to implement the mixin in a LitElement, which has a constructor, the scoped elements host (base class) also needs a constructor with the same return type (void I guess?). Otherwise TSC will scream at you. 

More details: https://github.com/microsoft/TypeScript/issues/40110

### DIFF
--- a/.changeset/clean-impalas-study.md
+++ b/.changeset/clean-impalas-study.md
@@ -1,0 +1,5 @@
+---
+'@open-wc/scoped-elements': patch
+---
+
+Add constructor to ScopedElementsHost type so that base constructors have the same return type as extensions. See [TypeScript issue](https://github.com/microsoft/TypeScript/issues/40110).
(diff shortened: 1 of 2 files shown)


bennypowers (commented):
I've solved these with @ts-expect-error, since I believe this is a bug:

https://github.com/microsoft/TypeScript/issues/37142

jorenbroekema:
I'm not sure if they are related though. This would fix the error of Base constructors not being seen as the same.

If you do:

```ts
& typeof ScopedElementsHost
```
in the mixin type, you need to ensure that the constructor of ScopedElementsHost has the same return type as the superclass' constructor that you pass into the mixin. I felt like this is a more reasonable fix than putting a @ts-expect-error, as the typescript maintainers said it's intended behaviour.

bennypowers (approved):

PR #2086, opened 2021-01-28
Update rollup to fail on error

pmcelhaney:
I got ran into mysterious error in CI with no context and exit code 0. 

> html-minifier-terser deleted something major, cannot proceed.

It's because Rollup is configured to swallow exceptions. After making this change directly in my node_modules, I was able to get a stack trace and diagnose the problem immediately.


LarsDenBakker:
The logOnError option is set, so the error is logged. Throwing on error for the minifier I don't think is a good idea, because the minifier can crash on otherwise valid lit-element CSS templates.

PR #2092, opened 2021-02-04
chore: add exports entry in package.json

manolakis:
## What I did

1. I added an exports entry in package.json


Westbrook:
If we catch this up to master, it makes sense to me!

thepassle:
Adding package exports is usually if not always a breaking change

PR #2111, opened 2021-03-12
chore(deps): remove unneeded babel plugins

abdonrd:
Now Rollup already support it.

- https://github.com/rollup/rollup/pull/3582
- https://github.com/rollup/rollup/pull/3548

### DIFF
--- a/.changeset/good-starfishes-brake.md
+++ b/.changeset/good-starfishes-brake.md
@@ -1,0 +1,5 @@
+---
+'@open-wc/building-rollup': minor
+---
+
+Remove unneeded babel plugins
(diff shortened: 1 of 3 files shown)


Westbrook (approved):
Are we _sure_ these are unneeded? In particular I know some down steam tools (bundlephobia, etc.) are not so keen on these syntaxes...

I'm happy to have this removed, if that's the right call, however.

abdonrd:
Any news here?

LarsDenBakker:
@Westbrook yes this can be removed, there is another babel pass that compiles it away as well.

PR #2120, opened 2021-03-23
Update Storybook addon metadata

coderkevin:
Hey open-wc! We've created a catalog of Storybook addons and yours is included: https://storybook.js.org/addons/storybook-addon-markdown-docs

To make your addon more discoverable in the catalog, we've curated some of its metadata, and I've included those edits in this PR.

To learn more about the catalog metadata, check out our documentation: https://storybook.js.org/docs/react/addons/addon-catalog

### DIFF
--- a/packages/storybook-addon-markdown-docs/package.json
+++ b/packages/storybook-addon-markdown-docs/package.json
@@ -26,6 +26,10 @@
     "src",
     "types"
   ],
+  "keywords": [
+    "storybook-addon",
+    "organize"
+  ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


stephenwade:
Please move `keywords` above `dependencies`. That will satisfy Prettier and should make this pass CI.

Westbrook:
It's also likely that `yarn format:prettier` would put the order down correctly as well.

coderkevin:
I just got back from a trip. I'll update soon. Thanks!

coderkevin:
Thanks!

PR #2119, opened 2021-03-22
Fixed module preload in building-utils.

soelen:
## What I did

image

1. main script was request 2 times when built
2. Investigated a bit, asked team open-wc on slack
3. Investigated more, found this neat [Article](https://developers.google.com/web/updates/2017/12/modulepreload)
4. Searched code snippet where link element is actually being created
5. Created PR

PR #2121, opened 2021-03-24
Attempt to update tests

Westbrook:
## What I did

1. Did #2119 again, to see if the test like me better.


Westbrook:
Correctly implemented as #2119

PR #2142, opened 2021-04-24
chore: keep /docs/ as single source; publish docs via `publish-docs`

daKmoR:
## What I did

1. chore: remove deprecated packages
2. chore: keep /docs/ as single source; publish docs via `publish-docs`


packages/dev-server-hmr/README.md
```diff
@@ -1,5 +1,3 @@
-# Demoing via storybook
+# Hot Module Replacement
 
-Plugin for "hot module replacement" or "fast refresh" with web components.
-
-See [our website](https://open-wc.org/docs/development/hot-module-replacement/) for full documentation or view the docs [on github](../../docs/docs/development/hot-module-replacement.md)
```
LarsDenBakker, line 5: I think the previous description is a bit friendlier. Users might end up on these docs via NPM or git, I like to at least explain what the package is for.

daKmoR: you will see the full content of the docs on npm as it gets copied and processed (with absolute URLs for links, images, ..:) in a prepublishOnly script

PR #2143, opened 2021-04-25
Clean up CONTRIBUTING and match modernweb-dev

stephenwade:
## What I did

I cleaned up and expanded the CONTRIBUTING instructions. I also filed modernweb-dev/web#1416 at Modern Web to make the CONTRIBUTING files match.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,10 +12,11 @@ Next, clone your fork onto your computer (replacing YOUR_USERNAME with your actu
 git clone git@github.com:YOUR_USERNAME/open-wc.git
 ```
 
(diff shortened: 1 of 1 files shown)


daKmoR:
do you think it makes sense to add something like (this is an outline and needs to be "properly" written)

## Update fork to the latest version

The repo updates often...

1. Add Open Web Component as an upstream
```bash
git add remote upstream REPOURL
```

2. Update your fork
```bash
git checkout master
git rebase upstream/master
```

Now new branches creates from your forks master will be up to date

stephenwade:
@daKmoR Yeah, I like that. I'll write up something later today.

daKmoR:
looking good - thx 🤗

PR #2144, opened 2021-04-25
chore: Update husky to v6

stephenwade:
## What I did

Updated husky to v6.


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn lint-staged
```
daKmoR, line 4: huh is that a new way of handling husky? 🤔

I'm curious

stephenwade: yep, husky changed the configuration completely in v6. here's their blog post about it:
https://blog.typicode.com/husky-git-hooks-javascript-config/

daKmoR:
thank you 🤗

PR #2188, opened 2021-05-04
Update mocha and move types to dev deps

edoardocavazza:
## What I did

1. Update mocha to version 8

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -44,6 +44,6 @@
     "@types/mocha": "^5.2.7",
     "@types/sinon-chai": "^3.2.3",
     "chai-a11y-axe": "^1.3.1",
-    "mocha": "^6.2.2"
+    "mocha": "^8.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/testing/package.json
```diff
@@ -39,11 +39,13 @@
...
+    "mocha": "^8.0.0"
+  },
+  "devDependencies": {
     "@types/chai": "^4.2.11",
     "@types/chai-dom": "^0.0.9",
     "@types/mocha": "^5.2.7",
-    "@types/sinon-chai": "^3.2.3",
-    "chai-a11y-axe": "^1.3.1",
-    "mocha": "^6.2.2"
+    "@types/sinon-chai": "^3.2.3"
```
LarsDenBakker, line 49: why do you think it should be moved to dev deps? it's used by this package, so users get the correct types

edoardocavazza: Fair... I am going to revert this change

PR #2190, opened 2021-05-05
chore: finish and publish update to lit2 blog post

daKmoR:
## What I did

1. chore: finish and publish update to lit2 blog post

### DIFF
--- a/docs/blog/updating-to-lit-2-0/index.md
+++ b/docs/blog/updating-to-lit-2-0/index.md
@@ -128,13 +128,13 @@ npm i -D @open-wc/dev-server-hmr@next
 
 ## Service Worker strategy change
 
-As our page grows so did what you needed to download when visiting our page. Our strategy was to make the full page available offline whenever you visited our website. That was fine when we had a view pages but now an update can lead to you downloading 10 MB even though you might only be interested in a single article. It also meant that you where looking at the old cached version of the website until every single page was downloaded. So from today we are changing our strategy to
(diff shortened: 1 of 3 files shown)


docs/blog/updating-to-lit-2-0/index.md
```diff
@@ -114,6 +126,21 @@ npm i -D @open-wc/dev-server-hmr@0.1.2-next.0
   };
   ```
 
+## Service Worker strategy change
+
+As our page grows so did what you needed to download when visiting our page. Our strategy was to make the full page available offline whenever you visited our website. That was fine when we had a view pages but now an update can lead to you downloading 10 MB even though you might only be interested in a single article. It also meant that you where looking at the old cached version of the website until every single page was downloaded. So from today we are changing our strategy to
```
thepassle, line 131:
```suggestion
As our page grows so did what you needed to download when visiting our page. Our strategy was to make the full documentation site available for offline usage whenever you visited our website. That was fine when we had only a few pages but now an update can lead to our users downloading 10 MB even though you might only be interested in a single article. It also meant that the user would see the old cached version of the website until every single page was downloaded. So from today we are changing our strategy to
```

docs/blog/updating-to-lit-2-0/index.md
```diff
@@ -114,6 +126,21 @@ npm i -D @open-wc/dev-server-hmr@0.1.2-next.0
...
+
+- caches only visited pages
+- caches assets of visited pages (up to 100 files then it replaces older entries)
+- on service worker activation it will reload the page if a newer version is available
+
+At some point we want to offer you choices to enable offline for the entire website or for sections of the website but it's not there yet.
```
thepassle, line 137:
```suggestion
In the future, we'll iterate on this pattern, and allow users to opt in to enabling offline mode for the entire website, or select parts of the website.
```

Issue #2206, opened 2021-05-11
[eslint-plugin-lit-a11y] Remove dependency on intl-list-format


TimvdLippe:
## Expected behavior

Installing `eslint-plugin-lit-a11y` should use a small disk footprint

## Actual Behavior

`eslint-plugin-lit-a11y` depends on `intl-list-format` which is 9.1MB on disk, currently the largest dependency in all `node_modules` of Chrome DevTools. It appears that it was only used for Node 10 compatibility (https://github.com/open-wc/open-wc/blob/4f55bf52fde479d9b2e2ed61cea78b29bd97082a/packages/eslint-plugin-lit-a11y/lib/rules/role-has-required-aria-attrs.js#L15)

Since Node 10 is EOL, can this large dependency be removed?

## Additional context

Happens on the latest version (1.0.1).

TimvdLippe:
@thepassle can we reopen this issue please?

thepassle:
For sure. Sorry this is taking so long, we're in prerelease mode for open-wc because we're waiting for lit to release the new version, but its taken a bit longer than anticipated

TimvdLippe:
Thanks for fixing this! It removed about 60k lines from our `node_modules` :tada: https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/3268293/

PR #2208, opened 2021-05-11
chore: remove intl list format

thepassle:
## What I did

1. removes intl-list-format for old node versions https://github.com/open-wc/open-wc/issues/2206


Westbrook:
@thepassle did we want to move forward with this?

bennypowers:
we just need to regenerate the lock file then we can merge this, right?

thepassle:
@bennypowers can you merge this when CI passes? Then we can merge your PR after

PR #2211, opened 2021-05-15
Fix broken link

prafullkotecha:
## What I did

1. Removed Morningstar entry.


daKmoR:
it seems morning star is no longer using web components but is based on vue now...

probably better to remove it 😅

prafullkotecha:
@daKmoR - How about now? Removed the entry.

daKmoR:
thank you 🤗

PR #2214, opened 2021-05-20
refactor(scoped-elements): patch polyfill instead of forking it

manolakis:
This commit removes the forked scoped registries polyfill and patches the webcomponents one.

## What I did

1. removes the polyfill code
2. patches the original polyfill

Also, this change patches the native HTMLElement only for those elements created before applying the polyfill, so it should have better performance.


daKmoR:
using patch-package only works for yourself... which is fine for an app but we can't release a library needing it... 

(e.g. every user would need to manually copy the patch file and run patch-package themselves to make it work)

manolakis:
are you sure? I though it work also as a dependency, although I didn't test it postinstall script should be executed when a dependency is installed 🤔

manolakis:
Them I'm closing the PR

packages/scoped-elements/package.json
```diff
@@ -45,6 +46,10 @@
     "@open-wc/dedupe-mixin": "^1.3.0",
     "@webcomponents/scoped-custom-element-registry": "0.0.1"
   },
+  "devDependencies": {
+    "patch-package": "^6.4.7",
```
LarsDenBakker, line 50: shouldn't this be a prod dependency?

LarsDenBakker:
I'd prefer to not rely on modifying files from other packages in a postinstall step. Could we ship our own variant of the polyfill? We could reuse the original source and apply the patch manually, using patch-package or if the modification is simple with string manipulation.

justinfagnani:
What modifications are being made here?

packages/scoped-elements/package.json
manolakis: I don't think so.... anyway, I'm going to close the PR as is not going to be needed anymore because the change has been approved on the polyfill and merged today.

manolakis:
@justinfagnani we found a corner case when we tried to define a custom element whose class was created __before__ applying the polyfill.

https://github.com/webcomponents/polyfills/issues/440

In a first approach there was a tax applied over all custom element classes, so it was reasonable to not include it in the polyfill. However, we found a way to aaply that tax only on the affected custom element classes so it has been merged into the polyfill today https://github.com/webcomponents/polyfills/pull/441

Once a new version of the polyfill is released we will use it directly.

PR #2219, opened 2021-05-25
fix(eslint-plugin-lit-a11y): deprecate no-invalid-change-handler

### LINKED ISSUE #2218: [eslint-plugin-lit-a11y] no-invalid-change-handler should be deprecated
## Expected behavior

The no-onchange rule has been [deprecated](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md) in eslint-plugin-jsx-a11y as of Nov 2020. The corresponding rule (no-invalid-change-handler) should be deprecated and turned off in the recommended lit-a11y config.

The following code should not report an error with the recommended config:

```js
html`<select @change=${foo}></select>`
```

## Actual Behavior
It is turned on in the recommended lit-a11y config and is not deprecated.

The above code reports the following error: 

> blur must be used instead of change on `<select>`, unless absolutely necessary and it causes no negative consequences for keyboard only or screen reader users. eslintlit-a11y/no-invalid-change-handler

## Additional context

eslint-plugin-lit-a11y v1.0.1

The original reasoning behind this rule seemed to be that IE fired the change event when navigating between options with the arrow keys. From the [linked article](https://www.themaninblue.com/writing/perspective/2004/10/19/) in the jsx-a11y docs:

> The only problem is, when you use the keyboard to explore the options in a select menu (using cursor keys) it triggers the onchange event handler (in IE). So, if you have any actions that rely on that handler, they'll probably be executed before the user has time to listen to all the options in the select menu. It'll also mean they'll probably only ever be able to access the first item. 

This was from 2004 and seems to be out of date now. I wrote a [quick CodePen](https://codepen.io/geoff-rich/full/VwKxxaw) to see when the change event is fired for a `<select>` and tested in IE11, Chrome, and Firefox. In all browsers tested, tabbing to the select, opening the menu with Space, and using the arrow keys to navigate between different items did **not** fire a change event. A change event was only fired when an item was selected with Enter or if the arrow keys were used to select an option without opening the menu.

Since Lit targets modern browsers, legacy (<IE11) select behavior should not be an a11y concern.

geoffrich:
## What I did

Fixes #2218

1. Deprecate no-invalid-change-handler lint rule
2. Turn rule off in recommended config
3. Update docs (adapted from eslint-plugin-jsx-a11y [docs](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md))

### DIFF
--- a/.changeset/nasty-humans-promise.md
+++ b/.changeset/nasty-humans-promise.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-lit-a11y': patch
+---
+
+Deprecate no-invalid-change-handler rule
--- a/docs/blog/testing-workflow-for-web-components/index.md
+++ b/docs/blog/testing-workflow-for-web-components/index.md
@@ -169,7 +169,7 @@ render() {
 }
 ```
 
-Interesting, the test should be green... but it's not :thinking: Let's take a look.
+Interesting, the test should be green... but it's not 🤔 Let's take a look.
 
 ```
 ✖ has a static shadowDom
@@ -193,7 +193,7 @@ If we use `innerHTML` to compare the DOM, we'd have to rely on simple string equ
 
 Fortunately, we've got you covered. If you're using `@open-wc/testing` then it automatically loads the `@open-wc/semantic-dom-diff` chai plugin for us to use.
 
-So let's try it out :muscle:
+So let's try it out 💪
 
 ```js
 // old:
@@ -206,7 +206,7 @@ expect(el).shadowDom.to.equal(`
 `);
 ```
 
-Bam :tada:
+Bam 🎉
 
 ```
 a11y input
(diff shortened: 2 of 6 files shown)


bennypowers (approved):

bennypowers:
I can't remember, do we also need to update the docs site?

geoffrich:
Does updating the rules markdown file automatically update the docs page at https://open-wc.org/docs/linting/eslint-plugin-lit-a11y/rules/no-invalid-change-handler/?

bennypowers:
looks like it needs to be updated

https://deploy-preview-2219--open-wc-org.netlify.app/docs/linting/eslint-plugin-lit-a11y/rules/no-invalid-change-handler/

geoffrich:
Ah, didn't realize there was a separate markdown file. I pushed a commit updating that file with the same changes as the other documentation.

bennypowers:
There are a few linting issues holding this pr back. please run the format commands, and we'll get this in

geoffrich:
Okay, I've run the formatting now. Prettier was reporting issues in the [Testing Workflow for Web Components](https://open-wc.org/blog/testing-workflow-for-web-components/) blog post. The auto-format mangled some of the :emoji: syntax, so I went ahead and replaced them with actual emojis (which matches how the post looks on [DEV](https://dev.to/open-wc/testing-workflow-for-web-components-g73)).

bennypowers:
Thanks a mint!

PR #2224, opened 2021-06-07
feat(scoped-elements): load the scoped registries polyfill

manolakis:
## What I did

1. I removed the forked polyfill to use the webcomponents one.


packages/scoped-elements/package.json
```diff
@@ -43,7 +43,7 @@
   "dependencies": {
     "@lit/reactive-element": "^1.0.0-rc.1",
     "@open-wc/dedupe-mixin": "^1.3.0",
-    "@webcomponents/scoped-custom-element-registry": "0.0.1"
+    "@webcomponents/scoped-custom-element-registry": "0.0.2"
   },
   "types": "types/index.d.ts",
   "sideEffects": [
```
lazka, line 49: The sideEffects array can be removed now I think

manolakis: That's right! thanks!

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -1,4 +1,4 @@
-import './loadPolyfill.js';
+import '@webcomponents/scoped-custom-element-registry';
```
daKmoR, line 1: does this now include the logic to not double load if it's executed 2 times from e.g. nested node_modules? 🤔

manolakis: Yes, it includes all that logic plus all the fixes we send last month, even the logic to "upgrade" CE created before applying the polyfill

daKmoR: sweet 💪

daKmoR (approved):

PR #2233, opened 2021-06-23
fix(lit-helpers): Removed references to removed type definitions

ptc-jhirzel:
## What I did

1. Removed references to already removed d.ts files. This caused compilation issues when using it in a typescript enabled project.

### DIFF
--- a/.changeset/tame-queens-divide.md
+++ b/.changeset/tame-queens-divide.md
@@ -1,0 +1,5 @@
+---
+"@open-wc/lit-helpers": patch
+---
+
+fix(lit-helpers): Removed type references to deprecated `live`, `spread`, and `spreadAttributes` directives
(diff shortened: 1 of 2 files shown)


ptc-jhirzel:
The types and files referenced in the index.d.ts were removed here:
https://github.com/open-wc/open-wc/commit/f3cbb2a70d7df68b0ac0142c0b33e08fa48b866d#diff-6f680b8f5385a1e6d869c32be85c5e641ee4a6417e4121a0a3a8255d9aeee567

bennypowers:
@LarsDenBakker we planning on implementing those any time in the next few days?

ptc-jhirzel:
Is there anything I can do to aid in the process?

bennypowers (approved):

ptc-jhirzel:
Sorry I was on vacation and haven't had a look at this. Seems like the required node tests are still running. Is that something that must be invoked manually?

ClaudioHoffmann:
This exact issue causes TypeScript builds to fail with the released 0.4.0 and 0.4.0-next.1 versions…

ClaudioHoffmann:
Actually, type checking in strict mode has been broken as of 703f57d655f6c8016f891731ec2c22abff73d455, since there never was a `packages/lit-helpers/src/live.d.ts`.

PR #2246, opened 2021-07-15
docs: link to cem/a

bennypowers:
## What I did

1. Add a Custom Elements Manifest page under "Tools"
2. install `rocket-preset-code-tabs`


bennypowers:
one of the deps to `rocket-preset-code-tabs` needs node 14 or higher.

acc'd to https://nodejs.org/en/about/releases/ v12 has another year of maintenance to go. How should we proceed?

bennypowers:
I fixed the incompatible node 12 tests in this pr by removing node 12 from the test matrix. 14 is lts, it's widely supported, and we've already dropped 12 in `create`, so this should be ok. cc @LarsDenBakker @daKmoR

PR #2248, opened 2021-07-15
chore: use latest mdjs dependencies

Westbrook:
## What I did

1. Bumped some dependencies to catch up to MDJS and it's dependency listings...


jorenbroekema (approved):

manolakis (approved):
Looks good to me

PR #2273, opened 2021-08-24
fix(semantic-dom-diff): snapshot testing with WTR

manolakis:
## What I did

1. Snapshot testing with WTR updated to not modify the snapshot files running the tests, which was causing race conditions when tests were run by several browsers.


Westbrook (approved):
Lgtm!!

PR #2283, opened 2021-09-03
fix: use latest axe-core and allow not required attributes in role testing

Westbrook:
## What I did

1. updated `axe-core` to get the freshest and hottest a11y rules
2. corrected `RoleHasRequiredAriaAttrsRule` to test for all required attributes but not to fail when there are more attributes beyond that list on the element.


thepassle (approved):

bennypowers (approved):
nice. lgtm!

PR #2290, opened 2021-09-13
docs: what to know when developing with LitElement

Westbrook:
## What I did

1. I wrote a blog and stuff!


docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+set closeDelay(value: number) {
+  const closeDelay = value;
+  if (closeDelay === this.closeDelay) return;
+  this.requestUpdate('closeDelay', this.closeDelay);
+  this._closeDelay = closeDelay;
+}
+_closeDelay = 300;
+```
+
+This does eclipse the simplicity of a functional solution as we are choosing to apply the default by writing our own getter/setter pair on the property. At the same time, however, we've fallen into a possible trap of the functional fallback approach (and as the `willUpdate()` approach above), we're allowing our entire render lifecycle to be triggered for what might not be an actual change to our component state. Where `this.closeDelay` to already equal `300` and the application to change it to `undefined`, all of these approaches we've looked at so far would cause whatever could occur in your render lifecycle to occur without need.
```
davidmaxwaterman, line 72: "Where `this.closeDelay` to already equal" -> "Were ..."
"lifecycle to occur without need" -> "lifecycle to occur needlessly"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+get closeDelay(): number {
+  return this._closeDelay || 300;
+}
+set closeDelay(value: number) {
+  const closeDelay = value;
+  if (closeDelay === this.closeDelay) return;
+  this.requestUpdate('closeDelay', this.closeDelay);
+  this._closeDelay = closeDelay;
+}
+_closeDelay = 300;
```
davidmaxwaterman, line 69: Would it be better to mark this private (I never used it, but I see if can be done in JS with `#closeDelay`), or is implicit private ('_') good enough?

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+## So you like Typescript?
+
+Typescript loves a good `"Property is not definitely assigned in the constructor"` warning, and if you like Typescript enough to use it a lot, you'll likely run into it at some point. It's telling you this, because in contrast to what you've been told Typescript is NOT smarter than you and it can't tell if something _should_ always be available, only if it _might_ not be available, sometime. You can set an initial value to a property, and it'll never yell:
```
davidmaxwaterman, line 96: "sometime" -> "sometimes" (or just remove ", sometime" completely).

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+```js
+class MyThing extends LitElement {
+  myProperty = 'string'; // always available, always a string
+}
+```
+
+If you're looking for that to _have_ to be initialized to use your element, so _you_ KNOW it's going to always have a value, but you want the consumer to initialize it, then you can tell Typescript that being the `!` operator once and be done with it:
```
davidmaxwaterman, line 104: "then you can tell Typescript that being the `!` operator" -> "then you can tell Typescript that by using the `!` operator"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+    // for one time availability confirmation
+    super.firstUpdated(changedProperties);
+    if (!this.myProperty) console.warn('`myProperty` is unset');
+  }
+}
+```
+
+### Required properties
+
+In this way, you can also manage required properties/attributes. No, it doesn't fall within the available management of a tool-based contract with the consumers of your component, however, a tool-based contract is not strictly enforceable. You as a component author can tell Typescript or a linter to error on certain things, but you consumer can tell them not to just as easily. Deciding to _only_ leverage tooling for this sort of capability might mean less work for you, but it doesn't guarantee better outcomes for your consumers. Any component author will need to decide the risks they are willing to foist onto their consumers when publishing a component, and this is yet another item to manage on that list.
```
davidmaxwaterman, line 132: "but you consumer can tell them not to" -> "...the consumer..." (or "your consumer"?)

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Event management
+
+Event listeners add on `this` directly in a custom element DO NOT need to be cleaned up when disconnected from the `document`. Once all references to the element are released, the same garbage collection that cleans up the element itself will clean up the events bound to it. What's more, when binding events to `this`, you don't need to bind the method, so you can call a class method directly without any `.bind(this)` or `() => {}` whatsoever.
```
davidmaxwaterman, line 136: "Event listeners add on" -> "...added..."
"the events bound to it" -> "the event listeners bound to it"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+class MyThing extends LitElement {
+  myProperty = 'string'; // always available, always a string
+}
+```
+
+If you're looking for that to _have_ to be initialized to use your element, so _you_ KNOW it's going to always have a value, but you want the consumer to initialize it, then you can tell Typescript that being the `!` operator once and be done with it:
+
+```js
+class MyThing extends LitElement {
+  myProperty!: String; // no initial value, but a string is required from the consumer
```
davidmaxwaterman, line 108: "String" -> "string" ?

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+```js
+@customElement('menu-trigger')
+export class MenuTrigger extends LitElement {
+  @property({type: String})
+  trigger?: string;
+
+  public willUpdate(changedProperties: PropertyValues<this>): void {
+    if (changedProperties.has('trigger') {
+      this.removeEventListener(changedPeroperties.get('trigger'), this.eventHandler);
```
davidmaxwaterman, line 146: "changedPeroperties" -> "changedProperties"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  private eventHandler(): void {
+    // do stuff.
+  }
+}
+```
+
+Events are even easier if you know the name of the event you're wanting to listen for will be the same throughout the lifecycle of your application. With that knowledge you can listen just once without needing to add/remove the listener based on even name changes overtime:
```
davidmaxwaterman, line 157: "overtime" -> "over time"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+You aren't able to use numbers directly, no. However, the foot gun of lacking clarity in your code that unearths makes preventing this input a pretty good idea. In a world where `px` is just one of many units that a CSS property could accept (`%`, `vh`, `vwmax`, `pt`, `em`, `rem`, `pt`, `pc`, ad (almost) infinitum) being explicit with your code with not only help your consumers, but it'll help your future self maintain the components that you create. On top of that, there are numbers that you'll eventually want to apply to your styles that won't beg for the application of a unit at all:
```
davidmaxwaterman, line 198: "clarity in your code that unearths makes" -> "clarity in your code that doing so unearths makes"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+<figcaption>
+Thanks @bennypowers for the code sample here.
+</figcaption>
+</figure>
+
+---
+
+One of the best things about `lit-html` the renderer underlying LitElement is that, if you want to live on that wild side, you could easily create your own directive that applied number typed properties as `px` and leverage it in your own work. Here are [the docs](https://lit.dev/docs/templates/custom-directives/) for doing just that! If you're still not convinced, check out some [directives with which I've experimented](https://dev.to/open-wc/doing-a-flip-with-lit-html-2-0-3gn4).
```
davidmaxwaterman, line 223: "things about `lit-html` the renderer underlying LitElement" -> "things about `lit-html`, the renderer underlying LitElement, is that"

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+NOTE: the above article, "Doing a FLIP with lit-html@2.0", was written against an RC of `lit@2.0` and may not be 100% current. I'll be looking to update it here, soon.
+
+## Conclusion
+
+When you're learning a new piece of software, by all means, start by comparing it to something that you know. It's like a cheat code to getting started down the path of learning something new. Once you've done that, don't stop there, get into a real use case with it and learn what sort of capabilities or techniques it unlocks or supports. Only then can you really get into the question of why it's doing so and whether in the context that it should be used (or the context that you might use it) it's the sort of tool you want to leverage for the job.
```
davidmaxwaterman, line 229: "whether in the context that it should be used " -> "whether in the context that it is intended to be used "

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+---
+
+If you do get to building something with LitElement, come share it here and let's chat about the what's and why's of what you've done/are trying to do. I look forward to seeing it here in the comments, or hit me up on the [Lit & Friends Slack](https://join.slack.com/t/lit-and-friends/shared_invite/zt-llwznvsy-LZwT13R66gOgnrg12PUGqw)!
```
davidmaxwaterman, line 233: "what's and why's" -> "whats and whys"

davidmaxwaterman (commented):
Mostly just suggested improvements to grammar, but also some questions about code. Feel free to ignore/disagree :)

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+---
+
+## Default Values
+
+If you'd like to define default values for a property on a LitElement based custom element, there are a number of ways that you could do this. While some do require management across the entire element class, we'll ignore those today as there are a number of options that do not require such work.
+
+First off, there's a very direct comparison between LitElement and a functional approach to UI development where you could consider the `render()` method of a LitElement as an almost 1 to 1 conversion from the functional definitions found in other offerings. In this way, you could assume `render()` is the only entry into your properties and leverage your fallbacks like so:
```
bennypowers, line 36:
```suggestion
First off, LitElement's `render()` method of a LitElement is (from the user's perspective) an almost 1 to 1 conversion from the functional definitions found in other offerings. In this way, you could treat `render()` as the only entry into your properties and define fallbacks at the top of your render function's body like so:
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+```js
+render() {
+  const closeDelay = this.closeDelay ?? 300;
+  ...
```
bennypowers, line 41:
```suggestion
  return html`<x-dialog delay=${closeDelay}>Thank you!</x-dialog>`;
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
bennypowers: could add a section here on how to write a 'defaulting property' decorator with typescript

docs/blog/things-to-know-when-developing-with-lit-element/index.md
bennypowers, line 136:
```suggestion
Event listeners added directly on `this` in a custom element _do not_ need to be cleaned up when disconnected from the `document`. Once all references to the element are released, the same garbage collection that cleans up the element itself will clean up the events bound to it. What's more, when calling `addEventListener` on `this`, the method's `this` reference automatically reverts to the instance. You don't need to bind the method, so you can call a class method directly without any `.bind(this)` or class field arrow-functions.
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
bennypowers, line 146:
```suggestion
      this.removeEventListener(changedProperties.get('trigger'), this.eventHandler);
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
bennypowers, line 157:
```suggestion
Events are even easier if you know the name of the event you're wanting to listen for will be the same throughout the lifecycle of your application. With that knowledge you can listen just once without needing to add/remove the listener based on even name changes over time:
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+    // do stuff.
+  }
+}
+```
+
+Done and done!
+
+## styleMap usage
+
+Contrary to what you might read in the documentation site (PRs are incoming to correct, so you may miss it altogether!), `styleMap()` does accept `[name: string]: string | undefined | null;` which allows you to leverage `undefined` to prevent a property from being added to the style of an element, e.g.:
```
bennypowers, line 177: by the time we publish this line will prolly' be out of date

bennypowers, line 177:
```suggestion
Lit's `styleMap()` directive helps when setting the `style` attribute on HTML elements from JavaScript. It accepts an object with css-property keys and string, `undefined`, or `null` values. This means you can prevent a CSS property from being added to the element by passing `null` or `undefined` as the value, e.g.:
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
bennypowers:
```suggestion
The current version of Lit's styleMap excludes numbers as values. You might expect Lit to automatically convert numbers to `px` values, but on second thought that isn't actually what you'd want. In CSS, where `px` is just one of many units that a numeric CSS property could accept (`%`, `vh`, `vwmax`, `pt`, `em`, `rem`, `pt`, `pc`, _ad infinitum_), there's no way for Lit to know or even assume what kind of number you're using. On top of that, you might want to apply unit-less numbers directly to your styles:
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+```js
+class XL extends LitElement {
+  static styles = css`
+    output {
+      background-color: hsl(var(--hue, 0) 50 100);
+    }
+  `;
+  @property({ type: Number }) hue = 0;
+  render() {
+    return html` <output style=${styleMap({ '--hue': this.hue })}></output> `;
```
bennypowers, line 211:
```suggestion
    return html` <output style=${styleMap({ '--hue': this.hue.toString() })}></output> `;
```

docs/blog/things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,233 @@
...
+  }
+}
+```
+
+<figcaption>
+Thanks @bennypowers for the code sample here.
+</figcaption>
+</figure>
+
```
bennypowers, line 220:
```suggestion

Defaulting numbers to `px` would be a foot-gun. Instead, Lit encourages you to be explicit with your CSS code. Doing so not only helps your consumers, but your teammates and future self as well when it comes time to maintain the components that you create.

```

bennypowers (changes requested):
some changes here and there, mostly phrasing.

I'm wondering if there are some other points we might want to address like

- build systems
- testing
- ssr or quote-unquote isomorphic js
- litelement for b2b dashboard spas
- litelement for design systems
- litelement for reusable widgets / microfrontends

docs/blog/things-to-know-when-developing-with-lit-element/index.md
Westbrook: Honestly not sure than devs in most other contexts know what `#closeDelay` would mean in this case, and think it distracts from the broader point to try to describe it. 

There are lots of other sections that could be included here, and I think the right call is to rename the article "Some things..." to imply there is always more to know!

bennypowers (approved):

docs/blog/some-things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,235 @@
...
+<figcaption>
+Thanks @bennypowers for the code sample here.
+</figcaption>
+</figure>
+
+Defaulting numbers to `px` would be a foot-gun. Instead, Lit encourages you to be explicit with your CSS code. Doing so not only helps your consumers, but your teammates and future self as well when it comes time to maintain the components that you create.
+
+---
+
+One of the best things about `lit-html`, the renderer underlying LitElement, is that, if you want to live on that wild side, you could easily create your own directive that applied number typed properties as `px` and leverage it in your own work. Here are [the docs](https://lit.dev/docs/templates/custom-directives/) for doing just that! If you're still not convinced, check out some [directives with which I've experimented](https://dev.to/open-wc/doing-a-flip-with-lit-html-2-0-3gn4).
```
daKmoR, line 225:
```suggestion
One of the best things about `lit-html`, the renderer underlying LitElement, is that, if you want to live on that wild side, you could create your own directive that applied number typed properties as `px` and leverage it in your own work. Here are [the docs](https://lit.dev/docs/templates/custom-directives/) for doing just that! If you're still not convinced, check out some [directives with which I've experimented](https://dev.to/open-wc/doing-a-flip-with-lit-html-2-0-3gn4).
```

what is easily is in the eye of the beholder...

docs/blog/some-things-to-know-when-developing-with-lit-element/index.md
```diff
@@ -0,0 +1,235 @@
...
+First off, LitElement's `render()` method of a LitElement is (from the user's perspective) an almost 1 to 1 conversion from the functional definitions found in other offerings. In this way, you could treat `render()` as the only entry into your properties and define fallbacks at the top of your render function's body like so:
+
+```js
+render() {
+  const closeDelay = this.closeDelay ?? 300;
+  return html`<x-dialog delay=${closeDelay}>Thank you!</x-dialog>`;
+}
+```
+
+From here, you could get very close to simply copy and paste the rest of a functional UI component into a LitElement for an early test of its capabilities. Leveraging scoped variables like this, rather than class properties, does leave you in much the same place that functional alternatives to UI development do; needing additional tools for memoization of that scope. As using LitElement means you're already in a class context, we can use the capabilities of a class directly rather than synthesizing them with memoization.
```
daKmoR, line 45:
```suggestion
From here, you could get very close to copy and paste the rest of a functional UI component into a LitElement for an early test of its capabilities. Leveraging scoped variables like this, rather than class properties, does leave you in much the same place that functional alternatives to UI development do; needing additional tools for memoization of that scope. As using LitElement means you're already in a class context, we can use the capabilities of a class directly rather than synthesizing them with memoization.
```

PR #2304, opened 2021-09-23
chore: add format action for PR reviews

bennypowers:
## What I did

1. add an action which
    - any time a PR *REVIEW COMMENT* is created or edited
    - IFF the comment contains the string `CI format please`
    - run `yarn format:prettier && git add . && git commit && git push`
2. add a PR template which contains a note to reviewers 

## Questions:
- Do we want a different magic string, or to create an action with bash or JS or something to parse the comment?
- Would it be better to get a list of changed files and add them only? how? does prettier make that easy? (a cursory check of their docs didn't turn up much)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,11 @@ Open Web Components provides a set of defaults, recommendations and tools to hel
 [![BrowserStack Status](https://www.browserstack.com/automate/badge.svg?badge_key=M2UrSFVRang2OWNuZXlWSlhVc3FUVlJtTDkxMnp6eGFDb2pNakl4bGxnbz0tLUE5RjhCU0NUT1ZWa0NuQ3MySFFWWnc9PQ==--86f7fac07cdbd01dd2b26ae84dc6c8ca49e45b50)](https://www.browserstack.com/automate/public-build/M2UrSFVRang2OWNuZXlWSlhVc3FUVlJtTDkxMnp6eGFDb2pNakl4bGxnbz0tLUE5RjhCU0NUT1ZWa0NuQ3MySFFWWnc9PQ==--86f7fac07cdbd01dd2b26ae84dc6c8ca49e45b50)
 [![Renovate enabled](https://img.shields.io/badge/renovate-enabled-brightgreen.svg)](https://renovatebot.com/)
 
+
+
+
+
+
 ## Usage
 
 ```bash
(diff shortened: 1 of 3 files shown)


daKmoR:
uh very nice 💪

let's go with a more "specific" phrase...  how about `@formatbot make it pretty` 
`format` could end up a lot in "normal" conversations

and IMHO don't bother with only getting the changed files... prettier only takes a few seconds... if it starts to take 20+ seconds then we can do some git magic to find all the changes files and pass them individually to prettier

daKmoR:
is that only the "initial" message at the top of the PR? 🤔  or every comment as well? 🤔

bennypowers:
https://github.com/FormatBot there is an actual github user by that name :/ is that an issue? maybe we should use a string like `FORMAT_PR` or something like that which is unlikely to come up in normal conversation?

This should run only on _PR review comments_
so, _not_ the PR description, _not_ regular pr comments, only reviews.

bennypowers:
something else to consider: do we need to set a git name and email?

daKmoR:
@formatbot make it pretty

daKmoR:
how about

- `CI make it pretty`
- `Format Action make it pretty`
- `CI format it`

I wanna have something that is easy to remember and could "almost" come up in a conversion... 

my vote is for `CI format it`

README.md
```diff
@@ -9,6 +9,11 @@ Open Web Components provides a set of defaults, recommendations and tools to hel
 [![BrowserStack Status](https://www.browserstack.com/automate/badge.svg?badge_key=M2UrSFVRang2OWNuZXlWSlhVc3FUVlJtTDkxMnp6eGFDb2pNakl4bGxnbz0tLUE5RjhCU0NUT1ZWa0NuQ3MySFFWWnc9PQ==--86f7fac07cdbd01dd2b26ae84dc6c8ca49e45b50)](https://www.browserstack.com/automate/public-build/M2UrSFVRang2OWNuZXlWSlhVc3FUVlJtTDkxMnp6eGFDb2pNakl4bGxnbz0tLUE5RjhCU0NUT1ZWa0NuQ3MySFFWWnc9PQ==--86f7fac07cdbd01dd2b26ae84dc6c8ca49e45b50)
 [![Renovate enabled](https://img.shields.io/badge/renovate-enabled-brightgreen.svg)](https://renovatebot.com/)
 
+
```
bennypowers, line 12: CI format please

bennypowers: CI format please

bennypowers:
LOL ok 

I tested this by editing README.md using github's web editor, adding extra new lines to the file, committing, then commenting in a PR review `CI format please`.

The action works just fine, but github-actions bot needs to sign the CLA 😄 
cc @daKmoR

Westbrook (approved):

PR #2282, opened 2021-09-01
[WIP] Add valid-lang rule

andrico1234:
TODO: Add description

Currently stuck, i talk about why [here](https://github.com/open-wc/open-wc/issues/1925)

PR #2316, opened 2021-10-03
eslint-plugin-lit-a11y: use template-analyzer directly from lit plugin

43081j:
This is pretty much a PoC to show how we can use TemplateAnalyzer directly rather than maintaining a copy.

Draft until 43081j/eslint-plugin-lit#116 is merged.


43081j:
hm strange, im sure i ran prettier. will re-format it again tonight.

you may also want to wait for 43081j/eslint-plugin-lit#118 to be resolved before merging this.

thepassle:
CI format please

thepassle:
@bennypowers Should it kick off the formatting automatically?

thepassle:
> the lint-staged task is meant to do it, which is why i assumed it had. but apparently not. ill just run it manually

Yeah, benny implemented a fancy pants format action recently https://github.com/open-wc/open-wc/pull/2304/files but im not sure how to use it 😄

bennypowers:
> > the lint-staged task is meant to do it, which is why i assumed it had. but apparently not. ill just run it manually
> 
> Yeah, benny implemented a fancy pants format action recently https://github.com/open-wc/open-wc/pull/2304/files but im not sure how to use it 😄

you have to make that comment _specifically in a pr review_

bennypowers (commented):
CI format please

43081j:
should be able to, i haven't disabled that.

ill run it later when im back on my laptop anyway 👍

packages/eslint-plugin-lit-a11y/lib/rules/accessible-emoji.js
```diff
@@ -5,8 +5,8 @@
 
 const emojiRegex = require('emoji-regex');
 const ruleExtender = require('eslint-rule-extender');
-const { TemplateAnalyzer } = require('../../template-analyzer/template-analyzer.js');
```
bennypowers, line 8: CI format please

bennypowers: this one triggered it
we should add prettier action problem matcher thingy to get highlights on lines with errors

bennypowers: https://github.com/open-wc/open-wc/actions/runs/1311057217 it says "action required"

not exactly sure which action it requires though :D

bennypowers: https://github.com/open-wc/open-wc/runs/3812670994?check_suite_focus=true

🤷‍♂️ I blame James

43081j:
ran the formatter myself. weirdly, though, it reformatted stuff that i haven't touched and didn't go anywhere near the imports you were expecting it to format...

just did an `npm run format`

andrico1234:
@43081j , not sure if you've seen, but I've added tests in the #2282 MR to ensure that the location is correct.

43081j:
turns out the locations were right all along and i was struggling with this problem for no reason. the locations are already the location in the source, not in the normalised document.

no problem here. so should all be fine. 

@andrico1234 might be worth you making a temporary branch off this branch to see if your tests pass as expected

andrico1234:
After branching off your template-analyzer branch, I implemented the `valid-lang` code. The correct cases pass and fail, but the LOC doesn't resolve correctly any more. 

The LOC for the `noLangPresent` cases is off by 1. expected: 6, actual: 5.

The LOC for the `invalidLang` cases are off by a bit more. expected : 12, actual: 5.

None of the other tests are failing. But it looks like the LOC returned regardless for the `html` element is the start of the expression, as opposed to the beginning of the element/attribute.

43081j:
can you point me at that branch? so i can check it out and do a bit of debugging. as i dont have any failing test cases myself yet, so would be useful. i manually tested a bunch of positions too and all seemed fine, so may be an edge case i haven't bumped into yet

andrico1234:
Here's a link to the [rule's source](https://github.com/andrico1234/temp-open-wc/blob/valid-lang-test/packages/eslint-plugin-lit-a11y/lib/rules/valid-lang.js), here's a link to the [test](https://github.com/andrico1234/temp-open-wc/blob/valid-lang-test/packages/eslint-plugin-lit-a11y/tests/lib/rules/valid-lang.js).

I'll also test again once the next version is published eslint-plugin-lit is published, thanks!

43081j:
image

@andrico1234 problem solved. looks like the locations are correct now.

for who might be interested, a couple of things needed fixing (in this branch, not in the lang branch or lit plugin):

* parse5 locations can be null - if this happens, someone did something really funky or its a tree correction node (e.g. a missing head tag, something added during normalisation)
* the analyzer can return null for locations sometimes (similar to above point, weird situations and tree correction nodes)
* analyzer methods generally need the eslint source code object, i forgot to pass this in the first time around (*this is why your positions were off* because it was throwing an error, catching it and resolving to the quasi's location instead)

PR #2326, opened 2021-10-07
chore: update rocket to 0.10.0

daKmoR:
## What I did

1. Long overdue update to latest rocket to get an MPA service worker (caching per page visit)
2. Needed to bring over some css


bennypowers:
i also ran into that issue with the css

in the long run we need to cache those in css files and inline style tags

for now moving to an img tag makes sense

PR #2329, opened 2021-10-15
fix(eslint-plugin-lit-a11y): remove eslint hard dep

### LINKED ISSUE #2328: [eslint-plugin-lit-a11y] Add ESLint 8 as peer dependency
## Expected behavior

Installing ESLint 8 with `eslint-plugin-lit-a11y` should work

## Actual Behavior

`eslint-plugin-lit-a11y` gets a nested dependency on ESLint, rather than a peer dependency

## Additional context

https://github.com/open-wc/open-wc/blob/2c485ec00c2409d98c89efe377c879abc7984c0d/packages/eslint-plugin-lit-a11y/package.json#L29 currently only lists version 5, but it probably should be updated to `^5 || ^6 || ^7 || ^8`

bennypowers:
fixes #2328

## What I did

1. remove eslint dependency from `eslint-plugin-lit-a11y` - it already has a peerDep to eslint >= 5

## Questions
Is this a semver major?

### DIFF
--- a/.changeset/cuddly-cows-wait.md
+++ b/.changeset/cuddly-cows-wait.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-lit-a11y': major
+---
+
+chore: remove intl list format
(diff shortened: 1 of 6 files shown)


bennypowers:
Yes I'm fine with that

PR #2345, opened 2021-11-05
fix(testing-helpers): ts types

bennypowers:
## What I did

1. put litrenderable in a ts source
this fixes generated `.d.ts` files, which would break when they try to import `lit-html` instead of `lit`


packages/testing-helpers/src/renderable.ts
```diff
@@ -0,0 +1,12 @@
+import type { TemplateResult } from 'lit';
+
+export type LitHTMLRenderable = (
+    TemplateResult
```
bennypowers, line 4: ci format please

daKmoR (approved):

PR #2356, opened 2021-11-08
Updated Material web component link

akmittal:
## What I did

1. Updated Material Components link earlier link was not working


Westbrook:
Thanks for the submission here, but I'm not sure this is the best link for our goals here. We've recently updated this to https://material-components.github.io/material-web/demos/index.html and I think that'll be the better destination to send people to for now.

PR #2363, opened 2021-11-26
feat/form-participation

calebdwilliams:
## What I did

1. Added the foundations of form-control and tests based on feedback from [this discussion topic](https://github.com/open-wc/open-wc/discussions/2344).

This PR implements the patterns in [this proof of concept](https://github.com/calebdwilliams/form-control-poc).


packages/form-control/package.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "@open-wc/form-control",
```
Westbrook, line 2: Important pre-ship question, do you feel there will be other pieces of "form participation functionality" that you would want to add later in a way that would benefit from naming this `@open-wc/form-participation` with multiple features to this process in the sub-directories of `src`?

packages/form-control/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './FormControlMixin';
```
Westbrook, line 1: We should lean toward including file extensions beyond the use of package entry points. In that way we could _possibly_ update to `typescript@4.5` so we have export map support and list these entries there, as an alternative.

Westbrook: Separately, do we want to prefer the default entry point a "everything in the package so get ready with your tree shake enabled bundler" or is there an "absolute minimum export" that would be useful to use and allow consumers to build up from?

packages/form-control/tsconfig.build.tsbuildinfo
```diff
@@ -0,0 +1 @@
...ac291f4220eedd52eb","b32e45214553c978dbc1b403992ed76b2209b8ad15329490ad78d88639e89368","2581f7c4e458fe9ba8d1101354bbdcbcfc4c7ef07ee45e713f01206ca66fe9b8","590a68d417311f514b15cb3175f8250ee8f9ab2f1c3599ef0d689ea07741e7d5","01d45a0fe4a03207f86567881721b5f5ab06c31283aff8ba15bc19e3fd2d23ac",{"version":"0359800d3b440f8515001431cde1500944e156040577425eb3f7b80af0846612","affectsGlobalScope":true},{"version":"dc45d3b40fe1c20d7823e179b8e41530ae32c221e0e4d759397f110647d8d566","affectsGlobalScope":true}],"options":{"allowSyntheticDefaultImports":true,"composite":true,"declaration":true,"declarationMap":true,"downlevelIteration":true,"emitDeclarationOnly":true,"esModuleInterop":true,"experimentalDecorators":true,"module":99,"outDir":"./types","rootDir":"./src","sourceMap":true,"strict":false,"target":5},"fileIdsList":[[32],[31,32],[31,32,33],[31,33],[34,35],[35,36,37,38],[34],[39]],"referencedMap":[[31,1],[33,2],[34,3],[32,4],[36,5],[39,6],[35,7],[37,8]],"exportedModulesMap":[[31,1],[33,2],[34,3],[32,4],[36,5],[39,6],[35,7],[37,8]],"semanticDiagnosticsPerFile":[41,40,31,33,34,32,6,7,11,10,2,12,13,14,15,16,17,18,19,3,4,23,20,21,22,24,25,26,5,30,27,28,29,1,9,8,36,39,35,38,37]},"version":"4.5.2"}
```
Westbrook, line 1: Probably want to ignore this?

packages/form-control/src/FormControlMixin.ts
```diff
@@ -0,0 +1,528 @@
...
+     * @private
+     */
+    ___requestUpdate() {
+      this.___checkForError();
+      /** @ts-ignore */
+      if (this.requestUpdate) {
+        /** @ts-ignore */
+        this.requestUpdate();
+      }
+    }
```
Westbrook, line 423: What could we do now to prepare the architecture here for integrating with other non-lit libraries?

packages/form-control/src/FormControlMixin.ts
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * A list of Validator objects that will be evaluated when a control's form
+     * value is modified or optionally when a given attribute changes.
+     *
+     * When a Validator's callback returns false, the entire form control will
+     * be set to an invalid state.
+     */
+    static get formControlValidators(): Validator[] {
+      return [];
+    }
```
Westbrook, line 20: I'm not sure the question I had about this has been answered yet. Being a `static` would I need to make two custom inputs if one of them required a min-length and another enforced a max-length? I've done no testing on this, but from the outside `static` tends to mean that.

packages/form-control/src/FormControlMixin.ts
```diff
@@ -0,0 +1,528 @@
...
+     */
+    get validity(): ValidityState {
+      return this.internals.validity;
+    }
+
+    /**
+     * Exists to control when an error should be displayed
+     * @private
+     */
+     ___forceError = false;
```
Westbrook, line 117: This uses `___` to privatize properties and methods, some are less likely to be inadvertently duplicated in other cases, but `___onFocus` isn't so unique across multiple stacked features. Would we benefit any from using `Symbol('onFocus')`, hiding these values in a cache, or similar?

packages/form-control/src/FormControlMixin.ts
```diff
@@ -0,0 +1,528 @@
...
+    constructor(...args: any[]) {
+      super(...args);
+      this.addEventListener('focus', this.___onFocus);
+      this.addEventListener('blur', this.___onBlur);
+      this.addEventListener('invalid', this.___onInvalid);
+
+      let tick = 0;
+      this.___ready = new Promise((resolve, reject) => {
+        const id = setInterval(() => {
+          if (tick >= 100) {
```
Westbrook, line 136: I'm not seeing where `tick` gets incremented in a way that would ever meet this condition...

packages/form-control/src/FormControlMixin.ts
```diff
@@ -0,0 +1,528 @@
...
+      return this.formControlValidators.find(validator =>
+        validator.attribute === attribute
+      );
+    }
+
+    /** The ElementInternals instance for the control. */
+    internals = this.attachInternals() as unknown as IElementInternals;
+
+    /** Keep track of if the control has focus */
+    focused = false;
```
Westbrook, line 54: Some state you track privately, but this you're tracking publicly, what's the benefit here? 

Unrelated to the technical implementation, I ask due to the fact that many elements I use already leverage this property to proxy `:focus-visible`.

Westbrook (commented):
Quick early qeustions.

packages/form-control/package.json
calebdwilliams: Form participation makes a lot of sense, I'll make that change in the next push.

packages/form-control/src/index.ts
calebdwilliams: The mixin _could_ be the the base import. I'll admit to not being super familiar with this space (re: TypeScript 4.5 entry points), happy to defer do whatever makes the most sense (any good examples or tutorials on this I should check out to get caught up?).

It could be that eventually we have more options, but given the current state, the mixin is probably the minimum export.

packages/form-control/tsconfig.build.tsbuildinfo
calebdwilliams: Very much so, yes.

packages/form-control/src/FormControlMixin.ts
calebdwilliams: It _might_ not be necessary any more based on how I have it set up. I'll do some testing.

packages/form-control/src/FormControlMixin.ts
calebdwilliams: No, those tend to key off of attributes and can use properties. For required, for example, the validator is 

```typescript
export const requiredValidator: Validator = {
  attribute: 'required',
  key: 'valueMissing',
  message: 'You must include a value',
  callback(instance: RequiredHost, value: any): boolean {
    let valid = true;

    if (instance.required && !value) {
      valid = false;
    }

    return valid;
  }
};
```

Based off the API here, the evaluator is run on the change of the `'required'` attribute. The callback function can look for whatever property or attribute it needs since it has access to the instance. To be more robust the check for required could be something more like
```typescript
const required = instance.required || instance.hasAttribute('required');
```

Ultimately it depends on how a given user wants to organize their validator callback per validator. If you know you always want your component to be required (like an agree to terms component), you could just as easily write:

```typescript
const alwaysRequiredValidator: Validator = {
  key: 'valueMissing',
  message: 'This field is required',
  callback(instance: HTMLElement, value: string) {
    return !!value;
  }
}
```

so it all really depends on the use case. The above won't depend or evaluate on anything except value change.

packages/form-control/src/FormControlMixin.ts
calebdwilliams: Good call. Would you prefer using `Symbol` or private fields?

packages/form-control/src/FormControlMixin.ts
calebdwilliams: Yeah, looks like that got missed.

packages/form-control/src/FormControlMixin.ts
calebdwilliams: Yeah, this is mostly a convenience thing. It might be the case that a user wants their own show error algorithm. I suppose I should let them take care of that, however.

packages/form-control/src/FormControlMixin.ts
calebdwilliams: OK, this won't be necessary any more. Instead of calling requestUpdate, I'm grabbing the get/set off the prototype or the instance if they exist and calling the after manipulating `value` now, so updates should propagate.

packages/form-participation/package.json
```diff
@@ -0,0 +1,31 @@
...
+      "@web/test-runner-commands": "^0.5.13",
+      "@web/test-runner-playwright": "^0.8.8",
+      "construct-style-sheets-polyfill": "^3.0.4",
+      "element-internals-polyfill": "^0.1.47",
+      "lit": "^2.0.2",
+      "sinon": "^12.0.1",
+      "typescript": "^4.4.4"
+    },
+    "sideEffects": false,
+    "dependencies": {
```
michaelwarren1106, line 27: seems like this should be in `devDependencies`?

packages/form-participation/src/validators.ts
```diff
@@ -0,0 +1,75 @@
+import { Validator } from './index';
+
+export interface RequiredHost extends HTMLElement {
+  required: boolean;
+};
+
+export const requiredValidator: Validator = {
+  attribute: 'required',
+  key: 'valueMissing',
+  message: 'You must include a value',
```
michaelwarren1106, line 10: nit: if this is going to be the default required message, maybe we should use a friendlier tone? or something less personal like not using "you must" but having messaging like in the minLengthValidator like `Value must not be empty`

packages/form-participation/src/validators.ts
```diff
@@ -0,0 +1,75 @@
+import { Validator } from './index';
+
+export interface RequiredHost extends HTMLElement {
+  required: boolean;
+};
+
```
michaelwarren1106, line 6: general question: how many default validators need to be provided? if there are more, should we add them here, or as a fast follow PR? I don't mind PRing some more default validators to add to this lib

packages/form-participation/test-web/MixinWithValidators.lit.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  it('will mark the validity state as invalid given the callback', async () => {
+    expect(el.validity.valid).to.be.false;
+  });
+
+  it('will have a reference to the parent form', async () => {
+    expect(el.form).to.equal(form);
+  });
+
+  it('will not recommend showing an error when pristine', async () => {
+    expect(el.validity.valid).to.be.false;
```
michaelwarren1106, line 93: should there be an expect here making sure that its actually in the pristine state when the other expects are called?

packages/form-participation/test-web/MixinWithValidators.lit.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    form = await fixture(html`<form @submit="${e => e.preventDefault()}">
+      <with-checked name="with-checked"></with-checked>
+    </form>`);
+    el = form.querySelector('with-checked');
+  });
+
+  afterEach(fixtureCleanup);
+
+  it('will not participate in the form if checked is false', async () => {
+    expect(new FormData(form).get('with-checked')).to.equal(null);
```
michaelwarren1106, line 152: question: is this native behavior? at first glance it seems like there ought to be a form entry just set to false?

michaelwarren1106:
general question about documentation. Where should that go? README in the package and also on the open-wc site somewhere? I can help with doing docs if need be

packages/form-participation/package.json
calebdwilliams: Done

packages/form-participation/src/validators.ts
calebdwilliams: That's a fair point, I think if this is something open-wc would like to continue with, these should all be massaged a bit more (and maybe a more definitive list of validators created). Happy to revisit all of these.

packages/form-participation/src/validators.ts
calebdwilliams: As above, happy to have that discussion. I didn't want to spend too much time on it just yet, but leave that open for discussion. Would we need min, max, isNumber, isDate, etc?

packages/form-participation/test-web/MixinWithValidators.lit.test.ts
calebdwilliams: Not other than the `fixtureCleanup` this really is dependent on how the example was coded. Maybe not the best test in the world.

calebdwilliams:
Happy to accept feedback including if y'all think this isn't the right level of abstraction or if it doesn't make sense in this org. @Westbrook and @thepassle have shown at least a little interest previously. 

Again, if this isn't something that makes sense for this org, I'll probably just publish it on my own, but I figured it could use more eyes/minds than just my own and has potential to help a lot of peoples' lives easier.

packages/form-participation/src/validators.ts
Westbrook: I think following up is a good call. Also, ensuring that documentation clearly outlines making your own so that people can test/vend/leverage their own over time is good.

Westbrook:
See https://github.com/open-wc/open-wc/tree/master/packages/building-rollup for general documentation strategies in this project. I could also see a README first approach to to effectively "stealth release" this start, but need to get into the final agreements with the team here before those decisions are made.

packages/form-participation/src/validators.ts
Westbrook: I think the bigger question is whether customization of messages here is simple. Anywhere that text strings are delivered to a user we need to keep in mind the need to translate them, etc.

calebdwilliams: There are a number of strategies built in currently to modify error messages including an instance callback that passes the validation key to the user and expects a returned string. That callback could obviously do anything.

packages/form-participation/test-web/utils.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    form.removeEventListener('submit', prevent);
+  });
+});
+
+
+describe('The formValues utility', () => {
+  let form: HTMLFormElement;
+
+  beforeEach(async () => {
+    form = await fixture(html`<form>
```
Westbrook, line 60:
```suggestion
    form = await fixture<HTMLFormElement>(html`<form>
```
Does this prevent the need to cast through `unknown`?

packages/form-participation/test-web/MixinWithValidators.lit.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  beforeEach(async () => {
+    form = await fixture(html`
+      <form>
+        <form-control-lit-validators
+          name="formControl"
+        ></form-control-lit-validators>
+      </form>
+    `);
+    // @ts-ignore
+    el = form.querySelector<FormControlLitValidators>('form-control-lit-validators');
```
Westbrook, line 61: Does casting once here prevent needing to do it in multiple tests below?
```suggestion
    el = form.querySelector<FormControlLitValidators>('form-control-lit-validators') as FormControlLitValidators;
```

packages/form-participation/test-web/FormControlMixin.lit.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  beforeEach(async () => {
+    form = await fixture(html`
+      <form>
+        <form-control-lit-no-validators
+          name="formControl"
+        ></form-control-lit-no-validators>
+      </form>
+    `);
+    // @ts-ignore
+    el = form.querySelector<FormControlLitNoValidators>('form-control-lit-no-validators');
```
Westbrook, line 42: What does TS complain about here? Do we need to place the test element on the `HTMLElementTagNameMap`?

packages/form-participation/src/FormControlMixin.ts
```diff
@@ -0,0 +1,517 @@
...
+     * Validator is associated with the attribute, it will return null.
+     */
+    static getValidator(attribute: string): Validator {
+      return this.formControlValidators.find(validator =>
+        validator.attribute === attribute
+      );
+    }
+
+    /** The ElementInternals instance for the control. */
+    internals = this.attachInternals() as unknown as IElementInternals;
```
Westbrook, line 62: Does `attachedInternals` have the same "fire once" requirement as `attachShadow`? What happens if someone had already attached internals in an earlier mixin down the line? _Might_ not be anything we can do, but would be worth documented in that case.

packages/form-participation/src/FormControlMixin.ts
```diff
@@ -0,0 +1,517 @@
...
+    get validationMessage(): string {
+      return this.internals.validationMessage;
+    }
+
+    /**
+     * The element's validity state after evaluating the control's Validators.
+     * This property implements the same patterns as the built-in constraint
+     * validation strategy.
+     */
+    get validity(): ValidityState {
```
Westbrook, line 120: Should/can we also cover the `checkValidity` method that we see on native `input` elements here?

packages/form-participation/src/FormControlMixin.ts
```diff
@@ -0,0 +1,517 @@
...
+        this[validate](this.value);
+      }
+    }
+
+    connectedCallback() {
+      if (super.connectedCallback) {
+        super.connectedCallback();
+      }
+
+      /** Initialize the form control  and perform initial validation */
```
Westbrook, line 154:
```suggestion
      /** Initialize the form control and perform initial validation */
```

Westbrook (commented):
What the recommended way to get a demo of this running from the repo. Have a couple of minutia checks I want to get into as the overall structure it starting to come together nicely.

Also, this will be a part of the core team's meeting on Wednesday, after which I hope we can have a firm answer on next steps!

packages/form-control/src/FormControlMixin.ts
calebdwilliams: @Westbrook is there a preference on this one? I've updated the PR to use `Symbol` for the time being. OK to resolve this conversation?

packages/form-participation/src/validators.ts
calebdwilliams: OK, coming up with a list of default validators:

- [ ] Required/valueMissing
- [ ] minlength/tooShort
- [ ] maxlength/tooLong
- [ ] isNumber/typeMismatch
- [ ] isDate/typeMismatch
- [ ] min/rangeUnderflow (number and date?)
- [ ] max/rangeOverflow (number and date?)
- [ ] pattern (regex)/patternMismatch

Any others we can think of?

packages/form-participation/test-web/utils.test.ts
calebdwilliams: Yes. Updated.

packages/form-participation/src/FormControlMixin.ts
calebdwilliams: Yeah, it's a once-only thing. Let me think on this.

packages/form-participation/test-web/FormControlMixin.lit.test.ts
calebdwilliams: Nothing apparently. I'll remove it.

packages/form-participation/src/FormControlMixin.ts
calebdwilliams: Adding.

packages/form-participation/src/FormControlMixin.ts
calebdwilliams: Oof. Good catch. Updated.

Westbrook:
We're excited to take this into the larger Open WC project, however to best support getting the right people to work on it long term, we're looking at getting a stand along repo for this. Scalability wise, any druthers on what it should be named? In theory something that works well with `@open-wc/...` but flexible. Not sure if `form-participation` is too specific or `element-mixins` is too general, etc.

calebdwilliams:
@Westbrook I think form-participation makes sense for these features. I think we'll eventually be interested in offering some more features down the road. While what I have now is a generic form control mixin, some other features might make sense in the long term (off the top of my head):

- A Lit directive that links host and shadow-included form element state
- Lit controllers for selectively adding features like value participation or validity

That is, of course, assuming we are cool with eventually adding other packages to that repo (kind of like having a separate repo just for form participation purposes). Another valid set up would just be having one package for form participation in general with several different utilities, but I can definitely see a world where there's a Lit-optimized version of the current mixin being published …

Westbrook:
Created https://github.com/open-wc/form-participation and added @calebdwilliams and @michaelwarren1106 both with write permissions. Looking forward to seeing how this comes together!

Westbrook:
Closing as we've move this work to a new repo!

PR #2375, opened 2022-01-03
Publish to Netlify from CI

Westbrook:
## What I did

1. Update Netlify CLI
2. Build and publish docs from Github Actions 🤞🏼 
3. Save money!

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,72 @@
+name: Preview
+
+on: pull_request
+
+jobs:
+  release:
+    # Prevents changesets action from creating a PR on forks
+    if: github.repository == 'open-wc/open-wc'
+    name: Preview
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@master
+        with:
+          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
+          fetch-depth: 0
+
+      - name: Setup Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+          cache: 'yarn'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install Dependencies
+        run: yarn --frozen-lockfile
+
+      - name: Build the preview URL from the branch name
+        run: echo ::set-output name=url::https://${GITHUB_HEAD_REF}--open-wc-org.netlify.app/
+        id: preview
+      
+      - name: Build Documentation Site
+        env:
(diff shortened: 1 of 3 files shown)


Westbrook:
Looks like maybe I need a special setting in the build command to purge all the `localhost:8000` references? view-source:https://netlify-publish--open-wc-org.netlify.app/guides/knowledge/styling/index.html

@daKmoR what's Rocket expecting here?

bennypowers:
https://github.com/modernweb-dev/rocket/blob/main/packages/core/src/helpers/index.js

bennypowers:
those vars are provided in the netlify build runner
https://docs.netlify.com/configure-builds/environment-variables/#netlify-configuration-variables

bennypowers:
this might help? https://cli.netlify.com/commands/env

Westbrook:
Is there really no way to provide base URLs to Rocket other than those provided specifically by Netlify? Are consumers not expected to leverage any other tools for publishing their site?

Exposing those env vars manually seemed to help, but some URLs still get published at `localhost:8000`... See line 18: view-source:https://netlify-publish--open-wc-org.netlify.app/guides/knowledge/styling/index.html Thoughts?

bennypowers:
well that function from above is really just a plugin, so the sky's the limit, right?

Westbrook:
@daKmoR any ideas why Rocket process the canonical URL different than other URLs? `<link rel="canonical" href="http://localhost:8000/">` Can track this to a source, even with the `absoluteBaseUrl` corrected in other contexts...

.github/workflows/preview.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Build the preview URL from the branch name
+        run: echo ::set-output name=url::https://${GITHUB_HEAD_REF}--open-wc-org.netlify.app/
+        id: preview
+
+      - name: Publish to Netlify
+        env:
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+          PREVIEW_URL: ${{ steps.preview.outputs.url }}
```
daKmoR, line 39: according to [this](https://github.com/modernweb-dev/rocket/blob/main/packages/core/src/helpers/index.js#L13) it's called deploy url? what is preview URL?

and it should also set a context I think

how about this?

```suggestion
          DEPLOY_URL: ${{ steps.preview.outputs.url }}
          CONTEXT: deploy-preview
```

daKmoR (approved):

PR #2381, opened 2022-01-12
Fix the preview URLs

abdonrd:
## What I did

Fix the preview URLs.

It doesn't work in some URLs like:
https://github.com/open-wc/open-wc/pull/2066#issuecomment-1005795429

Slugified branch name from:
https://github.com/adobe/spectrum-web-components/blob/main/.circleci/config.yml#L152-L154


Westbrook (approved):
🥳 LGTM.

PR #2389, opened 2022-01-18
feat(scoped-element): work without polyfill if possible

daKmoR:
BREAKING CHANGE: Work without polyfill if possible (and do not auto load polyfill)

ScopedElementsMixin 2.x tried to be as convenient as possible by automatically loading the scoped custom elements registry polyfill.
This however led to a fatal error whenever you registered any component before ScopedElementsMixin was used.

The error especially happened when you would import a component applying ScopedElementsMixin into an existing application, fundamentally going against the "import and use" nature of web components.

Therefore, we decided to not load the polyfill inside the mixin, but let the developer (optionally) load it on top of his/her app.
This means that, depending on the app of the developer, the change can be breaking. After updating to this latest version, two scenarios will be possible:
1. **everything is fixed and less js is loaded than before**: the app doesn't need scoping (in this case the app either continues to work as is, or the fatal error mentioned above will disappear)
2. **the fatal error is replaced by a console error from ScopedElementsMixin**: the app needs scoping, because different versions of the same component are used (in this case, the fatal error mentioned above will disappear and a console error asking you to load the polyfill on top of your app will appear)

Only the second case requires an action. The remaining error can be resolved via a small migration path, a breaking bugfix if you will, which should be treated as if it were a security breaking change: consumers do the breaking bug fix on their end, because releasing a new major version would not be beneficial, neither to us nor to our consumers.

Not loading the polyfill by default will also allow sites to opt out of it altogether. This means until the browser ships scoped registries, the developer can choose to fall back to the global registry, by not loading the polyfill. This will save bandwidth & complexity since it doesn't need to be loaded by the client in that case.

All previous 2.x versions will be deprecated and scoped element will behave as follows:

1. If polyfill is not loaded it will use the global registry as a fallback
2. Log error if actually scoping is needed and polyfill is not loaded
3. If you manually create elements you will need to handle polyfilled and not polyfilled cases now

```diff
-  const myButton = this.shadowRoot.createElement('my-button');
+  const myButton = this.createScopedElement('my-button');
```

This also removes `@webcomponents/scoped-custom-element-registry` as a production dependency.

If you need scoping be sure to load the polyfill before any other web component gets registered.

It may look something like this in your HTML

```html
<script src="/node_modules/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js"></script>
```

or if you have an SPA you can load it at the top of your app shell code

```js
import '@webcomponents/scoped-custom-element-registry';
```

You need scoping if you want to

- use 2 major versions of a web component (e.g. in an SPA pageA uses 1.x and pageB uses 2.x of color-picker)
- or you want to use the same tag name with different implementations (use tag color-picker from foo here and from bar here)


patches/@lion+form-core+0.16.0.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/node_modules/@lion/form-core/src/validate/ValidateMixin.js
++++ b/node_modules/@lion/form-core/src/validate/ValidateMixin.js
+@@ -103,8 +103,8 @@ export const ValidateMixinImplementation = superclass =>
+       return {
+         ...super.slots,
+         feedback: () => {
+-          // @ts-ignore we load a polyfill to support createElement on shadowRoot
+-          const feedbackEl = this.shadowRoot.createElement('lion-validation-feedback');
++          const root = this.shadowRoot.createElement ? this.shadowRoot : document;
++          const feedbackEl = root.createElement('lion-validation-feedback');
```
daKmoR, line 12: this is here to make the portal search not break => once this is released and lion updated we can remove this and use the latest lion versions

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -86,20 +85,40 @@ const ScopedElementsMixinImplementation = superclass =>
...
-          this.registry.define(tagName, klass),
-        );
+        try {
+          this.registry = new CustomElementRegistry();
+        } catch (error) {
+          this.registry = customElements;
+          supportsScopedRegistry = false;
+        }
+        for (const [tagName, klass] of Object.entries(scopedElements)) {
+          const registeredClass = this.registry.get(tagName);
```
manolakis, line 104: I wonder if we should do the same check in the `defineScopedElement` method, as now they are sharing the same global registry and current implementation could cause silent issues.

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -86,20 +85,40 @@ const ScopedElementsMixinImplementation = superclass =>
...
 
-    /** @override */
     createRenderRoot() {
       const {
         scopedElements,
         shadowRootOptions,
         elementStyles,
       } = /** @type {typeof ScopedElementsHost} */ (this.constructor);
 
+      let supportsScopedRegistry = true;
```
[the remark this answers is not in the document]
daKmoR: nice idea 💪

packages/scoped-elements/src/ScopedElementsMixin.js
daKmoR: makes sense - will add 👍

daKmoR: moved logic and added a test

manolakis (approved):

.changeset/nervous-parrots-rush.md
```diff
@@ -0,0 +1,47 @@
+---
+'@open-wc/scoped-elements': minor
```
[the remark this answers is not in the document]
daKmoR: see the reasoning in the description - the most relevant part

> That is why we decided to treat this change as if it were a security-breaking change - consumers should do the breaking bug fix on their end, but releasing a new major version would not be beneficial, both to us and our consumers.

basically, all of 2.x releases are "bad"... loading the polyfill automatically was a bad call...

daKmoR:
A real-world reminder of why this change is important

> This is awesome news, most of my recent issues with Lion were related to this polyfill so making it an app concern will make my life a lot easier, thanks a lot :+1:

by Alan Dávalos

.changeset/nervous-parrots-rush.md
```diff
@@ -0,0 +1,47 @@
...
+
+BREAKING CHANGE: Work without polyfill if possible (and do not auto load polyfill)
+
+ScopedElementsMixin 2.x tried to be as convenient as possible by automatically loading the scoped custom elements registry polyfill.
+This however led to a fatal error whenever you registered any component before ScopedElementsMixin was used.
+
+This especially happens when you import a web component using ScopedElementsMixin into an existing application.
+
+This fundamentally breaks the way we think web components should work. We believe it should be "import and use".
+That is why we decided to treat this change as if it were a security-breaking change - consumers should do the breaking bug fix on their end, but releasing a new major version would not be beneficial, both to us and our consumers.
```
tlouisse, line 13:
```suggestion
The error especially happened when you would import a component applying ScopedElementsMixin into an existing application, fundamentally going against the "import and use" nature of web components.

Therefore, we decided to not load the polyfill inside the mixin, but let the developer (optionally) load it on top of his/her app.
This means that, depending on the app of the developer, the change can be breaking. Two scenarios will be possible:
1. the app doesn't need scoping (in this case the app either continues to work as is, or the fatal error mentioned above will disappear)
2. the app needs scoping, because different versions of the same component are used (in this case, the fatal error mentioned above will disappear and an error asking you to load the polyfill on top of your app will appear )

Only the second case is breaking: the fatal error can be resolved via a small migration path. A breaking bugfix if you will, which should be treated as if it were a security breaking change: consumers do the breaking bug fix on their end, because releasing a new major version would not be beneficial, neither to us nor to our consumers.
```

docs/docs/development/scoped-elements.md
```diff
@@ -70,12 +70,40 @@ Version 2 of Scoped Elements only supports [lit](https://lit.dev/) with `lit-ele
    }
    ```
 
-<inline-notification type="tip">
+5. (optional) load the polyfill if you need scoping
 
-ScopedElements loads a polyfill of the scoped registry for you . Why do we do this? The spec is quite mature but not yet implemented in a browser and the polyfill is pretty new so we want to control it to be able to patch things if needed to stay backward compatible.
-The polyfill we use is [@webcomponents/scoped-custom-element-registry](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry).
+   Defining sub elements via `scopedElements` is very useful on it's own as it makes it clear what your element requires. However, if you need the actual scoping feature for example to use two major version or two different classes with the same tag name then you will need to load a polyfill.
```
tlouisse, line 75:
```suggestion
   Defining sub elements via `scopedElements` is very useful on its own as it makes it clear what your element requires. However, if you need the actual scoping feature for example to use two major version or two different classes with the same tag name ,then you will need to load a polyfill.
```

docs/docs/development/scoped-elements.md
```diff
@@ -70,12 +70,40 @@ Version 2 of Scoped Elements only supports [lit](https://lit.dev/) with `lit-ele
...
+
+   or if you have an SPA you can load it at the top of your app shell code
+
+   ```js
+   import '@webcomponents/scoped-custom-element-registry';
+   ```
+
+   <inline-notification type="tip">
+
+   As long as you only use one version of a web component ScopeElementsMixin will work with the polyfill. So start of without the polyfill once you need it ScopeElementsMixin will log an error.
```
tlouisse, line 95:
```suggestion
   As long as you only use one version of a web component ScopeElementsMixin will work with the polyfill. So start of without the polyfill. Once you need it, ScopeElementsMixin will log an error.
```

tlouisse (approved):

lazka:
I wish this would have been a major bump. Now I have to revert it everywhere and block the update until I can figure out how to go forward..

daKmoR:
yes, we acknowledge that it is a controversial change - we still think it was the best option.

Non the less - we are sorry for the inconvenience caused.

Did you try adding the polyfill at app/website level then it should continue to work as is

```
npm i @webcomponents/scoped-custom-element-registry

// add to your app/website
<script src="/node_modules/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js"></script>
```

For more details please see

https://open-wc.org/blog/scoped-elements-without-polyfill/

why no semver? - the previous releases were "broken" and having a combination of 2.x and 3.x would still break horribly

> Therefore, we decided to not load the polyfill inside the mixin, but let the developer (optionally) load it on top of his/her app.
> This means that, depending on the app of the developer, the change can be breaking. After updating to this latest version, two scenarios will be possible:
> 1. **everything is fixed and less js is loaded than before**: the app doesn't need scoping (in this case the app either continues to work as is, or the fatal error mentioned above will disappear)
> 2. **the fatal error is replaced by a console error from ScopedElementsMixin**: the app needs scoping, because different versions of the same component are used (in this case, the fatal error mentioned above will disappear and a console error asking you to load the polyfill on top of your app will appear)
> 
> Only the second case requires an action. The remaining error can be resolved via a small migration path, a breaking bugfix if you will, which should be treated as if it were a security breaking change: consumers do the breaking bug fix on their end, because releasing a new major version would not be beneficial, neither to us nor to our consumers.

peschee:
@daKmoR Should there be a mention over at @lion as well regarding the polyfill inclusion?

https://lion-web.netlify.app/guides/how-to/get-started/

lazka:
> Did you try adding the polyfill at app/website level then it should continue to work as is

Yes, that's what I did where possible. But I don't control users of my web components. I'm now releasing new versions which are ported to use `createScopedElement()` to no longer crash and recommending the polyfill.

daKmoR:
@peschee what would you put into the getting started section?

Lion now works fine without the poylfill and there is no longer a chance of an fatal error just because you load lion (as long as you don't need scoping)

Yeah if you need scoping you need to load the polyfill but that is hopefully not the case as you get started? 🤔

Issue #2393, opened 2022-01-27
Regular code snippets disappear in mdjs


thepassle:
Given the following story:

`my-story.stories.js`:
`````md
```js script
export default {
  title: 'Foo',
};
```

# Hello

world

```js
console.log('code block doesnt show')
```

```js script
console.log('does execute')
```

`````

I would expect a code snippet to be output, but it seems like somewhere in the `mdjsToMd` function in `storybook-addon-markdown-docs` it gets removed. Interestingly, the `js script` does get executed.

![image](https://user-images.githubusercontent.com/17054057/151416112-c3827bc1-e53c-4adc-8a05-a0cc82aa4c02.png)

![image](https://user-images.githubusercontent.com/17054057/151415702-284364e0-cdb7-4b1f-91bb-b9f3db45e550.png)

I have a reproduction repo here: https://github.com/thepassle/sb-repro

## Expected behavior

Code snippets in mdjs files are output correctly

## Actual Behavior

Code snippets disappear

Westbrook:
The MDJS pipeline for Storybook is super out of date and has a number of vulnerabilities in its dependency chain. It would be great to get @daKmoR's thoughts on how to get it up-to-date with his latest work in this area to ensure we're preparing consumers for success here.

thepassle:
> to ensure we're preparing consumers for success here

Yeah, I'm updating some projects at work to modern web tooling, and there's quite a bit of usage of mdjs in those projects unfortunately. It'd be great not to have to refactor them to something else again.

daKmoR:
@thepassle thank you for the reproduction repo - I would have found a solution without it.

Can you check the 1 line change and verify that it solves it for your case - if so we can do a release right away.

@Westbrook that does not update any dependencies or so... - maybe have a separate issue for that?

thepassle:
Checked it locally and it indeed fixes it:

![image](https://user-images.githubusercontent.com/17054057/151935543-4774735c-466c-4597-aa25-54570103eac3.png)

Thanks for taking a look 🙂

PR #2394, opened 2022-02-01
fix(sb-md-docs): user provided code blocks should remain in final output

daKmoR:
## What I did

1. set sanitize to false (which was the default before but changed) => that should not be an issue as it's "revalidated" by storybooks mdx anyways

closes https://github.com/open-wc/open-wc/issues/2393


thepassle (approved):

thepassle:
thanks :)

image

Issue #2392, opened 2022-01-26
Page not found at https://open-wc.org/init/


bougault:
## Expected behavior

Page at https://open-wc.org/init/ should display details about "init" the project

## Actual Behavior

Page at https://open-wc.org/init/ is on 404 HTTP status

Link appears in CLI header when run `npm init @open-wc` command

## Additional context

![openwc-init](https://user-images.githubusercontent.com/7371900/151142725-964ec897-d76d-420c-83d8-689b923ac075.jpg)

Great 404 page though 👍

m4thieulavoie:
I hope it's okay I contributed [a PR](https://github.com/open-wc/open-wc/pull/2396) that should fix it 👍

PR #2396, opened 2022-02-02
fix: copy _redirects file to _site directory

m4thieulavoie:
Fixes https://github.com/open-wc/open-wc/issues/2392

## What I did

1. Copy the `_redirects` file to `_site` so that when being deployed over to netlify, the config follows along


Westbrook:
@daKmoR, I've not used redirects here before. Does this look good to you?

PR #2397, opened 2022-02-03
Make `oneEvent` generic

WickyNilliams:
Currently `oneEvent` is documented to always return `Promise<CustomEvent>`, but there is no reason it has to be `CustomEvent`. 

We can add a generic type param `TEvent`, constrained to `Event` to make this function more flexible. The default type is `CustomEvent` for backward compatibility

### DIFF
--- a/.changeset/hot-pets-applaud.md
+++ b/.changeset/hot-pets-applaud.md
@@ -1,0 +1,5 @@
+---
+'@open-wc/testing-helpers': patch
+---
+
+Add generics for oneEvent test helper function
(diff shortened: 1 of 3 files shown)


bennypowers:
Looks good but tsc is choking on the JSdoc. Can you type it up in a .d.ts and try importing from there?

WickyNilliams:
Ah! Didn't realise this was a fairly new TS/jsdoc feature. Seems like it was added in latest release if i test with different versions in the TS playground. I'll take a look at adding in a d.ts later today

WickyNilliams:
Finally got around to it. Hopefully this works!

PR #2398, opened 2022-02-04
Add "spread", "spreadEvents", and "spreadProps" directives

Westbrook:
## What I did

1. Add "spread", "spreadEvents", and "spreadProps" directives

### TO DO:
- [x] more test
- [x] docs
- [x] confirm types/build/test works in CI
- [x] tag @justinfagnani in to confirm that I've not done anything horrifically stupid here
- [x] look into a pre-release to confirm the types/build/publication work for consumers


bennypowers (approved):
Life's Good, Trust Me

bennypowers:
Changeset?

Westbrook:
Hey @justinfagnani would you be able to give this a once over for major foot guns? As we've previously discussed there's lots of things this likely doesn't do, or even doesn't take into account, but for the small scope it's working in I'd love to hear your thoughts.

Some things I wasn't sure if I should take into account...
- should I do something internally to check identity of the applied object or will `lit-html` not call into the `update` method in that case?
- Does `lit-html` doesn't bind and unbind event listeners when elements come on and off the document? I've done that here, but it might be work that I could save if that's not happening else where.
- I covered `all`, `events only`, and `props only` because they seemed the most useful, but would it make sense to have a `booleans only` usage and/or a helper to brand the properties of an array? Overkill?

bennypowers:
@justinfagnani @LarsDenBakker How do FAST and Hybrids' `html` tags handle this, if at all?

castastrophe:
I was thinking about the typing on the spread data object and was thinking something like this:

```
type SpreadData<T = {} | Element, K extends keyof T = any> = {
    [key in K]: T[K];
};
```

Would optionally allow you to pass an element or element class and derive the types from the generic input.

Westbrook:
Seems like adding to the complexity of the types internally is [unnecessary](https://www.typescriptlang.org/play?#code/C4TwDgpgBAymBOECGATAIk4SA8AVKAvFAN4C+UAPlAKIA2EAthAHbAA0UA0lBAB7AsUAZygBrCCAD2AMyj4iSZiAB8hEgCgoWqAG1xIKAEtmXALoAuOTs6mA3OtL31oSLEMMw9OIlRqAShAAxpLwKNhCwPDGAOYcAK7MosySAO7Myk7BzBFQAhEAjAAKiGBqABK4ALIAMnSMLMD2WTl5wPmW3sjomEhqxJpQAHTDrUUlbAOSwAAWEPC40zGW+QBMAMwALBOO6s3AuRARK31Q7VAA5LO0tJLnUDt7BxFrapcQ17dNktn7rRuvVxu50y3xyQncnggZxgEK8CC6alGXx+UHBHnoKw6sIgnV8RFaK2RYOxayx6Jx8LxT2AayJ+zRkI2ZMhuJQiMOwA2tiAA)? IIUC, leveraging that type of your data, would be accepted internal of the spread functionality just the same.

daKmoR:
I gave it a quick try using SSR and it does seem to support it 😭

```html
<rocket-header ${spread({ foo: 'bar' })}>
```

**expected outcome:**
```html
<rocket-header foo="bar">
```

**actual outcome:**
```html
<rocket-header lit$342275082$0="">
```

it somewhat makes sense as it imperatively sets the attributes... so I guess it would need special integration with lit-ssr?
thing is it even "breaks" hydration... e.g. once you use spread inside your component it can no longer be hydrated 🙈 (error is: `Uncaught Error: Unhandled shorter than expected iterable`)

@aomarks what are your thoughts?

The question is should this be a blocker? should this directive be client only? 🤔 are we ok with "breaking" SSR with it 🤔

PS: I currently have a use case in SSR... sooo from my point of view it would be really nice to support - but no clue how much effort that would be 😅

justinfagnani:
Tagging myself, @kevinpschaaf and @augustjk here...

Element-bindings are imperative only right now. In order for SSR to support them we'd need a declarative form - a value that an element directive can return that Lit can natively handle and render client-side and SSR. This unfortunately gets to the heart of the issues we haven't solved yet with spread: deduplication and order-dependence of bindings.

arekucr:
Any update on this?

daKmoR:
I think the SSR part will not be fixed any time soon... so as an action point I would say we should
1. make that clear in the docs
2. merge and release it

it will help in a lot of situations... just not in all 😅

daKmoR:
@Westbrook do you agree to add a not "does not support SSR" and then merge and release?

Pickra:
This is awesome @Westbrook, I'm sure alot of people are excited about it.
This will probably change, but currently I only want to pass HTML attributes that exist on an HTML element, which makes me think I should be using `spreadProps`. Not an ideal way to test, but I copy/pasted your [spread code](https://github.com/open-wc/open-wc/pull/2398/files#diff-036dd270c0e3cfadd0049ad31cc7214f43ecbb64c7ca49c4bbf9cf553c8e7bc9)...

`spread` rendered all the attributes, `spreadProps` didn't render **all** of the attributes. I'm not sure if that's intended or I missed something, my setup is wrong, not a good way to test...

```typescript
@customElement('cool-element')
export class CoolElement extends LitElement {
  @property({ type: Object }) attrs: { [key: string]: unknown; } = {};

  render() {
    return html`<p ${spread(this.attrs)}>cool bro</p>`;
  }
}

// Usage
const attrs = { id: 'dabest', role: '🔥', 'aria-disabled': 'true', tabindex: '0' }

html`<cool-element .attrs=${attrs}></cool-element>`;
```

`spreadProps` Rendered:
- id
- role

`spread` Render:
- tabindex
- aria-disabled
- id
- role

Westbrook:
I’ll try to make some time for the docs on this, soon. SSR is a big question, but simply not supporting it seems fine. The bigger piece missing is the last in realities of lit-html. If those things are OK, then sure let’s ship it. 

@Pickra if you used `spreadProps` for your test case, the issue is likely that post of those are NOT properties and can’t be spread as such.

Westbrook:
Updated and possibly ready to go???

PR #2409, opened 2022-02-20
Added Svelte to list of the base libraries

artemjackson:
## What I did

1. Added Svelte to list of the base libraries


Westbrook (commented):
While the library itself _can_ create WCs, it's not really meant for that, and many of the maintainers are actively against WCs and expanding the support of them within Svelte, so I'm not sure this is a good addition to our documentation on the subject.

PR #2410, opened 2022-02-23
Force lit-html dependency tree for correct types construction

Westbrook:
## What I did

1. Ensure that `lit-html@2.0` is the preferred version so that it can be depended on via bare module specifiers in types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "http-server": "^13.1.0",
     "husky": "^7.0.4",
     "lint-staged": "^10.5.4",
+    "lit-html": "^2.0.0",
     "mocha": "^8.3.2",
     "mocha-chai-snapshot": "^1.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -84,6 +84,7 @@
     "http-server": "^13.1.0",
     "husky": "^7.0.4",
     "lint-staged": "^10.5.4",
+    "lit-html": "^2.0.0",
```
daKmoR, line 87: why is that needed? 🤔

Westbrook: Other wise lit-html@1.0 becomes the “default” due to Haunted, Lion and Rocket et al and the types get built as `lit/nodes_modules/lit-html` which isn’t guaranteed to be where that dependency is in other projects. This way it gets built at `lit-html` and can be resolved more naturally by consuming projects.

daKmoR (approved):

PR #2411, opened 2022-02-23
Fulfill deprecation notice to simplify type delivery

Westbrook:
## What I did

1. Remove deprecated exports to prevent types collisions.

Merge/release after #2410 has been released.


packages/testing-helpers/index.js
```diff
@@ -16,5 +14,3 @@ export {
 
 export { litFixture, litFixtureSync } from './src/litFixture.js';
 export { stringFixture, stringFixtureSync } from './src/stringFixture.js';
-
-export { html, unsafeStatic };
```
daKmoR, line 20:
```js
import { html } from '@open-wc/testing';
```

is to my knowledge the only way to make sure that `html` matches with the `render` that testing expects...

e.g. it's the only way to be able to write a template fixture in lit 2 while my components are using lit 1 for example

thepassle: > is to my knowledge the only way to make sure that html matches with the render that fixture in testing expects...

But if your source code uses lit-html v1, or your storybook stories (which we reuse quite a bit for tests), that's not enough

PR #2412, opened 2022-02-23
Feat/customize render

thepassle:
## What I did

1.

### DIFF
--- a/packages/testing-helpers/test-web/fixture.test.js
+++ b/packages/testing-helpers/test-web/fixture.test.js
@@ -63,6 +63,12 @@ describe('fixtureSync & fixture', () => {
     testElement(elementAsync);
   });
 
+  it('supports customized render functions', async () => {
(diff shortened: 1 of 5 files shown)


packages/testing-helpers/test-web/fixture.test.js
```diff
@@ -63,6 +63,12 @@ describe('fixtureSync & fixture', () => {
     testElement(elementAsync);
   });
 
+  it('supports customized render functions', async () => {
+    const render = sinon.stub();
+    await fixture(html`<div></div>`, { render });
```
Westbrook, line 68: We’d need a test of this using `scopedElements` as it gets pretty fancy internally.

thepassle: decided to not support this usecase, as discussed on slack

daKmoR (approved):

PR #2419, opened 2022-03-03
Change return type from Assertion to Promise<Assertion>

chrispaterson:
## What I did

1. Changed return type of Equal from `Assertion` to `Promise<Assertion>`
2. Changed type `any` to `unknown`, which is preferred in TypeScript

Without this change, `tsserver`  gives the following warning:

`[tsserver 80007] [I] 'await' has no effect on the type of this expression.`  

Yet examples show `await`ing the assertions and the assertions must be awaited to work (in some cases).

### DIFF
--- a/packages/chai-a11y-axe/chai-a11y-axe-plugin.d.ts
+++ b/packages/chai-a11y-axe/chai-a11y-axe-plugin.d.ts
@@ -2,11 +2,11 @@
 
 declare namespace Chai {
   interface Assertion {
-    accessible(options?: Object): Assertion;
+    accessible(options?: Object): Promise<Assertion>;
   }
 
   interface Assert {
-    isAccessible(fixture: any, options?: Object): Assertion;
-    isNotAccessible(fixture: any, options?: Object): Assertion;
+    isAccessible(fixture: any, options?: Object): Promise<Assertion>;
+    isNotAccessible(fixture: any, options?: Object): Promise<Assertion>;
   }
 }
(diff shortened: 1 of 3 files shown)


packages/chai-a11y-axe/chai-a11y-axe-plugin.d.ts
```diff
@@ -2,11 +2,11 @@
...
   interface Assertion {
-    accessible(options?: Object): Assertion;
+    accessible(options?: Object): Promise<Assertion>;
   }
 
   interface Assert {
-    isAccessible(fixture: any, options?: Object): Assertion;
-    isNotAccessible(fixture: any, options?: Object): Assertion;
+    isAccessible(fixture: any, options?: Object): Promise<Assertion>;
+    isNotAccessible(fixture: any, options?: Object): Promise<Assertion>;
```
Westbrook, line 10: I wonder why I've never run into issues using `await expect(el).to.be.accessible();` in TS before. Do you think this is a general thing or something specific to your usage?

chrispaterson: I should mention that it's tsserver and it only gives a warning.  Regular compile from CLI with `tsc` doesn't have an issue

Westbrook: Hrm. I wonder what’s different between the two. Does this change work fine with the CLI? Haven’t had a chance to confirm.

castastrophe:
I would love to see this merged as well, is there anything I can do to help?

Westbrook (approved):
Let's do it!

PR #2432, opened 2022-04-05
fix(deps): Fix deps issues with @esm-bundle/chai

ianhuezo:
## What I did

1. @esm-bundle/chai has an issue if I use vite/vitest for web component testing.  The hot fix version in @esm-bundle/chai has those deps fixed.  

Also, created a repo that reproduces the failure here, https://github.com/ianhuezo/vitest-failing-imports/tree/main/

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@babel/core": "^7.13.16",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@changesets/cli": "^2.16.0",
-    "@esm-bundle/chai": "4.3.4-fix.0",
+    "@esm-bundle/chai": "^4.3.4-fix.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "@babel/core": "^7.13.16",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@changesets/cli": "^2.16.0",
-    "@esm-bundle/chai": "^4.3.4",
+    "@esm-bundle/chai": "4.3.4-fix.0",
```
Westbrook, line 46: Can we include the range here, again?

ianhuezo: Included the range again

Westbrook (approved):
LGTM! Thanks for the contribution.

PR #2434, opened 2022-04-12
fix(testing): esm-bundle/chai version update

ianhuezo:
## What I did

1. Adding changeset for the recent PR, https://github.com/open-wc/open-wc/pull/2432


ianhuezo:
No problem, I'm just happy to contribute a fix for something that I was having trouble with.

PR #2437, opened 2022-04-26
chore(semantic-dom-diff): update dependencies

bennypowers:
## What I did

1. bump dependencies in `semantic-dom-diff`

In monorepos, prevents

>  SyntaxError: The requested module './../../../node_modules/@web/test-runner-commands/browser/commands.mjs' does not provide an export named 'sendMouse' (at pfe-cta.spec.ts:6:1)


Westbrook (approved):
LGTM!

PR #2442, opened 2022-04-26
 fix: support node 17 & 18 by using md5 hashing

bennypowers:
## What I did

1. remove node 12 from ci matrix
2. add node 16 and 18 to ci matrix

node 12's EOL is in 4 days

https://nodejs.org/en/about/releases/

cc @maartenst


Westbrook:
Benny, did you cancel the CI run on this, or is there something as yet unaddressed here? https://github.com/open-wc/open-wc/runs/6303666611?check_suite_focus=true

web-padawan (approved):

PR #2446, opened 2022-05-05
docs: add LRNWebcomponents to component libraries

btopro:
Added link to LRNWebcomponents which contains HAX and 100s of other openly available, primarily Lit based elements.


Latanya541:
Open

docs/_data/componentLibraries.js
```diff
@@ -101,6 +101,12 @@ const componentLibraries = [
     description:
       'Lion web components is a set of highly performant, accessible and flexible Web Components. They provide an unopinionated, white label layer that can be extended to your own layer of components.',
   },
+  {
+    name: 'LRNWebcomponents',
+    url: 'https://github.com/elmsln/lrnwebcomponents',
```
Westbrook, line 106: Do you have a public URL that more directly points to the components in the library, and possibly how to consume them?

btopro: https://webcomponents.psu.edu/ isn't comprehensive but has a lot of them, you mean like that?

Westbrook: That definitely gave devs a bette rides of what they might be able to get from the library, yes

web-padawan (approved):

PR #2456, opened 2022-06-14
Check if `__registry` actually exists on the current class, not a parent

### LINKED ISSUE #2455: @open-wc/scoped-elements: subclassing components with different scoped elements fails
# Setup
Version 2.1.1 of `@open-wc/scoped-elements`.

Assume you have 2 web components:

```js
class Foo extends ScopedElementsMixin(LitElement) { 
  static get scopedElements() {
    return {
      'x-lorem': Lorem
    };
  }
}

class Bar extends Foo {
  static get scopedElements() {
    return {
      ...super.scopedElements,
      'x-ipsum': Ipsum
    };
  }
}
```

# Scenario
1. `<x-foo>` is registered & created
2. `<x-bar>` is registered & created

When the `createRenderRoot()` method on `ScopedElementsHost` is called, it will check if `this.constructor.__registry` exists and if not, it will create one and register the scoped elements on it. This works fine in step 1. But in step 2, `this.constructor.__registry` already is defined (on the parent Foo class).

So when `<x-bar>` is rendered, `<x-ipsum>` won't exist as a custom element. In the devtools you can see that `<x-ipsum>` does not have a shadow root.

If you turn the step order around, things work fine.

# Solution

A solution to this could be to check if `__registry` is actually a property of the current class (and not of a parent class):

```js
if (!Object.hasOwn(this.constructor, '__registry')) {
  // Create registry and register scoped elements
}
```

jpzwarte:
Fixes #2455

### DIFF
--- a/packages/scoped-elements/src/ScopedElementsMixin.js
+++ b/packages/scoped-elements/src/ScopedElementsMixin.js
@@ -95,7 +95,9 @@ const ScopedElementsMixinImplementation = superclass =>
         elementStyles,
       } = /** @type {typeof ScopedElementsHost} */ (this.constructor);
 
-      if (!this.registry) {
+      // Polyfill implementation for Object.hasOwn
+      // https://github.com/tc39/proposal-accessible-object-hasownproperty/blob/main/polyfill.js
+      if (!Object.prototype.hasOwnProperty.call(Object(this.constructor), '__registry')) {
         this.registry = supportsScopedRegistry ? new CustomElementRegistry() : customElements;
         for (const [tagName, klass] of Object.entries(scopedElements)) {
           this.defineScopedElement(tagName, klass);
--- a/.changeset/curly-rice-greet.md
+++ b/.changeset/curly-rice-greet.md
@@ -1,0 +1,5 @@
+---
+'@open-wc/scoped-elements': patch
+---
+
+Check if `__registry` actually exists on the current class, not a parent.


LarsDenBakker:
@manolakis what do you think?

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -95,7 +95,18 @@ const ScopedElementsMixinImplementation = superclass =>
         elementStyles,
       } = /** @type {typeof ScopedElementsHost} */ (this.constructor);
 
-      if (!this.registry) {
+      // @ts-ignore
+      const createRegistry =
+        !this.registry ||
+        (!Object.prototype.hasOwnProperty.call(this.constructor, '__registry') &&
```
manolakis, line 101: I would change the order to break the circuit earlier when the regitry is atteched to the instance instead of the class

```js
(this.registry === this.constructor.__registry && !Object.prototype.hasOwnProperty.call(this.constructor, '__registry'));
```

packages/scoped-elements/src/ScopedElementsMixin.js
```diff
@@ -95,7 +95,18 @@ const ScopedElementsMixinImplementation = superclass =>
         elementStyles,
       } = /** @type {typeof ScopedElementsHost} */ (this.constructor);
 
-      if (!this.registry) {
+      // @ts-ignore
+      const createRegistry =
```
manolakis, line 99: I would rename the variable to something like `shouldCreateRegistry` or `registryNotCreated` to reflect an intention or state instead of an action.

manolakis (commented):
Good catch @jpzwarte !

I just added some minor comments, but the code looks good to me. Thanks!

packages/scoped-elements/src/ScopedElementsMixin.js
jpzwarte: Done

packages/scoped-elements/src/ScopedElementsMixin.js
jpzwarte: Done

manolakis (approved):

PR #2464, opened 2022-07-07
Update index.md

davie-robertson:
link to starter app was pointing to old repo

## What I did

1. changed link in body text from `https://github.com/open-wc/open-wc-starter-app` to `https://github.com/open-wc/open-wc`


Westbrook:
While the starter-app itself is a bit out of date, I'm not sure the updated URL you've chosen here is correct, particularly in that it's not an "app" but the entire OWC project... not sure there's a "right" place to link this as we've gone all in on leveraging the generator as opposed to cloning a project.

davie-robertson:
I agree that this new link it to the whole project, but my logic was that the old link was pointing to the wrong place and the code there hadn't been changed/updated for something like 3 years - and the instructions that I was linking from still apply to the new link, i.e. how to use the generator.

Issue #2466, opened 2022-07-11
`eslint-plugin-lit-a11y` issue in `role-supports-aria-attr` rule


akornmeier:
## Expected behavior

Execute ESLint a11y rules against code with known issues and receive proper results.

## Actual Behavior

The `role-supports-aria-attr` rule appears to be missing a valid role type check that is causing rule logic to run on an undefined variable. Details below:

```
TypeError: Cannot destructure property 'props' of 'roles.get(...)' as it is undefined.
Occurred while linting /lit-element-starter-ts/src/my-element.ts:41
Rule: "lit-a11y/role-supports-aria-attr"
    at Object.enterElement (/lit-element-starter-ts/node_modules/eslint-plugin-lit-a11y/lib/rules/role-supports-aria-attr.js:46:32)
    at visit (/lit-element-starter-ts/node_modules/eslint-plugin-lit/lib/template-analyzer.js:257:29)
    at /lit-element-starter-ts/node_modules/eslint-plugin-lit/lib/template-analyzer.js:264:25
    at Array.forEach (<anonymous>)
    at visit (/lit-element-starter-ts/node_modules/eslint-plugin-lit/lib/template-analyzer.js:263:30)
    at TemplateAnalyzer.traverse (/lit-element-starter-ts/node_modules/eslint-plugin-lit/lib/template-analyzer.js:272:9)
    at TaggedTemplateExpression (/lit-element-starter-ts/node_modules/eslint-plugin-lit-a11y/lib/rules/role-supports-aria-attr.js:39:20)
    at ruleErrorHandler (/lit-element-starter-ts/node_modules/eslint/lib/linter/linter.js:1114:28)
    at /lit-element-starter-ts/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
```

akornmeier:
Issue fixed (https://github.com/open-wc/open-wc/pull/2472).

PR #2467, opened 2022-07-11
fix(eslint-plugin-lit-a11y): add missing valid role check in `role-supports-aria-attr` rule

akornmeier:
## What I did

1. Fixes issue #2466

### DIFF
--- a/.changeset/thin-onions-run.md
+++ b/.changeset/thin-onions-run.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-lit-a11y': patch
+---
+
+Fix bug #2466 issue in role-supports-aria-attr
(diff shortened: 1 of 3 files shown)


bennypowers (changes requested):
Can you add a regression test?

akornmeier:
> Can you add a regression test?

Yes definitely, but I was thinking of a way to do this while adhering to the current testing pattern. There are two rules that assume the use of a valid role in the package: 

1. [role-supports-aria-attr](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/docs/rules/role-supports-aria-attr.md)
1. [role-has-required-aria-attrs](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/docs/rules/role-has-required-aria-attrs.md)

The use of an invalid role `<div role="pizza">:pizza:</div>` would be caught by the [aria-role](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/docs/rules/aria-role.md) rule.

What technique would be best for testing the by-pass logic in `role-supports-aria-attr` that an invalid role would not throw an exception in the DOM?

akornmeier:
@bennypowers - Checking in after I added a regression test. Is there anything else required to get the PR merged? Thanks!

bennypowers (approved):

akornmeier:
Is there an ETA on the next release?

bennypowers:
cicd is stuck 🤦

bennypowers:
just YOLO'd it, hopefully it won't blow up 😅

PR #2479, opened 2022-08-09
fix(testing): support for Node16 TypeScript module resolution

zolk:
Similar to #2454, adds support for Node16 module resolution in TypeScript for the `testing` package.

Tested:
- Still works with "moduleResolution": "node"
- Works with "moduleResolution": "node16"

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -13,8 +13,7 @@
   },
   "author": "open-wc",
   "homepage": "https://github.com/open-wc/open-wc/tree/master/packages/testing",
-  "module": "index.js",
-  "type": "module",
+  "main": "index.js",
   "exports": {
     ".": "./index.js",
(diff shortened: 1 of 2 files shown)


packages/testing/package.json
```diff
@@ -13,8 +13,7 @@
   },
   "author": "open-wc",
   "homepage": "https://github.com/open-wc/open-wc/tree/master/packages/testing",
-  "module": "index.js",
-  "type": "module",
```
daKmoR, line 17: why remove the `"type": "module",`?

zolk: Removing this allows the package to be used in apps that use CommonJS w/ Node16 module resolution in TS. This same change was made to the `scoped-elements` package as part of #2454.

jpzwarte:
Fixed in https://github.com/open-wc/open-wc/pull/2507

Issue #2495, opened 2022-10-17
dedupe-mixin has DeprecationWarning  No "main" or "exports" field


mservaising:
At run time I get following error for dedupe mixin package
`[DEP0151] DeprecationWarning: No "main" or "exports" field defined in the package.json`

solution: Add export field.

PR #2496, opened 2022-10-17
fix: #2495 no export in package.json (fixing DeprecationWarning DEP0151)

mservaising:
## What I did

1. I added the export in the package.json of dedup-mixin package.

### DIFF
--- a/packages/dedupe-mixin/package.json
+++ b/packages/dedupe-mixin/package.json
@@ -32,5 +32,8 @@
     "dedupe",
     "mixins"
   ],
-  "sideEffects": false
+  "sideEffects": false,
+  "exports": {
+    ".": "./index.js"
+  }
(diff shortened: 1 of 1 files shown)


thepassle:
this is a breaking change

mservaising:
ok, just to understand, why it is a breaking change ?

thepassle:
Adding a package exports map is usually (if not always) a breaking change. In this case it will _only_ allow imports like `import { foo } from '@open-wc/dedupe-mixin'`. So if anybody imports it differently, for example `import { foo } from '@open-wc/dedupe-mixin/index.js'`, that will now no longer work

mservaising:
then have you any other fix for the issue I got above (using ing-web 3) but I suppose it is the same with lion

PR #2505, opened 2022-10-26
Nord Design System component library

tpluscode:
## What I did

1. add Nord Design System component library


Westbrook (approved):
LGTM! Thanks 🙏

PR #2532, opened 2022-11-18
feat(eslint-config): bump eslint-plugin-html to 7.1.0

web-padawan:
## What I did

1. Upgraded `eslint-plugin-html` to `7.1.0` (the only breaking change in v7 is that [Node 10 support is dropped](https://github.com/BenoitZugmeyer/eslint-plugin-html/blob/main/CHANGELOG.md)).


Westbrook (approved):
LGTM!

PR #2535, opened 2022-11-18
chore: bump lint-staged and husky to latest versions

web-padawan:
## What I did

1. Updated `lint-staged` to v13 that [dropped Node 12 support](https://github.com/okonet/lint-staged/releases/tag/v13.0.0),
2. Updated `husky` to v6 that also [dropped Node 12 support](https://github.com/typicode/husky/releases/tag/v8.0.0).

Now when we no longer use Node 12 in CI, let's update these.


Westbrook (approved):
LGTM!

PR #2567, opened 2023-01-21
Update getting-started.md

jgustavoas:
`lit-html` and `lit-element` refer to an older version of `Lit` library.

## What I did

1. I updated the reference to the base library in the last section of the text. The original document still mentions `lit-html` and `lit-template`, but they were upgraded to `Lit`  in the `@open-wc` generator.

### DIFF
--- a/docs/guides/developing-components/getting-started.md
+++ b/docs/guides/developing-components/getting-started.md
@@ -1,4 +1,4 @@
-# Developing Components: Getting Started
+# Developing Components >> Getting Started ||10
 
 To get started building web components, we recommend using our generator to set up your project.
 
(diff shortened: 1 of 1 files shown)


docs/guides/developing-components/getting-started.md
```diff
@@ -1,4 +1,4 @@
-# Developing Components >> Getting Started ||10
+# Developing Components: Getting Started
```
Westbrook, line 1: If you could revert this change, we'd be happy to accept the changes below.

jgustavoas: Hi, Westbrook! I found the highlighted change on line 1 by clicking '**Edit this page on GitHub**' on the bottom of the [page](https://open-wc.org/guides/developing-components/getting-started/), but not in the version of the document that I changed on Jan 20. I'm afraid to create an unnecessary pull request trying to get the first one accepted.

Westbrook:
```suggestion
# Developing Components >> Getting Started ||10
```

Westbrook (approved):

PR #2579, opened 2023-02-21
Start a series of post about testing web components with @web/test-runner

Westbrook:
## What I did

1. Wrote a blog post. I'm not sure whether it best belongs here or on moder-web.dev or both, but I'd likely also post it to web.dev at a similar time as here, if yall like the context, etc. etc.
2. After this is done, I'd like to dig into some specific areas of testing as outlined in the post, so hopefully I can get a little velocity under me here.

## TO DO
- [ ] Run grammarly
- [ ] Profit?


Westbrook:
@web-padawan, if you wanted to get a preview of what I'm working on, I'd love any notes you might have on this...

PR #2580, opened 2023-02-24
Update componentLibraries.js

Marshal27:
## What I did

  1. Add `shadow-container-query-polyfill` to componentLibraries.js

### DIFF
--- a/docs/_data/componentLibraries.js
+++ b/docs/_data/componentLibraries.js
@@ -124,6 +124,12 @@ const componentLibraries = [
     description:
       'Shoelace provides a collection of professionally designed, every day UI components built on a framework-agnostic technology. Every web component is built with accessibility and developer experience in mind.',
   },
+   {
+    name: 'Shadow Container Query Polyfill',
+    url: 'https://github.com/Marshal27/shadow-container-query-polyfill',
+    description:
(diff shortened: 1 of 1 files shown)


web-padawan (commented):
Thanks for the PR. It's great to see more stuff developed around Web Components 👏 
 
This looks interesting, but the section is definitely wrong as it's not a component library.
We could consider mentioning it under a different page. @daKmoR @thepassle WDYT?

web-padawan:
@Marshal27 Added your polyfill to [Web Components bookmarks](https://webcomponents.today/polyfills-and-shims/) - see https://github.com/web-padawan/webcomponents.today/pull/38.

Marshal27:
@web-padawan I appreciate you adding a reference to `Web Components bookmarks`. 

Regarding this PR, I took a chance by creating it, in hopes the team would route it to the appropriate place if needed. (I appreciate you doing that.)

My logic was that this list includes several UI/UX design systems, and I thought adding the polyfill here may be helpful as the polyfill could complement the implementation of those design systems if used in concert with one of them.

In the end, my goal was to bring visibility to the polyfill for those who may be interested in using it. I appreciate you, and the team, for any references you all may feel appropriate.

Westbrook:
We don't really have a section for polyfills right now, and I;m not sure where we'd put one one or what else we'd list, so I'm gonna close this for now.

PR #2584, opened 2023-03-10
Update lint-staged and husky examples

rivajunior:
## What I did

1. Remove `gid add` because it isn't necessary anymore. See lint-staged [v10 changelog](https://github.com/okonet/lint-staged#v10)
2. Change the husky example to use a command to create the desired hook


Westbrook (approved):
LGTM. Thanks!

PR #2663, opened 2023-04-24
Use released versions instead of canary

koddsson:
Hey @Westbrook!

Here are the released versions instead of the canary version. If this works, maybe we can merge your PR?


Westbrook (approved):
LGTM!

PR #2667, opened 2023-05-01
fix: getScopedTagName bug

thepassle:
## What I did

In v1, the ScopedElementsMixin would add a random ID to your custom elements, like for example: `<ing-feat-foo-43539>` as the scoping mechanism to avoid duplicate custom elements being defined, which results in runtime errors. In @open-wc/scoped-elements v2, the ScopedElementsMixin has been aligned more closely with the Scoped Custom Element Registries proposal, which uses a CustomElementRegistry as the scoping mechanism, so we no longer have to add a random ID to the custom elements.

ScopedElementsMixin v1 provided a getScopedTagName method, to easily get the generated tagName for a given element, for example:

```js
this.getScopedTagName('ing-feat-foo'); // 'ing-feat-foo-43439'
```

This method has been deprecated in @open-wc/scoped-elements v2 (and frankly, should have just been removed in v2). In v2, `getScopedTagName` just returns the `tagName`:

```js
getScopedTagName(tagName) {
  return tagName;
}
```

Which is a bit of a problem, because this causes the following statement:

```js
if(!this.getScopedTagName(tag)) {
  this.defineScopedElement(tag, klass);
}
```

To return `true` in v1, and
To return `false` in v2

This change makes the behavior of `getScopedTagName` more in line with how it was in v1.

### DIFF
--- a/.changeset/shaggy-geckos-smash.md
+++ b/.changeset/shaggy-geckos-smash.md
@@ -1,0 +1,5 @@
+---
+"@open-wc/scoped-elements": patch
+---
+
+fix: getScopedTagName bug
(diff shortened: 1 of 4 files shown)


LarsDenBakker (approved):
Looks good to me, maybe describe a bit more about the bug for future reference?

PR #2677, opened 2023-05-28
[WIP / Discussion] Allow for non-lit specific implementation

KonnorRogers:
## What I did

1. Moved the `adoptStyles` call to be a hook

## Why?

Ideally I would love to be able to use this mixin with any web component, and as far as I can tell this is really the only Lit-specific portion.

Technically this PR doesn't solve the problem, but if the base mixin has a hook to adopt styles, we could get a separate entrypoint for the Mixin for Lit-specific implementations with the actual `adoptStyles` implementation.

I didn't do it here because I figured I'd open it up to discussion first if this is a use-case you all would like to support and making sure I didn't miss anything along the way.

### DIFF
--- a/packages/scoped-elements/src/ScopedElementsMixin.js
+++ b/packages/scoped-elements/src/ScopedElementsMixin.js
@@ -123,8 +123,7 @@ const ScopedElementsMixinImplementation = superclass =>
       }
 
       if (createdRoot instanceof ShadowRoot) {
-        adoptStyles(createdRoot, elementStyles);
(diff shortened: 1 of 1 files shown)


thepassle:
The ScopedElementsMixin is indeed built for usage with LitElement, and depends on overriding lits `createRenderRoot` lifecycle method. Moving the `adoptStyles` function isn't going to be enough to make the mixin compatible with non-Lit elements, unless they also implement a similar lifecycle hook. I also don't think the `adoptStyles` function is actually an issue here, I think it would just noop when used with non-lit elements.

PR #2679, opened 2023-06-04
Update lit-html.md

### LINKED ISSUE #2678: Lit: intermediate -> Add a visited/not-visited filter. Little mistake in the code
In it: intermediate -> Add a visited/not-visited filter -> View final result line 5

static  properties{

should be

static  properties={

xurdedix:
fix #2678 

I have only added the equal that was missing

a minimal contribution to his excellent work. Thank you


thepassle:
Super nitpick: would you mind removing the double space between `static  properties` as well while we're here? 🙂

Issue #2577, opened 2023-02-07
Test with Chai@5.0


Westbrook:
https://github.com/chaijs/chai/discussions/1500 announces an initial beta of Chai@5.0 which move to ESM so we could get back on the main build of the project and benefit from dropping support for lots of old browsers...

PR #2697, opened 2023-07-17
fix(chai-dom-diff): scope promise result to dom assertions

### LINKED ISSUE #2675: [sementic-dom-diff] Typings added in 0.19.8 interfere with all Should assertions
## Expected behavior
When i have a test that doesn't use semantic-dom-diff I expect the should assertion not to be asynchronous 

## Actual Behavior
An equal assertion to be synchronous 

## Describe what happens instead.
When using eslint it picks up the typings from [chai-dom-diff-plugin.d.ts](https://github.com/open-wc/open-wc/blob/fa438178c08855865c154e7f060da3126ad0b15f/packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts#LL32C1-L35C6) and reports async calls should be awaited.

`error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the void operator.`

## Additional context
See eslint result in the console output of this stackblitz project:
https://stackblitz.com/edit/semantic-dom-diff-typeings?file=src%2FArithmetic.test.ts

Which packages and versions are you using.

@open-wc/testing: "3.1.8",
@typescript-eslint/eslint-plugin: "^5.51.0",
@typescript-eslint/parser: "^5.51.0",
eslint: "^8.40.0",
eslint-plugin-lit: "^1.8.3",

Also see: [package.json](https://stackblitz.com/edit/semantic-dom-diff-typeings?file=package.json)

bennypowers:
Fixes #2675 

## What I did

1. Remove our augmentations from Chai's global namespace
2. Extend Chai's namespace for our methods
3. Re-implement language chains on our namespace objects

step 3 there is particularly brittle. I have my :crossed_fingers: that chai 5 will be out before this becomes a real problem.

## Testing instructions

Please test with both TDD/BDD styles (`assert`/`expect`). I typically use `expect` so I didn't exhaustively test `assert.`

### DIFF
--- a/.changeset/rude-eels-switch.md
+++ b/.changeset/rude-eels-switch.md
@@ -1,0 +1,5 @@
+---
+'@open-wc/semantic-dom-diff': patch
+---
+
+Refines TypeScript typings for DOM assertions
--- a/packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
+++ b/packages/semantic-dom-diff/chai-dom-diff-plugin.d.ts
@@ -5,33 +5,46 @@ import { DiffOptions } from './get-diffable-html';
 declare global {
   namespace Chai {
     interface Assertion extends LanguageChains, NumericComparison, TypeComparison {
-      dom: Assertion;
-      lightDom: Assertion;
-      shadowDom: Assertion;
+      dom: DomDiffAssertion;
+      lightDom: DomDiffAssertion;
+      shadowDom: DomDiffAssertion;
+      equal: DomDiffEqual;
+      equals: DomDiffEqual;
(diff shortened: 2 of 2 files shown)


JackRobards:
Would it be possible to move forward with this PR? From my testing on a large project these new types work great at fixing the issue!

It is a bit frustrating to use the `@open-wc/testing` library and Chai assertions in TypeScript tests without a fix for this. So many assertions get incorrectly marked as returning promises which can be confusing.

bennypowers:
might be better to invest in upgrading to chai 5, which has been out for quite some time

see #2577 #2795

PR #2706, opened 2023-08-26
Add new testing blog post

Westbrook:
## What I did

1. Wrote the second post about testing with the OWC generator and Web Test Runner.

Would love your thoughts. "Scheduled" it for Friday, but all things can be edited.


docs/blog/testing-with-js-is-like-magic/index.md
```diff
@@ -0,0 +1,226 @@
+---
+title: 'Testing with JS is like magic, but is it science?'
```
bennypowers, line 2: Better as a subtitle

Westbrook: The whole thing? Or after the comma? or?

docs/blog/testing-with-js-is-like-magic/index.md
```diff
@@ -0,0 +1,226 @@
...
+});
+```
+
+This test confirms that not only is `counter` a getable property as confirmed in 'has a default title "Hey there" and counter 5' and the original version of 'increases the counter on button click', but it is also settable. Being this property leverages the [`@property` decorator](https://lit.dev/docs/api/decorators/#property) from [Lit](https://lit.dev), I can tell you for sure that it does have a setter, though our tests certainly hadn't guaranteed that previously. So, we'd be adding robustness to the generated code if we chose to add something like this. I would say that testing a setter with its getter _might_ leave something to be desired. Once we get deep into them in subsequent articles, confirming that the new value of `counter` is leveraged by the rendered DOM via [snapshot testing](https://modern-web.dev/docs/test-runner/commands/#snapshots) (not my favorite, I'll share more about that later) or actually rendered to the page via [visual regression testing](https://github.com/modernweb-dev/web/tree/master/packages/test-runner-visual-regression) (key for UI libraries) would be a valuable extension of the test here.
+
+#### `counter` can be increases
+
```
TarunAdobe, line 153: line 152 - is this a typo or intended?

TarunAdobe (changes requested):

PR #2707, opened 2023-09-27
feat: add getScopedElementsTemplate to publish

RoXuS:
## What I did

Added `getScopedElementsTemplate` to publish to use it with litFixtureSync function.


Westbrook:
Can you please share some extended usage of this so I can understand what this is unlocking?

RoXuS:
Of course, if we use fixtureSync with scoped elements like this:
```
    const el = fixtureSync<DtContractHeader>(fixtureData, {
      scopedElements: {
        'dt-contract-header': DtContractHeader,
      },
    });
```
Then we have this error:
image

fixtureSync call litFixtureSync without `getScopedElementsTemplate`.

```
export function litFixtureSync(template, options = {}, getScopedElementsTemplate) {
...
```

jamiejojo:
Why do I keep receiving messages from GitHub’s
Jamie

Westbrook:
There are a bunch of tests that fail with this change. Please take a closer look so we can fully review this update.

See: https://github.com/open-wc/open-wc/actions/runs/6434919239/job/17475126467?pr=2707

PR #2710, opened 2023-10-02
fix(scoped-elements): add type module to pkg json

jorenbroekema:
## What I did

1. add type module field to pkg json

I checked all other packages, this one is the only one missing from the looks of it


thepassle:
I think this is technically a breaking change, but I think considering this is a browser-only module it should probably be fine? We can revert it if anyone runs into any issues

also prettier failed apparently, can you run format?

jorenbroekema:
Fixed!

thepassle (approved):

PR #2714, opened 2023-10-18
update material web links and info

e111077:
## What I did

1. Changed the link to material web library
2. Changed the name to the new branding (Material web (MWC))


e111077:
No changeset required because i didn't change one of the main packages

e111077:
U R amazing @thepassle

PR #2723, opened 2023-10-25
Allow lit 3

alexgagnon:
## What I did

1. Update peerDependencies to include Lit@3
2. Fix issue in `packages/codelabs/src/intermediate/lit-html.md` that was causing failures in prettier
3. Fix issue in `packages/dev-server-hmr/src/hmrPlugin.js that was causing failures in tsc
4. Provide files for testing against both Lit@2 and Lit@3

### DIFF
--- a/packages/lit-helpers/package.json
+++ b/packages/lit-helpers/package.json
@@ -35,10 +35,10 @@
     "helpers"
   ],
   "peerDependencies": {
-    "lit": "^2.0.0"
+    "lit": "^2.0.0 || ^3.0.0"
   },
   "sideEffects": false,
-  "dependencies": {
+  "devDependencies": {
     "lit": "^3.0.0",
     "lit2": "npm:lit@2"
   }
--- a/packages/lit-helpers/package.json
+++ b/packages/lit-helpers/package.json
@@ -25,7 +25,8 @@
     "*.d.ts",
     "*.js",
     "src",
-    "!src/spread.ts"
+    "!src/spread.ts",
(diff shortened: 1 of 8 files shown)


packages/lit-helpers/package.json
```diff
@@ -36,5 +37,9 @@
   "peerDependencies": {
     "lit": "^2.0.0"
```
Westbrook, line 38: Do we actually need to vend multiple exports/add multiple hard dependencies, or can we just loosen this `peerDependency` to include `^2.0.0 || ^3.0.0` so that the consuming project can continue to outline the versions of the dependencies they need rather than forcing both versions that the change does now?

alexgagnon: **sigh** I switched laptops and forgot to put the only change that was actually needed... new commit updates the "peerDependencies", and I updated "dependencies" to "devDependencies"

JosefJezek:
any progress?

Westbrook:
This _should_ have been fast forwarded as part of other changes. Can you check the latest versions?

alexgagnon:
> any progress? Why do you create the spread for lit 2, file spread-lit2.js?

spread-2.ts references lit@2 while spread.ts is updated for lit@3 so that both major versions can be tested. 

I doesn't look like this has made it into upstream yet, and this is getting out of date. I'll close this and open another one with most recent changes

Westbrook:
This should be working in currently published versions. If not feel free to reopen a new, updated PR based on the current state of the library. Many thanks!

PR #2726, opened 2023-10-31
Use split versions for all lit dependencies

Westbrook:
## What I did

1. Split all versions of `lit` dependencies for consumption of ^2 || ^3.


koddsson (approved):
Looks good to me :)

PR #2733, opened 2023-11-02
feat: scoped elements v3

thepassle:
## What I did

1.


bashmish:
Amazing work @thepassle 🎉

thepassle:
> Amazing work @thepassle 🎉

@pndewit too :)

bashmish:
Then kudos to @pndewit too 💯

lazka:
How should already published web components be handled now regarding the polyfill? Should we bundle the polyfill with the web component, or do a major bump for all published web components and require users to add a polyfill?

thepassle:
If you're using `@open-wc/scoped-elements@^2.0.0`, that version has built-in detection to see if the polyfill is loaded or not. If the polyfill is not loaded, it will fallback to the global customElements registry. If the polyfill is loaded, it uses the polyfill. In version `3.0.0`, we made the polyfill required to use. So if you use components that use `@open-wc/scoped-elements@^3.0.0`, you'll have to load the polyfill.

Polyfills should always be loaded by the consuming application (polyfilling is an app-level concern), and never bundled with a library or component.

lazka:
> Polyfills should always be loaded by the consuming application (polyfilling is an app-level concern), and never bundled with a library or component.

ok, so this means a breaking change for all components :(

thanks for the clarification

thepassle:
Yes, updating from v2.0.0 to v3.0.0 is a semver major change, thats also why it was released as such. If youre happy using v2.0.0 of the mixin, you can just keep on using that version instead.

lazka:
> Yes, updating from v2.0.0 to v3.0.0 is a semver major change, thats also why it was released as such.

Just for clarification: I meant it's a breaking change for all my components if I upgrade the mixin, since now my component users now also require the polyfill, while previously it was optional.

> If youre happy using v2.0.0 of the mixin, you can just keep on using that version instead.

Yeah, I think I'll do that for now. Here is hoping that someday we'll get browser support :)

lazka:
> Indeed. Thats why the mixin was released as a major version.

Yes, but usually major releases of my internal dependencies don't force me to also release new major versions. This is the first time for me at least.

thepassle:
> major releases of my internal dependencies don't force me to also release new major versions

This sometimes happens, sometimes not, but its definitely not uncommon. In this case it does require a major version, because the change is indeed breaking. As such, it has been released as a major version. As mentioned above, if you're happy with using v2.0.0, you're free to keep using that one.

clukhei:
@thepassle  is there a roadmap for how long v2 will be supported until we have to move to v3? I am in the same situation as lazka and thinking whether i should move to v3  or stay at v2. My concern is for the event that some dependencies in v2 needs update (like security patches), will scope-elements v2 be updated?

lazka:
(fyi, I've decided to stay on v2, and if needed I'll vendor the old version, since it's only one small file anyway: https://github.com/open-wc/open-wc/blob/%40open-wc/scoped-elements%402.2.4/packages/scoped-elements/src/ScopedElementsMixin.js)

thepassle:
> @thepassle is there a roadmap for how long v2 will be supported until we have to move to v3? I am in the same situation as lazka and thinking whether i should move to v3 or stay at v2. My concern is for the event that some dependencies in v2 needs update (like security patches), will scope-elements v2 be updated?

We still have the v2 version on a separate branch, so if any security patches or critical bugfixes need to be done we can still patch those. Note that v2 barely has any dependencies as is, though and the ones it does have are generally very stable ones, so I don't foresee any problems there.

I'll reiterate that if you're happily using v2, it's fine to just keep using v2 if it suits your usecases. Its also fine to vendor it yourself like @lazka mentioned, but I don't necessarily see the merit of that immediately 🙂

PR #2736, opened 2023-11-03
chore(testing): remove unused dependencies

pndewit:
## What I did

1. Removed the deprecated and unused @open-wc/chai-dom-equals dependency

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -43,10 +43,8 @@
   ],
   "dependencies": {
     "@esm-bundle/chai": "^4.3.4-fix.0",
-    "@open-wc/chai-dom-equals": "^0.12.36",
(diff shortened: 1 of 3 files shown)


packages/testing/package.json
```diff
@@ -43,10 +43,8 @@
   ],
   "dependencies": {
     "@esm-bundle/chai": "^4.3.4-fix.0",
-    "@open-wc/chai-dom-equals": "^0.12.36",
     "@open-wc/semantic-dom-diff": "^0.20.0",
     "@open-wc/testing-helpers": "^2.3.1",
-    "@types/chai": "^4.2.11",
```
thepassle, line 49: Is this really unused?

pndewit: We only use the `@types/chai-dom` types, so yes :)

PR #2748, opened 2023-11-08
add function for safely setting properties

break-stuff:
## What I did

1. Added a function with a `try/catch` to prevent errors when setting properties from breaking applications


package.json
```diff
@@ -16,8 +16,9 @@
     "prepare": "husky install",
     "release": "changeset publish && npm run format",
     "search": "rocket search",
-    "setup": "npm run setup:ts-configs && npm run setup:patch && npm run setup:postinstall",
+    "setup": "npm run setup:ts-configs && npm run setup:patch && npm run setup:playwright && npm run setup:postinstall",
     "setup:patch": "npx patch-package",
+    "setup:playwright": "npx playwright install",
```
Westbrook, line 21: Why was this required?

packages/lit-helpers/src/spread.ts
```diff
@@ -186,68 +184,80 @@ export const spreadEvents = directive(SpreadEventsDirective);
...
+}
+
+function safeSetProperty(element: Element, name: string, value: unknown) {
+  if (hasSetter(element, name)) {
+    element[name] = value;
+  } else {
+    console.warn(
+      `Could not set property "${name}" on ${element.tagName} because it has no "setter".`,
+    );
+  }
```
Westbrook, line 256: We probably want to do something like https://github.com/thepassle/app-tools/tree/master/env so that console warnings can be trimmed from the final deployment in a production environment.

packages/lit-helpers/src/spread.ts
```diff
@@ -186,68 +184,80 @@ export const spreadEvents = directive(SpreadEventsDirective);
...
+}
+
+function hasSetter(element: Element, name: string) {
+  return !!Object.getOwnPropertyDescriptor(element.constructor.prototype, name)?.set;
```
Westbrook, line 260: 👍🏼

packages/lit-helpers/test-web/spread.test.js
```diff
@@ -45,6 +50,7 @@ describe('spread', () => {
         number,
         '.array': array,
         '.object': object,
+        '.validity': undefined,
```
Westbrook, line 53: What's the actual expectation associated with this?

package.json
break-stuff: For some reason Playwright changed their package to no longer install the browser instances with their package and now they need to be manually installed. This includes it as part of the setup. I can remove it, but nevs will need to run `npx playwright install` after each installation to ensure the browser instances are in sync the Playwright version.

packages/lit-helpers/src/spread.ts
break-stuff: Perfect! I was thinking the same thing, but wasn't sure if there was already a mechanism in place for this.

packages/lit-helpers/test-web/spread.test.js
break-stuff: With this being set, it should not crash the application, but I can't get the tests to run for some reason. Are you able to for the lit helpers? They didn't seem to be set up at all.

jamiejojo:
Wtf is this

Jamie

break-stuff:
Was this comment meant for something else?

PR #2762, opened 2024-01-11
Added CodeStitch as a library to componentLibraries.js

ItsEthanH:
Accidently deleted the head repo for the last time I opened this. My bad!

What I did

    Added CodeStitch as a library to componentLibraries.js

Think I followed all the correct processes here. Let me know if I can help any more!


Westbrook:
I don't immediately see any custom elements in the link listed herein. If you want to recreate this with a more appropriate link and description we'd be happy to take another look.

PR #2763, opened 2024-01-11
Replace context.getSourceCode with context.sourcecode 

nikhilhundare:
## What I did

1.Eslint has released the alpha version which consists a breaking change - This version has deprecated context.getSourceCode() instead we can use context.sourceCode
2. This PR consists this change with backward compatibility (for version < 9)

### DIFF
--- a/packages/eslint-plugin-lit-a11y/lib/rules/valid-lang.js
+++ b/packages/eslint-plugin-lit-a11y/lib/rules/valid-lang.js
@@ -9,6 +9,7 @@ const { TemplateAnalyzer } = require('eslint-plugin-lit/lib/template-analyzer');
 const { getLiteralAttributeValue } = require('../utils/getLiteralAttributeValue.js');
 const { HasLitHtmlImportRuleExtension } = require('../utils/HasLitHtmlImportRuleExtension.js');
 const { isHtmlTaggedTemplate } = require('../utils/isLitHtmlTemplate.js');
+const { getContextSourceCode } = require('../utils/getContextSourceCode.js');
 
 const ValidLangRule = {
   meta: {
@@ -40,13 +41,13 @@ const ValidLangRule = {
                   analyzer,
                   element,
                   'lang',
-                  context.getSourceCode(),
+                  getContextSourceCode(context),
                 );
 
                 if (!lang) {
                   const loc = analyzer.resolveLocation(
                     element.sourceCodeLocation.startTag,
-                    context.getSourceCode(),
+                    getContextSourceCode(context),
                   );
 
                   return context.report({
@@ -58,7 +59,7 @@ const ValidLangRule = {
                 // We cast this as string, as lang could be a number, since getLiteralAttributeValue returns a primitive, this would throw an error as `language-tags` performs string methods on the argument
                 if (!tags.check(`${lang}`)) {
                   const loc =
-                    analyzer.getLocationForAttribute(element, 'lang', context.getSourceCode()) ??
+                    analyzer.getLocationForAttribute(element, 'lang', getContextSourceCode(context),) ??
                     node.loc;
 
                   return context.report({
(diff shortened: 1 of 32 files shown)


packages/eslint-plugin-lit-a11y/lib/utils/getContextSourceCode.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+function getContextSourceCode(context) {
+  try {
+    if (context.getSourceCode()) {
+      return context.getSourceCode();
+    }
+  } catch (e) {
+    return context.sourceCode;
+  }
+}
```
koddsson, line 13: Maybe we can add some more context on why this is needed and when we can remove it? I would also like to not throw and catch errors if possible.

```suggestion
/**
 * A backwards compatible way to get the source code of a given context.
 */
function getContextSourceCode(context) {
  return context.sourceCode ?? context.getSourceCode();
}
```

Is this now small enough to just inline into each of the rules? What do you think?

packages/eslint-plugin-lit-a11y/lib/rules/valid-lang.js
```diff
@@ -58,7 +59,7 @@ const ValidLangRule = {
                 // We cast this as string, as lang could be a number, since getLiteralAttributeValue returns a primitive, this would throw an error as `language-tags` performs string methods on the argument
                 if (!tags.check(`${lang}`)) {
                   const loc =
-                    analyzer.getLocationForAttribute(element, 'lang', context.getSourceCode()) ??
+                    analyzer.getLocationForAttribute(element, 'lang', getContextSourceCode(context),) ??
```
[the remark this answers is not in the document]
nikhilhundare: I think the formatter adds it

packages/eslint-plugin-lit-a11y/lib/utils/getContextSourceCode.js
nikhilhundare: I had tried this before, it breaks the execution with error "Function does not exist" and hence I added it in a try catch. I have updated the description of the method

koddsson: Hmm, but since we are checking the `context.sourceCode` now first and falling back to `context.getSourceCode()` shouldn't that still work?

justinfagnani:
Drive by comment here: it would be great to have the plugin tested in both eslint 8 and 9. I've done this before with typescript versions, and it's not the most trivial thing, but not that bad with typescript at least. That could be something done later of course.

packages/eslint-plugin-lit-a11y/lib/utils/getContextSourceCode.js
thepassle: Can we align the implementation with how we did it in eslint-plugin-ing-security? I think its similar to what kris suggests here, and Michel tested it with both eslint 8 and 9 so that also addresses @justinfagnani s drive-by comment

koddsson:
> Drive by comment here: it would be great to have the plugin tested in both eslint 8 and 9. I've done this before with typescript versions, and it's not the most trivial thing, but not that bad with typescript at least. That could be something done later of course.

Yeah this sounds like a good shout.

koddsson:
> > Drive by comment here: it would be great to have the plugin tested in both eslint 8 and 9. I've done this before with typescript versions, and it's not the most trivial thing, but not that bad with typescript at least. That could be something done later of course.
> 
> Yeah this sounds like a good shout.

Started here: https://github.com/open-wc/open-wc/pull/2765. Looking like it's working but failing for ESLint 9 on some `parserOptions` issues but maybe we can do something clever for that. Maybe add a flat config or something for the tests.

thepassle:
@nikhilhundare in Michel's PR on eslint-plugin-security, he also needed to do some changes on `HasLitHtmlImportRuleExtension.js`, can you check with him on that and see if we need those changes in lit-a11y as well?

koddsson (approved):
Looks good to me :)

Issue #2766, opened 2024-01-18
Rendering issue in @open-wc/scoped-elements v3 for the Lion project


okadurin:
## Expected behavior

Components in [Lion](https://github.com/ing-bank/lion) project render correctly after updating `@open-wc/scoped-elements` from `v2` to `v3`.

## Actual Behavior

During updating `@open-wc/scoped-elements` from `v2` to `v3` at the `Lion` project, we discovered that some of UI components do not render anymore.
Particularly if the `lion-listbox` element [comes first](https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-integrations/test/helpers/umbrella-form.js#L112) and then the `lion-select-rich` element [comes after it](https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-integrations/test/helpers/umbrella-form.js#L130), then `lion-select-rich` does not render. When debugging it was discovered that `lion-select-rich` is using the same registry which `lion-listbox` has set when it was instantiated. The reason of the issue is in the new way `@open-wc/scoped-elements@v3` sets the registry for the elements. [Here](https://github.com/open-wc/open-wc/blob/scoped-elements-v2/packages/scoped-elements/src/ScopedElementsMixin.js#L117) is how it was done in v2 and [here](https://github.com/open-wc/open-wc/blob/%40open-wc/scoped-elements%403.0.3/packages/scoped-elements/html-element.js#L60) is how it is done for v3 . You could notice that `shouldCreateRegistry` contains more checks and it was reduced just to `!this.registry` which is not enough for our case.

Note if `lion-select-rich` comes first and then `lion-listbox` comes second, everything is rendered correctly.

### Code demo
I have prepared 2 PRs for providing more context. 

The [first one](https://github.com/ing-bank/lion/pull/2171/files) is the code from `Lion`'s master with `@open-wc/scoped-elements@v2` where the particular test file is focused and polyfill enabled. The tests in this PR do not produce any errors. 

The [second PR](https://github.com/ing-bank/lion/pull/2170/files) is done on top of the first one with the only difference that `@open-wc/scoped-elements` is update to `v3` and the minimum required code changed to address breaking changes introduced in `v3`.

To run the tests and see the related errors, one needs to run the steps as follows:
* Checkout the branch from the second PR: `git chekout bugfix/rich-select-with-polyfill-and-scoped-elements-v3`
* Remove `node_modules` just in case: `npm run rm-all-node_modules`
* `npm install`
[...]

PR #2767, opened 2024-01-18
Backport registry creation check

okadurin:
The fix for the #2766 issue. 
The code in `packages/scoped-elements/html-element.js` is the [copy](https://github.com/open-wc/open-wc/blob/scoped-elements-v2/packages/scoped-elements/src/ScopedElementsMixin.js#L106-L117) from v2


packages/scoped-elements/html-element.js
```diff
@@ -57,7 +57,19 @@ const ScopedElementsMixinImplementation = superclass =>
...
+      const shouldCreateRegistry =
+        !this.registry ||
+        // @ts-ignore
+        (this.registry === this.constructor.__registry &&
+          !Object.prototype.hasOwnProperty.call(this.constructor, '__registry'));
+
+      /**
+       * Create a new registry if:
+       * - the registry is not defined
+       * - this class doesn't have its own registry *AND* has no shared registry
```
thepassle, line 70: can we add to this comment that this is important specifically for superclasses/inheritance?

packages/scoped-elements/test-web/components/vanilla/MyButton3.js
```diff
@@ -0,0 +1,13 @@
+import { LitElement } from 'lit';
+import { ScopedElementsMixin } from '../../../html-element.js';
+
+export class MyButton3 extends ScopedElementsMixin(LitElement) {
```
thepassle, line 4: Can we have this extend from HTMLElement instead? Nice to keep things as minimal as possible

packages/scoped-elements/test-web/lit-element.test.js
```diff
@@ -134,4 +136,13 @@ describe('lit-element ScopedElementsMixin', () => {
     expect(stepB.constructor).to.equal(MyButton2);
     expect(stepB.color).to.equal('blue');
   });
+
+  it('creates unique registries when one lit element inherits another lit element', async () => {
```
thepassle, line 140: If you've changed `MyButton3` to use HTMLElement instead of LitElement, can we move this test to `html-element.test.js`?

thepassle:
Looks good, few small comments but once those are fixed we can go ahead and merge/release this :)

packages/scoped-elements/test-web/lit-element.test.js
okadurin: Should I keep `FeatureC` in the `lit` directory and then import it in the `test` file from `'./components/vanilla/index.js'`? Or should I move it to `vanilla` directory and update its rendering part to use `shadowRoot` directly?

thepassle:
@okadurin thanks for the PR!

PR #2784, opened 2024-03-25
docs: fix typo in reference types example

chriscalo:


Westbrook (approved):
Nice catch! Thanks for the fix.

PR #2788, opened 2024-04-10
Remove babel from eslint config

koddsson:
## What I did

1. I removed the babel parser dependency as I don't think that's needed any more.


koddsson:
I don't get why eslint isn't crashing in my app but everything seems to indicate that `espree` doesn't support import attributes so 🤷🏻

PR #2795, opened 2024-05-13
Update to chai@5 instead of @esm-bundle/chai

koddsson:
I would like to update to chai@5 as a replacement for `@esm-bundle/chai`. I additionally needed to upgrade `sinon-chai`. `sinon-chai` is currently in a prerelease phase so I updated to that one. If this canary release goes well we can make a PR to `master` and communicate that the prerelease of `sinon-chai` is working as expected to the `chai` maintainers.

### DIFF
--- a/packages/chai-a11y-axe/test-web/chai-a11y-axe.test.js
+++ b/packages/chai-a11y-axe/test-web/chai-a11y-axe.test.js
@@ -1,7 +1,7 @@
-import { expect, use } from 'chai';
+import { use } from 'chai';
 import { chaiA11yAxe } from '../index.js';
 
-use(chaiA11yAxe);
+const { expect } = use(chaiA11yAxe);
 
 const wrappers = [];
 
--- a/packages/semantic-dom-diff/test-web/bdd-setup.js
+++ b/packages/semantic-dom-diff/test-web/bdd-setup.js
@@ -1,6 +1,6 @@
 /* istanbul ignore next */
 // do manual setup and not use testing to not have circle dependencies
-import * as chai from 'chai';
+import { use } from 'chai';
 // eslint does not understand export maps yet
 // eslint-disable-next-line import/no-unresolved
 import { cachedWrappers } from '@open-wc/testing-helpers/pure';
@@ -21,7 +21,7 @@ if (afterEach) {
 }
 
 // register-plugins
-chai.use(chaiDomDiff);
+const chai = use(chaiDomDiff);
 
 const { expect, assert, should } = chai;
 export { expect, assert, should };
--- a/packages/storybook-addon-markdown-docs/test-node/mdjsToCsf.test.js
+++ b/packages/storybook-addon-markdown-docs/test-node/mdjsToCsf.test.js
@@ -8,8 +8,7 @@ describe('mdjsToCsf', async () => {
   let expect;
   before(async () => {
     const chai = await import('chai');
-    chai.use(chaiSnapshot);
-    expect = chai.expect;
+    expect = chai.use(chaiSnapshot).expect;
   });
 
   it('transforms a mdjs file to csf', async function it() {


packages/building-rollup/test-node/integration.test.js
```diff
@@ -9,7 +8,13 @@ const { startDevServer } = require('@web/dev-server');
 
 const rootDir = path.resolve(__dirname, '..', 'dist');
 
-describe('integration tests', () => {
+describe('integration tests', async () => {
+  let expect;
```
thepassle, line 12: cant we just

```js
const { expect } = await import('chai');
```

instead of doing it in a `before` block?

koddsson: I was doing that before but had to change to this because I was getting silent failures in mocha. Some googling around brought me to https://github.com/mochajs/mocha/issues/2975#issuecomment-355282416 which seemed to indicate that wrapping it in a `before` would fix the issue. And it seems to fix it for me.

thepassle: wow TIL, thats wild

thepassle (approved):

packages/building-rollup/test-node/integration.test.js
```diff
@@ -9,7 +8,13 @@ const { startDevServer } = require('@web/dev-server');
...
+  before(async () => {
+    const chai = await import('chai');
```
Westbrook, line 14: `({expect}=async import(‘chai’));`?

koddsson: Happy to do this if you have a strong opinion but if it's ok I'll leave this for now. There's once place at least where we need to call `chai.use(..)` in between anyways and I think we can stop doing this in a `before` once we are using ESM anyways.

packages/building-rollup/test-node/integration.test.js
koddsson: It's a bit weird. This can then all be removed once this is all ESM.

packages/building-rollup/test-node/integration.test.js
Westbrook: Separately, does `@open-wc/testing` do this for consumers, already? Or will this be a requirement to migrating?

docs/docs/testing/semantic-dom-diff.md
```diff
@@ -26,10 +26,10 @@ While `semantic-dom-diff` can be used standalone (see below), it most commonly u
 > If you are using `@open-wc/testing` this is already done for you.
 
 ```javascript
-import 'chai/chai.js';
+import * as chai from 'chai';
 import { chaiDomDiff } from '@open-wc/semantic-dom-diff';
 
-window.chai.use(chaiDomDiff);
+chai.use(chaiDomDiff);
```
koddsson, line 32: This needs to be `const chai = use(chaiDomDiff);` actually.

https://github.com/chaijs/chaijs.github.io/pull/206#pullrequestreview-2038899153

packages/building-rollup/test-node/integration.test.js
Westbrook: Not a blocker. Mainly pointing it out in the case end users need to do this every time.

packages/testing/register-chai-plugins.js
```diff
@@ -2,7 +2,7 @@
 import { chaiDomDiff } from '@open-wc/semantic-dom-diff';
 import { chaiA11yAxe } from 'chai-a11y-axe';
 
-import chai from '@esm-bundle/chai';
+import * as chai from 'chai';
```
koddsson, line 5: Running into https://github.com/chaijs/chai/issues/1603 when trying to register all the plugins further down in this file. It needs to be fixed before we can land this PR.

packages/building-rollup/test-node/integration.test.js
koddsson: Yeah we do this for consumers. We load up all the plugins and then return a instance of chai with those loaded. Found a issue with chai while checking that out that we gotta fix before landing this: https://github.com/open-wc/open-wc/pull/2795/files#r1599496849

packages/testing/register-chai-plugins.js
jorgecasar: We can simple use `import { use } from 'chai'` and change `chai.use()` by `use()`

packages/testing/index.js
```diff
@@ -1,8 +1,8 @@
-// remove esline-disable when https://github.com/esm-bundle/chai/pull/61 is merged
-/* eslint-disable-next-line import/no-unresolved */
-import chai, { expect, should, assert } from '@esm-bundle/chai';
+import * as chai from 'chai';
```
jorgecasar, line 1: it's easy to import and export just the needed modules:
```
export { expect, should, assert } from 'chai':
```

Or export all chai modules

```
export * from 'chai';
```

packages/testing/index-no-side-effects.js
```diff
@@ -1,9 +1,4 @@
-// eslint does not understand export maps yet
-/* eslint-disable import/no-unresolved */
-
-// remove esline-disable when https://github.com/esm-bundle/chai/pull/61 is merged
-/* eslint-disable-next-line import/no-unresolved */
-import chai, { expect, should, assert } from '@esm-bundle/chai';
+import chai, { expect, should, assert } from 'chai';
```
jorgecasar, line 1: `chai` doesn't export default module

docs/docs/testing/semantic-dom-diff.md
```diff
@@ -26,10 +26,10 @@ While `semantic-dom-diff` can be used standalone (see below), it most commonly u
 > If you are using `@open-wc/testing` this is already done for you.
 
 ```javascript
-import 'chai/chai.js';
+import { use } from 'chai';
 import { chaiDomDiff } from '@open-wc/semantic-dom-diff';
 
-window.chai.use(chaiDomDiff);
+const chai = use(chaiDomDiff);
```
jorgecasar, line 32: I guess it's enough with:

```
import { use } from 'chai';
import { chaiDomDiff } from '@open-wc/semantic-dom-diff';
use(chaiDomDiff);
```

package.json
```diff
@@ -99,7 +98,7 @@
     "rollup": "^3.21.0",
     "rollup-plugin-copy": "^3.4.0",
     "sinon": "^15.0.2",
-    "sinon-chai": "^3.7.0",
+    "sinon-chai": "^4.0.0-0",
```
jorgecasar, line 101: We already can use version `4.0.0`

jorgecasar:
@LarsDenBakker could you drive into it?

Issue #2796, opened 2024-05-13
lit-a11y/list exists in recommended rules list but not in the overview.md docs


Shadebane:
## Expected behavior
['Recommended rules' ](https://github.com/open-wc/open-wc/blob/7a5123dbff4bce7c23b7d18445a450569693e068/packages/eslint-plugin-lit-a11y/lib/index.js#L38) is at a minimum a subset of ['Supported rules'](https://github.com/open-wc/open-wc/blob/7a5123dbff4bce7c23b7d18445a450569693e068/docs/docs/linting/eslint-plugin-lit-a11y/overview.md#supported-rules), so that there is no confusion around which rules might be turned on by default when using the recommended rules

## Actual Behavior
'lit-a11y/list' is missing from the list of [supported rules](https://github.com/open-wc/open-wc/blob/7a5123dbff4bce7c23b7d18445a450569693e068/docs/docs/linting/eslint-plugin-lit-a11y/overview.md#supported-rules), which is causing confusion on whether or not it is a rule that is meant to be in the [recommended list](https://github.com/open-wc/open-wc/blob/7a5123dbff4bce7c23b7d18445a450569693e068/packages/eslint-plugin-lit-a11y/lib/index.js#L38) of linting rules or if it has been deprecated or something. 

## Additional context

Plugin: eslint-plugin-lit-a11y v4.1.2

Just trying to get some clarity on whether or not this rule is meant to be in the supported rules list - it looks like it may just have been missed when this file was originally ported over 4 years ago. I see no reference to any discussion on it the [original PR](https://github.com/open-wc/open-wc/pull/1875/files#diff-e60898a9393e968bb90e68f77018a720309e87a29c9aa20276a00f7f8a8fa5d5R83-R106) that would suggest otherwise, at any rate.

thepassle:
Looks like the docs just got out of sync with source code. Would you be willing to create a pr to update it?

Shadebane:
> Looks like the docs just got out of sync with source code. Would you be willing to create a pr to update it?

Done. I kept the scope to this issue but there seem to be other examples outside this rule( [definition-list](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/docs/rules/definition-list.md), [heading-hidden](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/docs/rules/heading-hidden.md), - basically any rule added since the overview list was established).

[...]

PR #2797, opened 2024-05-13
Add lit-a11y/list to eslint-plugin-lit-a11y docs/../overview.md

Shadebane:
## What I did
Address https://github.com/open-wc/open-wc/issues/2796 via the following changes

1. Add copy of list.md to ``docs/docs/linting/eslint-plugin-lit-a11y/rules`` from the package rules
2. Add a reference to this newly copied file in ``docs/docs/linting/eslint-plugin-lit-a11y/overview.md``

Side note: I noticed that the list of rules within the package && this docs folder are not aligned for other rules as well - is maintaining this documentation in two locations like this adding unnecessary maintenance cost? Just an observation as an outsider.


Westbrook (approved):
LGTM

PR #2801, opened 2024-06-20
fix(eslint-plugin-lit-a11y) fix improper regex

claydiffrient:
The existing regex would prevent valid custom elements from being detected appropriately.

The [spec](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name) indicates that a valid custom name must start with a lowercase a-z but the next word between that character and the hyphen doesn't necessarily need to be present as the `+` in the existing regex indicates.

This makes it so custom elements such as `<d-button>` or `<q-tooltip>` are detected as such.

### DIFF
--- a/.changeset/pink-squids-attend.md
+++ b/.changeset/pink-squids-attend.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-lit-a11y': patch
+---
+
+Makes regex for detecting custom elements match the spec closer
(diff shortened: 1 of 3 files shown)


claydiffrient:
I'm not sure who to tag for a review for this.  @thepassle or @Westbrook perhaps?  Apologies if tagging is against protocol for the project.  Just would like to get traction on this if at all possible.

Westbrook (approved):
LGTM.

thepassle (approved):

thepassle:
Sorry, I totally missed this

claydiffrient:
Thanks so much!

PR #2811, opened 2024-08-21
Fix signature of equalSnapshot

cristinecula:
`equalSnapshot` and `notEqualSnapshot` are async.


justinfagnani:
Nice!

IMO, it would be really great to have these libraries written in TypeScript so these types of mistakes weren't possibly in the first place.

Westbrook:
@cristinecula if you could run format on this branch, I'd be happy to approve this and get it into the next release!

PR #2821, opened 2024-11-08
Update lit-helpers.md to install correct package version

IvandeWolf:
Update installation command. Using @next installs version 0.4.0-next.1 and that version does not include the spread directives while this page implies that it does. Using without @next installs version 0.7.0 which does include the spread directives as the documentation describes.

## What I did

1. Updated the documentation for the lit-helpers page to show the correct installation command, the one currently showing is outdated.


Westbrook (approved):

Westbrook:
Thanks for the update here!

PR #2823, opened 2024-12-06
fix(lit-helpers): support passing typed objects to spread directives

### LINKED ISSUE #2705: lit-helpers spreadProps types
## Expected behavior

When I use spreadPros like this:
```
const item: InterfaceItem...
${spreadProps(item)}
```
it would expect it to throw no error

## Actual Behavior

When I use spreadPros like this:

```
const item: InterfaceItem...
${spreadProps(item)}
```

I get a TypeScript error 

```
Argument of type 'InterfaceItem' is not assignable to parameter of type '{ [key: string]: unknown; }'.
  Index signature for type 'string' is missing in type 'InterfaceItem'.
```

my Interface looks like this 

```
interface InterfaceItem {
  text?: string;
}

I can work around this when i spread my object like this:

```
${spreadProps({...item})}
```

## Additional context
if I change the type definition from 
```
export declare const spreadProps: (_spreadData: {
    [key: string]: unknown;
})
```
to
```
export declare const spreadProps: (_spreadData: {
    [key: string]: any;
})
```
there is no error thrown, but i don't know if this is the right approach. Could also be related to this: https://github.com/microsoft/TypeScript/issues/50087#issuecomment-1199411533

### PR DESCRIPTION
**Pull request by LiadIdan**
## What I did

Replaced `{ [key: string]: unknown }` with `Record<string, any>` in the `spreadProps`, `spreadEvents`, and `spread` directives to fix TypeScript compatibility issues with typed objects.

Closes #2705

### DIFF
--- a/packages/lit-helpers/src/spread.ts
+++ b/packages/lit-helpers/src/spread.ts
@@ -30,9 +30,9 @@ type EventListenerWithOptions = EventListenerOrEventListenerObject &
 export class SpreadPropsDirective extends AsyncDirective {
     host!: EventTarget | object | Element;
     element!: Element;
-    prevData: { [key: string]: unknown } = {};
+    prevData: Record<string, any> = {};
 
-    render(_spreadData: { [key: string]: unknown }) {
+    render(_spreadData: Record<string, any>) {
         return nothing;
     }
     update(part: Part, [spreadData]: Parameters<this['render']>) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/lit-helpers/src/spread.ts [hunk 1]
```diff
@@ -30,9 +30,9 @@ type EventListenerWithOptions = EventListenerOrEventListenerObject &
 export class SpreadPropsDirective extends AsyncDirective {
     host!: EventTarget | object | Element;
     element!: Element;
-    prevData: { [key: string]: unknown } = {};
+    prevData: Record<string, any> = {};
```
Westbrook, line 33: What's the functional difference between `any` and `unknown` here?

Also, will this cause type issue in the other direction?

PR #2844, opened 2025-06-02
Chore/eslint upgrades

thepassle:
## What I did

1. added flatconfig support for `eslint-plugin-lit-a11y` and refactored to esm
2. updated `@open-wc/eslint-config`, updated to flatconfig/esm
3. dropped `eslint-config-airbnb-base`
4. moved from `eslint-plugin-import` to `eslint-plugin-import-x`

Issue #2829, opened 2025-01-15
Support ESLint 9 in @open-wc/eslint-config and eslint-plugin-lit-a11y


SisyphusZheng:
Hi Open WC team,

Thank you for maintaining these fantastic tools! I’d like to request support for ESLint 9 in the following packages:

	•	@open-wc/eslint-config
	•	eslint-plugin-lit-a11y

Expected Behavior

	•	Update peer dependencies to support ESLint 9.
	•	Ensure compatibility with the latest changes in ESLint 9. 

I tried updating the packages to work with ESLint 9, but I got stuck due to the complex package management dependencies. The peer dependencies prevent smooth compatibility with ESLint 9, leading to errors and warnings.

phihu:
We ran into the same problem. I had no time to properly the solution, yet, but running it from the cli, it seems to work as it should.

Updating to ESLint 9 changes how the configuration works (incl. extensions, plugins etc.) and requires the configuration to be in a separate file. See the eslint documentation for more details: https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file  

Our project is specified as "type":"module" in its package.json, thus we went with the 'eslint.config.mjs' format (note the '.mjs'), otherwise you need to use the CommonJS format in your 'eslint.config.js' (details, see the documentaion)

For some the '@open-wc' and 'lit-ally' plugins/rules to work, you need to run

```npm install --save-dev @eslint/eslintrc```

See: https://eslint.org/docs/latest/use/configure/migration-guide#using-eslintrc-configs-in-flat-config

Our config file ('eslint.config.mjs') looks roughly like that:

```javascript
import js from '@eslint/js';
import prettier from 'eslint-config-prettier';

import { FlatCompat } from '@eslint/eslintrc';
import path from 'path';
import { fileURLToPath } from 'url';
// mimic CommonJS variables -- not needed if using CommonJS
const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
const compat = new FlatCompat({
  baseDirectory: __dirname
});

export default [
  js.configs.recommended,
  prettier,
  ...compat.plugins('lit-a11y'),
  ...compat.extends('@open-wc'),
  {
    ignores: ['dist/*'],
  }
];

```

After that, we removed the '"eslintConfig"' definitions from our package.json and everything seems to work.

Hope that helps!

PR #2865, opened 2025-08-04
ESLint Flat Config for eslint-plugin-lit-a11y - README Update

caseywatts:
Thanks to @Lassepitkanen, in https://github.com/open-wc/open-wc/pull/2844 `eslint-plugin-lit-a11y` supports the ESLint 9.x "flat config" format directly, which addresses https://github.com/open-wc/open-wc/issues/2829 🎉 

This PR adds documentation to the README about how to get set up with it.

### DIFF
--- a/docs/docs/linting/eslint-plugin-lit-a11y/overview.md
+++ b/docs/docs/linting/eslint-plugin-lit-a11y/overview.md
@@ -62,7 +62,7 @@ Then configure the rules you want to use under the rules section.
 }
 ```
 
-## Configuration
+## Flat Config Configuration (old)
 
 You may also extend the recommended configuration like so:
 


caseywatts:
Here's what I'm seeing so far, trying to debug with the ESLint Config Inspector:
https://eslint.org/docs/latest/use/configure/debug#use-the-config-inspector

image


One of my issues is my version!
Running the previous command specifies `next` as the version `npm install eslint-plugin-lit-a11y@next --save-dev`
`next` does sounds like it would be latest - but that tag hasn't been used since version 1.x, so it installs a much older version. Whoops!

The change I'm looking for was released in [version 5.0.0](https://github.com/open-wc/open-wc/releases/tag/eslint-plugin-lit-a11y%405.0.0). Installing it without the `@next` gets me up to the latest version, which is `5.1.1` today - great!

I'll update the installation instructions to be `npm install eslint-plugin-lit-a11y --save-dev`

caseywatts:
I got it working!

These setup instructions should work for folks now 👏

PR #2891, opened 2026-05-28
Chore/fix types 2

thepassle:
## What I did

1.


adamstoffel:
The last few versions of `@open-wc/scoped-elements` are broken for Typescript users (who emit declarations from their code) because the new typings reference `dist-type/types.d.ts` which isn't accessible via the `package.json` exports. -- So it's impossible for typescript to construct downstream declarations that reference this file.

Even though `types.d.ts` previously existed (and was not accessible then either), `ScopedElementsHostConstructor` didn't depend on importing that file. Since it now does, the import is inacessible. 

Is another update planned? Can I assist in some way?

The PR descriptions were empty so I'm not exactly sure of the rationale behind the changes.

thepassle:
Ill try to take a look at this today, alternatively feel free to create a PR

Sorry that this turned out to be a breaking change! The goal of this PR was a fix in the types to better support `...super.scopedElements`.

thepassle:
@adamstoffel would you be able to provide a minimal reproduction? It's not entirely clear to me what the issue is
image

adamstoffel:
@thepassle 

I've opened a PR with the fix here: https://github.com/open-wc/open-wc/pull/2895

The issue is not obvious unless you have `"declaration": true` in your tsconfig.

Screenshot 2026-06-02 at 12 06 51 PM
Screenshot 2026-06-02 at 12 06 32 PM

thepassle:
Hm, I'm a bit confused as to why this doesn't work. I have another package, `@thepassle/app-tools`. Where package export `./api.js` looks like:

```
    "./api.js": {
      "types": "./types/api/index.d.ts",
      "default": "./api.js"
    },
```
Where `./types/api/index.d.ts"` starts with:
```
/** @typedef {import('./types.js').Config} Config */
/** @typedef {import('./types.js').Method} Method */
/** @typedef {import('./types.js').Plugin} Plugin */
/** @typedef {import('./types.js').CustomRequestOptions} CustomRequestOptions */
/** @typedef {import('./types.js').RequestOptions} RequestOptions */
/** @typedef {import('./types.js').MetaParams} MetaParams */
```
which points to `./types/api/types.d.ts`, which holds the type definitions for those. When I try building tsc with `"declaration": true` for that library, it does seem able to resolve the `./types.js` import, why doesn't it for scoped elements?

thepassle:
Or rather; forget about the `@thepassle/app-tools` example, why was this not an issue in `@open-wc/scoped-elements@3.0.6`, but IS an issue in `@open-wc/scoped-elements@3.0.9`?

adamstoffel:
It depends on whether the "hidden" type is needed for typescript to generate the output types in a downstream app. If it's not needed, the error won't manifest.

Here's a minimal repro that show examples of both problematic and non-problematic code: https://github.com/adamstoffel/repro-scoped-elements-types-issue
Screenshot 2026-06-02 at 1 17 51 PM
Screenshot 2026-06-02 at 1 17 41 PM

adamstoffel:
(excuse all the typos in my comments in those screenshots 😅)

thepassle:
Forgive me if I'm being slow, but that still doesn't explain why this wasn't a problem in 3.0.6, but is an issue in 3.0.9?

thepassle:
Ah actually, I got it now.

Alright. Let me reply to your PR to fix and continue there

adamstoffel:
It looks like it _was_ a problem on 3.0.6 for `html-element.js` but not `lit-element.js`.

~~The change that's caused the issue for the Lit version are the changes on lines 8 & 12 in `lit-element.js` here: https://github.com/open-wc/open-wc/commit/78901955db3e93b0b38f082721b1ba759a6915e0#diff-67374f7fc677c80b5def858ba7b0e4a2018235baca390eaeb2335ed94665d8a8L12

The typedef for `ScopedElementsHostConstructor` added a new reference to `types.js`~~

^That could be wrong. It could be L18 that's the issue. For whatever reason, TS now wants to directly reference the `types` file after that change.

adamstoffel:
For reference, here's the output .d.ts for version 3.0.6:

```typescript
import { LitElement } from "lit";
declare const Problem1_base: typeof LitElement & import("@open-wc/scoped-elements/lit-element.js").ScopedElementsHostConstructor;
export declare class Problem1 extends Problem1_base {
}
```

And this is 3.0.9 with the patch from my PR:
```typescript
import { LitElement } from "lit";
declare const Problem1_base: typeof LitElement & import("@open-wc/dedupe-mixin").Constructor<import("@open-wc/scoped-elements/types.js").ScopedElementsHost> & import("@open-wc/scoped-elements/types.js").ScopedElementsHostConstructor;
export declare class Problem1 extends Problem1_base {
}
```

PR #2895, opened 2026-06-02
fix: [scoped-elements] expose all typings files for downstream typesc…

adamstoffel:
## What I did

1. Updated `package.json` exports for `@open-wc/scoped-elements` so that all typing files are exposed. This resolve an issue where downstream TypeScript consumers would not be able to build declarations due to the `types.d.ts` file being unexported by this package.

### DIFF
--- a/packages/scoped-elements/package.json
+++ b/packages/scoped-elements/package.json
@@ -15,13 +15,9 @@
   "homepage": "https://github.com/open-wc/open-wc/tree/master/packages/scoped-elements",
   "type": "module",
   "exports": {
-    "./html-element.js": {
-      "types": "./dist-types/html-element.d.ts",
-      "default": "./html-element.js"
-    },
-    "./lit-element.js": {
-      "types": "./dist-types/lit-element.d.ts",
-      "default": "./lit-element.js"
+    "./*.js": {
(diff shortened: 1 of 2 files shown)


packages/scoped-elements/package.json
```diff
@@ -15,13 +15,9 @@
...
   "type": "module",
   "exports": {
-    "./html-element.js": {
-      "types": "./dist-types/html-element.d.ts",
-      "default": "./html-element.js"
-    },
-    "./lit-element.js": {
-      "types": "./dist-types/lit-element.d.ts",
-      "default": "./lit-element.js"
+    "./*.js": {
```
thepassle: Alright instead of doing a `*`, lets just add a package exports entry for `"./types.js"` instead. Then at least we can't accidentally "leak" something when/if we add a new file. Then this PR should be good to go

adamstoffel: updated!

thepassle:
Preview deploy is failing, but thats an existing issue that I've yet to fix.

Thanks for bearing with me on this, and helping me understand the issue better! Will be released shortly.

adamstoffel:
All good, this was kind of a tricky one for me to debug too. Appreciate the quick response!

PR #2899, opened 2026-06-10
feat(testing-helpers): add eventHasEmitted helper

eduardocruzpalacios:
# feat(testing-helpers): add eventHasEmitted helper

## Summary

Adds a new `eventHasEmitted` helper to `@open-wc/testing-helpers` (and re-exported from `@open-wc/testing`) that checks whether a DOM event fires within a given time window.

## Motivation

`oneEvent` is useful when you _expect_ an event to fire and want to await it. But there was no built-in way to assert that an event did **or did not** fire, e.g., verifying a side-effect triggered an event, or that a certain action **does  not** emit one. This helper fills that gap.

A concrete example is form validation: you want to assert that submitting an invalid form does **not** emit `form-submitted`, and that submitting a valid one **does**. Neither case is expressible with `oneEvent` alone.

## API

```js
eventHasEmitted(eventTarget, eventName, milliseconds = 100): Promise<boolean>
```

Returns `true` if the event fired within the timeout, `false` otherwise.

## Usage

```js
import { eventHasEmitted } from '@open-wc/testing';

// assert an event was emitted
el.submit();
expect(await eventHasEmitted(el, 'submitted')).to.be.true;

// assert an event was NOT emitted
expect(await eventHasEmitted(el, 'submitted', 50)).to.be.false;
```

## Implementation notes

- The event listener is **always cleaned up**: on event fire the timeout is cancelled and the listener removed. On timeout the listener is removed. No leaks regardless of outcome.
- Default timeout is `100ms`, configurable via the third argument. Kept short intentionally: the costly case is asserting an event  **did not** fire, so a large default would slow down negative assertions.


eduardocruzpalacios:
I close this PR to create another with a more idiomatic name for the util.

